PR #4, opened 2026-03-07
Benchmark UI, parameter metadata overhaul, and search

teo-mateo:
## Summary
- Overhauled `flag_metadata.py` with updated llama-bench parameters and documentation links
- Added full benchmark UI: tab, params panel, param reference, history, and live output components with `useBenchmark` hook
- Polished service config/logs panels and parameter reference component
- Implemented search (phase 1): filtering in services table with URL-persisted query and back-navigation preservation

## Test plan
- [ ] Verify benchmark tab renders and parameters load correctly on service details page
- [ ] Test search filtering on services table (min 3 chars)
- [ ] Confirm search query persists in URL and survives back navigation
- [ ] Check parameter reference docs links work

### DIFF
--- a/dashboard/flag_metadata.py
+++ b/dashboard/flag_metadata.py
@@ -731,7 +731,7 @@ LLAMACPP_LLAMA_BENCH_FLAGS = {
         "cli": "--numa",
         "type": "string",
         "description": "NUMA scheduling (distribute|isolate|numactl)",
-        "default": "disabled",
+        "default": "",
         "tip": "NUMA scheduling mode: <code>distribute</code> (spread across all nodes), <code>isolate</code> (current node only), or <code>numactl</code> (use numactl CPU map). <b>Essential for multi-socket systems</b> to optimize memory access. <code>distribute</code> with memory interleave often provides the best throughput.",
     },
     "n_cpu_moe": {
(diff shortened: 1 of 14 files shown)


teo-mateo:
Closing as stale after 6 months, not as unwanted.

Opened 2026-03-07; the frontend has been rewritten repeatedly since. Of the 14 files, 6 are new components this PR would add (`BenchmarkTab`, `BenchmarkHistory`, `BenchmarkLiveOutput`, `BenchmarkParamsPanel`, `BenchmarkParamReference`, `useBenchmark.js`) and the other 8 are existing files — `ServicesTable.jsx`, `ServiceDetailsPage.jsx`, `ServiceConfigPanel.jsx`, `ServiceDetailsHeader.jsx`, `ServiceLogsPanel.jsx`, `ParameterReference.jsx`, `index.css`, `flag_metadata.py` — all of which have changed substantially. GitHub reports `mergeable=UNKNOWN`; rebasing +1043/-107 onto that drift is more work than rebuilding from the current tree.

**The gap it was addressing is real and is now recorded as #204.** The benchmarking backend is complete (`dashboard/benchmarking/`, endpoints live, results DB, apply-flags-back path) but the React v2 SPA has no benchmark UI at all — only the legacy `dashboard/static/benchmark.html` reaches it. That was worth catching before this branch disappeared.

The branch `feature/llama-bench-param-updates` is left on the remote and the diff stays readable via `gh pr diff 4`; #204 points at it as prior art for whoever rebuilds this.

PR #6, opened 2026-05-14
Replace log polling with SSE streaming

teo-mateo:
## Summary

- Replaces the 3-second polling loop in `ServiceLogsPanel` with a real Server-Sent Events stream, so log lines arrive as Docker emits them instead of being re-fetched in 200-line chunks every tick.
- New backend endpoint `GET /api/services/<name>/logs/stream` uses a queue + daemon reader thread (two-phase: snapshot tail then follow), keepalives every 5s, clean `stream_end` on container exit, and a `stop_event` for safe cleanup on client disconnect.
- New `useServiceLogsSSE` hook on the frontend mirrors the existing services SSE fetch + ReadableStream pattern; pausing aborts the stream, refresh reconnects, footer shows Live / Paused / Container stopped.
- Log-streaming logic extracted into `dashboard/services/log_stream.py` for testability.

## Test plan

- [ ] `cd dashboard && venv/bin/pytest tests/test_service_logs_sse.py` — 18 included unit tests cover auth, 404, headers, event ordering, tail clamping, error propagation, UTF-8 safety
- [ ] `cd dashboard/frontend && npm run build` succeeds
- [ ] Manual: open a running service's log panel in /v2 — confirm lines arrive in real time (no 3s lag); pause stops updates; refresh reconnects; stop the container and confirm footer flips to "Container stopped"
- [ ] Manual: kill the dashboard backend mid-stream — confirm the stream closes cleanly on the client without spinning

## Notes

This is one of four parallel implementations of the same feature (`feature/sse-logs-{claude,deepseek-v4flash,mistral-api,qwen27b-bf16}`). This is the Claude one. The qwen branch is a week newer and already rebased on top of the recent SSE-snapshot fallback fix on main; the others (including this one) were authored on 2026-05-03.

Will conflict with the uncommitted font-size +/- additions to `ServiceLogsPanel.jsx` currently in the working tree on main — handle when committing those.

### DIFF
--- a/dashboard/frontend/src/components/ServiceLogsPanel.jsx
+++ b/dashboard/frontend/src/components/ServiceLogsPanel.jsx
@@ -15,9 +15,9 @@ export default function ServiceLogsPanel({ serviceName, runtime }) {
   )
 
   // Track last activity time for footer display
-  const lastUpdatedRef = useRef(null)
+  const [lastUpdated, setLastUpdated] = useState(null)
   useEffect(() => {
-    if (lines.length > 0) lastUpdatedRef.current = new Date()
+    if (lines.length > 0) setLastUpdated(new Date())
   }, [lines.length])
 
   // Auto-scroll unless the user scrolled up
@@ -70,9 +70,9 @@ export default function ServiceLogsPanel({ serviceName, runtime }) {
         <div className="flex items-center gap-2">
           <button
             onClick={handleRefresh}
-            disabled={!hasContainer}
-            className="p-1.5 text-gray-400 hover:text-gray-200 disabled:opacity-50 cursor-pointer"
-            title="Refresh logs"
+            disabled={!hasContainer || paused}
+            className="p-1.5 text-gray-400 hover:text-gray-200 disabled:opacity-50 disabled:cursor-not-allowed cursor-pointer"
+            title={paused ? 'Resume to refresh' : 'Refresh logs'}
           >
             <i className="fa-solid fa-rotate-right text-sm"></i>
           </button>
@@ -118,7 +118,7 @@ export default function ServiceLogsPanel({ serviceName, runtime }) {
 
       {/* Footer */}
       <div className="px-5 py-2 border-t border-gray-700 text-xs text-gray-500 flex justify-between">
-        <span>Last updated: {formatTime(lastUpdatedRef.current)} | {lines.length} lines</span>
+        <span>Last updated: {formatTime(lastUpdated)} | {lines.length} lines</span>
         <span>{footerStatus()}</span>
       </div>
     </div>
--- a/dashboard/frontend/src/components/ServiceLogsPanel.jsx
+++ b/dashboard/frontend/src/components/ServiceLogsPanel.jsx
@@ -1,85 +1,52 @@
 import { useState, useEffect, useCallback, useRef } from 'react'
-import { fetchAPI } from '../api'
-
-const LOG_POLL_INTERVAL = 3000
-const DEFAULT_TAIL = 200
+import useServiceLogsSSE from '../hooks/useServiceLogsSSE'
 
 export default function ServiceLogsPanel({ serviceName, runtime }) {
-  const [logs, setLogs] = useState('')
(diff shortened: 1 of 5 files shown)


teo-mateo:
@claude review this PR

teo-mateo:
- /tmp/llm-dock-pr6.RndoYB/dashboard/frontend/src/components/ServiceLogsPanel.jsx:36: Refresh bypasses pause. When paused is true, handleRefresh() still calls
    refresh(), and refresh is the continuous SSE startStream, so logs resume streaming while the UI still says Paused. The refresh button should either be disabled while
    paused or perform a one-shot fetch, not start SSE.
  - /tmp/llm-dock-pr6.RndoYB/dashboard/services/log_stream.py:39: stop_event does not interrupt the blocking Docker logs(... follow=True) iterator. On client disconnect,
    pause, or refresh, routes/services.py sets the event, but this reader only checks it after Docker yields another chunk. Quiet running containers can leave one daemon
    thread and Docker stream open per abandoned connection.
  - /tmp/llm-dock-pr6.RndoYB/dashboard/frontend/src/hooks/useServiceLogsSSE.js:87: The frontend now appends every live line forever. The old poller bounded the panel to
    the latest 200 lines; this path only uses tail for the initial snapshot. Long-running vLLM logs can grow React state and the <pre> without limit.
  - /tmp/llm-dock-pr6.RndoYB/dashboard/frontend/src/components/ServiceLogsPanel.jsx:17: lastUpdatedRef is mutated in an effect, but refs do not trigger a render. After
    the first log batch, the footer can stay at --:--:-- until some unrelated later render.

teo-mateo:
Addressed all four findings in `1ba83a3`.

## Original report

> - `dashboard/frontend/src/components/ServiceLogsPanel.jsx:36`: Refresh bypasses pause. When `paused` is true, `handleRefresh()` still calls `refresh()`, and `refresh` is the continuous SSE `startStream`, so logs resume streaming while the UI still says Paused. The refresh button should either be disabled while paused or perform a one-shot fetch, not start SSE.
> - `dashboard/services/log_stream.py:39`: `stop_event` does not interrupt the blocking Docker `logs(... follow=True)` iterator. On client disconnect, pause, or refresh, `routes/services.py` sets the event, but this reader only checks it *after* Docker yields another chunk. Quiet running containers can leave one daemon thread and Docker stream open per abandoned connection.
> - `dashboard/frontend/src/hooks/useServiceLogsSSE.js:87`: The frontend now appends every live line forever. The old poller bounded the panel to the latest 200 lines; this path only uses `tail` for the initial snapshot. Long-running vLLM logs can grow React state and the `<pre>` without limit.
> - `dashboard/frontend/src/components/ServiceLogsPanel.jsx:17`: `lastUpdatedRef` is mutated in an effect, but refs do not trigger a render. After the first log batch, the footer can stay at `--:--:--` until some unrelated later render.

## Fixes

1. **Refresh + pause** — `ServiceLogsPanel.jsx`: refresh button is now `disabled={!hasContainer || paused}` with a tooltip explaining the user must resume to refresh. No more silent SSE restart behind a Paused indicator.

2. **Reader thread leak** — `services/log_stream.py`: `iter_log_events` now wraps its main loop in `try/finally`. The `finally` sets `stop_event` *and* calls `.close()` on the Docker follow-generator stored in a holder dict. Closing the generator shuts the underlying socket, which unblocks the reader thread's `for chunk in follow:` immediately — no more waiting for Docker to volunteer the next chunk on a quiet container.

3. **Unbounded buffer** — `useServiceLogsSSE.js`: introduced `MAX_LINES = 2000`. The log-append handler now head-trims when the buffer would exceed the cap. State and the `<pre>` can't grow without limit.

4. **Stale footer timestamp** — `ServiceLogsPanel.jsx`: replaced `lastUpdatedRef`/`useRef` with `lastUpdated`/`useState` + `setLastUpdated`, so the "Last updated" footer re-renders on every batch.

## Verification

- `venv/bin/python -m pytest tests/test_service_logs_sse.py` — 18/18 pass
- `npm run build` — clean

PR #7, opened 2026-05-14
Chat UX bundle + vLLM thinking traces

teo-mateo:
Five focused commits bundled together.

## Commits

1. **`feat(chat): multi-select with recursive cascade delete`** — sidebar gets hover-to-checkbox affordance on root conversations and a morphing header bar (`N chats / Select all` ↔ `N selected / trash / Clear`). Backend gets a batch `/api/chat/conversations/delete` route. Deletes now use a recursive SQLite CTE so spinoffs go with their parent instead of orphaning in the DB.

2. **`feat(chat): widen layout, larger edit textarea, copy button on code blocks`** — chat response area `max-w-4xl → max-w-6xl`, bubbles `80% → 90%`. Editing a past message opens a 60vw / 140px-min textarea instead of one bounded by the original bubble. New `CopyablePre` wraps every code block (main, spinoffs, thinking) with a hover-revealed copy button.

3. **`feat(dashboard): cross-frontend nav links + log font-size controls`** — `Try v2 ↗` on legacy index, subtle `← Back to v1` on v2 header. `ServiceLogsPanel` gets −/+ font-size buttons (persisted to localStorage, clamped 10–22px).

4. **`fix(chat): stream vLLM thinking traces in real time`** — vLLM's qwen3 reasoning parser emits `delta.reasoning`, not `delta.reasoning_content` (which llama-server uses). Accept either field in both proxy and SSE parser so the `ThinkingBlock` renders token-by-token for both backends.

5. **`docs: add project CLAUDE.md`** — runbook for adding vLLM services from the CLI plus a few gotchas (`--task` removal in 0.20, custom-arch repo caching, embedding `--gpu-memory-utilization`, silent flag typos).

## Test plan

- [ ] `cd dashboard/frontend && npm run build` — already verified locally
- [ ] Multi-select: pick a parent with spinoffs, bulk delete, confirm spinoffs vanish from sidebar AND `sqlite3 dashboard/chat.db 'SELECT count(*) FROM conversations'` reflects the reduction
- [ ] Edit past message: opens wide textarea regardless of original message length
- [ ] Copy buttons: hover any code block in chat / spinoff / thinking, click copy → check clipboard
- [ ] Log font-size: ±, reload page, value persists
- [ ] vLLM thinking: ask Qwen3.6 anything via dashboard, confirm `<think>` content streams into `ThinkingBlock` as it generates (not just at end)
- [ ] Nav links: `/` → `/v2`, back; both ways

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -93,7 +93,7 @@ test:
 
 ```
 curl -s -X POST http://localhost:3320/v1/embeddings \
-  -H "Authorization: Bearer llmd-3379c4e668fe175d2a279d64e3f664112833" \
+  -H "Authorization: Bearer <API_KEY>" \
   -H "Content-Type: application/json" \
   -d '{"model":"vllm-nomic-embed-text-v1.5","input":"search_document: hello"}'
 ```
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,103 @@
+# LLM-Dock — Claude operational notes
+
+Tactical playbook for adding / running models in this llm-dock instance.
+The general developer guide is in `AGENTS.md`; this file holds gotchas
+that don't fit there.
+
+## Adding a new vLLM service from the CLI
+
+Workflow when the dashboard isn't open or you want a scripted change:
+
+1. **Download the model into the HF cache.** Containers run with
+   `HF_HUB_OFFLINE=1` (set in `dashboard/templates/vllm.j2`), so anything
+   the model needs must already be on disk before the container starts.
+   ```
+   /home/teodor/.local/bin/hf download <org>/<model>
+   ```
+2. **Append a service entry to `services.json`.** Keys: `alias`, `port`,
+   `api_key`, `model_name`, `params` (dict of CLI flags), `template_type:
+   "vllm"`. Pick a free port in 3300–3399 (used ports are visible in
+   `services.json` itself).
+3. **Rebuild `docker-compose.yml` from `services.json`.** Don't edit the
+   generated section directly — it's between `BEGIN DYNAMIC` /
+   `END DYNAMIC` markers and gets clobbered. Instead:
+   ```
+   cd /github/teo-mateo/llm-dock/dashboard && venv/bin/python -c "
+   from compose_manager import ComposeManager
+   mgr = ComposeManager('/github/teo-mateo/llm-dock/docker-compose.yml',
+                        '/github/teo-mateo/llm-dock/services.json')
+   mgr.rebuild_compose_file()"
+   ```
+4. **Bring the container up.** `cd /github/teo-mateo/llm-dock && docker
+   compose up -d <service-name>`. Use `--force-recreate` after editing
+   params on an existing service.
+5. **Tail logs to verify.** `docker logs -f <service-name>` — wait for
+   `Application startup complete` or `Uvicorn running on http://0.0.0.0:8000`.
+
+## Gotchas
+
+### vLLM ≥0.20 removed `--task`
+
+In vLLM 0.20.x (the version in `llm-dock-vllm`), the old `--task embed`
+flag is gone. Use:
+
+- `--runner pooling` — instead of `--task embed` (also `auto`, `draft`,
+  `generate`)
+- `--convert embed` — picks the embed pooler when the model could be used
+  for multiple pooling tasks (also `auto`, `classify`, `none`)
+
+To explore the current arg surface inside the image:
+
+```
+docker run --rm --gpus all --entrypoint vllm llm-dock-vllm serve --help=all 2>&1 | grep -- '--<flag>'
+```
+
+(Without `--gpus all`, `vllm serve --help` aborts in pydantic device
+validation before printing the parser.)
+
+### Custom-arch models need both their repos cached
+
+Models with `trust_remote_code` often reference Python modules in a
+*different* HF repo via `auto_map`. Example: `nomic-ai/nomic-embed-text-v1.5`'s
+config.json points to `nomic-ai/nomic-bert-2048--modeling_hf_nomic_bert`,
+so you must download **both**:
+
+```
+hf download nomic-ai/nomic-embed-text-v1.5
+hf download nomic-ai/nomic-bert-2048
+```
+
+Symptom if you forget: container exits with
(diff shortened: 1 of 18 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,103 @@
...
+Verify with `mgr.preview_service(name)` before bringing the container up.
+
+## Reference: working embedding service
+
+`vllm-nomic-embed-text-v1.5` (port 3320) — see `services.json`. Smoke
+test:
+
+```
+curl -s -X POST http://localhost:3320/v1/embeddings \
+  -H "Authorization: Bearer llmd-3379c4e668fe175d2a279d64e3f664112833" \
```
teo-mateo: This looks like a real bearer token being committed in the runbook example. Please replace it with a placeholder like `<API_KEY>` and rotate the key if it is valid.

dashboard/frontend/src/components/chat/ChatPage.jsx
```diff
@@ -63,6 +63,14 @@ export default function ChatPage() {
     }
   }, [remove, convId, navigate])
 
+  const handleDeleteMany = useCallback(async (ids) => {
+    if (!ids || ids.length === 0) return
+    await removeMany(ids)
+    if (convId && ids.includes(convId)) {
```
teo-mateo: Because backend deletes are now recursive, this only catches the active conversation when its own id is explicitly selected. If the user is viewing a spinoff and bulk-deletes its parent, the child is deleted too but the UI stays on `/chat/<deleted-child>` and reloads into a 404/error state. The delete flow should either include descendant ids in this check or navigate away after refresh when `convId` no longer exists.

dashboard/frontend/src/components/chat/CopyablePre.jsx
```diff
@@ -0,0 +1,43 @@
...
+
+export default function CopyablePre({ children, ...rest }) {
+  const [copied, setCopied] = useState(false)
+
+  const handleCopy = (e) => {
+    e.stopPropagation()
+    const codeNode = Children.toArray(children).find(isValidElement)
+    const text = extractText(codeNode || children)
+    if (!text) return
+    navigator.clipboard.writeText(text).then(() => {
```
teo-mateo: `navigator.clipboard` is unavailable on insecure origins except localhost, and this dashboard can be accessed over plain HTTP from another host. In that case clicking Copy throws before the promise chain and the button silently does nothing. Please guard `navigator.clipboard?.writeText` and provide a fallback such as a temporary textarea + `document.execCommand('copy')`.

dashboard/chat/db.py
```diff
@@ -194,12 +194,48 @@ def update_conversation(self, conv_id: str, **kwargs) -> Optional[Conversation]:
...
+        recursive CTE. Messages cascade via FK. Returns the number of
+        conversation rows deleted (sqlite3 reports -1 for DELETE-with-subquery,
+        so we count via the same CTE first).
+        """
+        if not conv_ids:
+            return 0
+        placeholders = ",".join("?" for _ in conv_ids)
+        cte = f"""
+            WITH RECURSIVE d(id) AS (
+                SELECT id FROM conversations WHERE id IN ({placeholders})
```
teo-mateo, line 208: `UNION ALL` can duplicate rows when the input contains both a parent and one of its descendants, so `COUNT(*)` can over-report the number of deleted conversations. `DELETE ... IN (SELECT id FROM d)` still deletes the right rows, but the API response becomes inaccurate. Use `UNION` or count distinct ids.

teo-mateo (commented):
Review comments from Codex. I left inline notes for the concrete issues I found.

teo-mateo:
Addressed all four findings in `18d84f4`.

## Original review

> **`CLAUDE.md:96`** — This looks like a real bearer token being committed in the runbook example. Please replace it with a placeholder like `<API_KEY>` and rotate the key if it is valid.
>
> **`dashboard/frontend/src/components/chat/ChatPage.jsx:69`** — Because backend deletes are now recursive, this only catches the active conversation when its own id is explicitly selected. If the user is viewing a spinoff and bulk-deletes its parent, the child is deleted too but the UI stays on `/chat/<deleted-child>` and reloads into a 404/error state. The delete flow should either include descendant ids in this check or navigate away after refresh when `convId` no longer exists.
>
> **`dashboard/frontend/src/components/chat/CopyablePre.jsx:19`** — `navigator.clipboard` is unavailable on insecure origins except localhost, and this dashboard can be accessed over plain HTTP from another host. In that case clicking Copy throws before the promise chain and the button silently does nothing. Please guard `navigator.clipboard?.writeText` and provide a fallback such as a temporary textarea + `document.execCommand('copy')`.
>
> **`dashboard/chat/db.py:208`** — `UNION ALL` can duplicate rows when the input contains both a parent and one of its descendants, so `COUNT(*)` can over-report the number of deleted conversations. `DELETE ... IN (SELECT id FROM d)` still deletes the right rows, but the API response becomes inaccurate. Use `UNION` or count distinct ids.

## Fixes

1. **`CLAUDE.md`** — replaced the bearer token in the embedding smoke-test example with `<API_KEY>`. The leaked key has been rotated out of `services.json` and the running containers.

2. **`ChatPage.jsx`** — extracted `activeWillBeDeleted(ids)` helper that walks the active conversation's parent chain through the in-memory conversation list (captured pre-delete) and returns true if the active conv or any ancestor is in the deleted set. Both `handleDelete` and `handleDeleteMany` now use it, so opening a spinoff and deleting its parent navigates the UI back to `/chat` instead of stranding it on a 404.

3. **`CopyablePre.jsx`** — copy handler is now `async`, guards `navigator.clipboard?.writeText`, and falls back to a hidden-textarea + `document.execCommand('copy')` path. Copy button now works on insecure origins.

4. **`chat/db.py`** — `COUNT(*)` → `COUNT(DISTINCT id)`. Verified with a smoke test: building tree `p → {a → gc, b}`, calling `delete_conversations([p, a])` (input contains both a parent and one of its descendants) now correctly returns `4` instead of the previous over-count. The actual DELETE has always been correct because SQLite's `IN (SELECT ...)` dedupes naturally — only the API response counter was wrong.

## Verification

- `npm run build` — clean
- DB dedupe smoke test passes (`p→{a→gc,b}`; `delete_conversations([p,a])` returns 4, all rows gone)

PR #8, opened 2026-05-14
Drop polling for v2 service-detail runtime and GPU stats

teo-mateo:
Two leftover pollers in v2 were re-fetching data that the backend already exposes over SSE.

### `useServiceDetails`
- Was hitting `GET /api/services` every 3s just to pluck one row, doubling load on top of the existing `/api/services/stream` subscription.
- Now derives `runtime` from `useServicesSSE` via a `useMemo` selector. Same pattern as `useRunningServices`.
- Manual `fetchRuntime()` calls after `start` / `stop` / `restart` / `registerOpenWebUI` / `unregisterOpenWebUI` are gone — Docker event manager pushes a delta within ~1s of any container state change.
- `'not-found'` remains sourced from the per-service config endpoint (the authoritative existence check, robust against transient absence during compose reloads). SSE errors flow through as a secondary signal.
- Public return shape unchanged: `{ config, runtime, loading, error, transitioning, refetchConfig, actions }` — `ServiceDetailsPage` destructures these unchanged.

### `GpuMonitor`
- Was polling `GET /api/gpu` on a 3s `setInterval`.
- `GET /api/gpu/stream` has existed since the legacy dashboard's GPU SSE work but v2 never adopted it.
- New `useGpuSSE` hook follows the same `fetch` + `ReadableStream` pattern as `useServiceLogsSSE`, with reconnect-on-drop.
- `GpuMonitor` just consumes it and keeps the existing 20-point history buffer.

## Out of scope (intentionally)

`useServiceMetrics` still polls `/api/services/<name>/metrics` because there's no SSE backend for per-service metrics yet. Worth a follow-up — `event_manager.py` already runs per-container probes, so it could carry metrics in the same delta channel.

## Test plan

- [ ] `npm run build` — clean
- [ ] Open `/v2/service/<name>` — runtime status (starting/running/stopped pill) updates live within ~1s of `docker compose start/stop` from another shell
- [ ] DevTools Network: no recurring `GET /api/services` from the detail page (just the long-lived SSE)
- [ ] GPU monitor on `/v2` updates every ~3s with a single open `/api/gpu/stream` connection in Network tab (no recurring `/api/gpu` polls)
- [ ] Pull cable / pause backend: GPU + service-detail panels reconnect on their own within a few seconds

### DIFF
--- a/dashboard/frontend/src/components/GpuMonitor.jsx
+++ b/dashboard/frontend/src/components/GpuMonitor.jsx
@@ -1,10 +1,9 @@
 import { useState, useEffect } from 'react'
-import { fetchAPI } from '../api'
+import useGpuSSE from '../hooks/useGpuSSE'
 import GpuStats from './GpuStats'
 import GpuGraph from './GpuGraph'
 
-const MAX_POINTS = 20 // 20 points × 3s = 60s of history
-const POLL_INTERVAL = 3000
+const MAX_POINTS = 20 // 20 frames × ~3s server tick = ~60s of history
 
 function updateHistories(prev, gpus) {
   const next = { ...prev }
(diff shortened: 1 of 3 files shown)


teo-mateo:
Review note: `useServiceDetails.setPublicPort()` now refreshes only config after the port mutation, but the details header prefers `runtime.host_port` over `config.port`. Since the services SSE delta currently carries only lifecycle fields (`service_name`, `status`, `action`, `container_id`, `timestamp`) and not `host_port`, the service details header/link can keep showing the old port after moving a service to `3301` until a full services refresh/reload occurs.

Suggested fix: refresh the runtime service snapshot after `set-public-port`, or include `host_port` in the SSE update path so `runtime.host_port` is updated when compose/service port data changes.

teo-mateo:
Addressed in `2ce5bda`.

## Original note

> `useServiceDetails.setPublicPort()` now refreshes only config after the port mutation, but the details header prefers `runtime.host_port` over `config.port`. Since the services SSE delta currently carries only lifecycle fields (`service_name`, `status`, `action`, `container_id`, `timestamp`) and not `host_port`, the service details header/link can keep showing the old port after moving a service to `3301` until a full services refresh/reload occurs.
>
> Suggested fix: refresh the runtime service snapshot after `set-public-port`, or include `host_port` in the SSE update path so `runtime.host_port` is updated when compose/service port data changes.

## Fix

Went with option 1 — minimal blast radius. `setPublicPort` now also calls `refreshSSE()` (sourced from `useServicesSSE`), which re-opens the stream and pulls a fresh snapshot. The new snapshot carries the updated `host_port` for every service, so `runtime.host_port` reflects the change immediately.

Option 2 (carrying `host_port` in the delta) is the better long-term shape — same gap will hit anything else mutated on the backend that isn't a lifecycle transition. Left as a follow-up since it touches the SSE protocol on both ends.

PR #10, opened 2026-05-14
feat(chat): push auto-generated titles via SSE + larger checkbox hit area

teo-mateo:
## Summary

Two related chat-UX fixes:

1. **Auto-titles now appear in the sidebar without a refetch race** — the first-turn title was being generated in a Python daemon thread spawned *after* the SSE response had already closed. The existing ChatPage \`useEffect [streaming] -> refresh()\` fires on \`message_saved\` (strictly before the daemon's model call + DB write completes), so it kept getting \"New Conversation\" back. The title only landed on the next navigation/reload.

   Fix: run title gen inline at the tail of the same SSE response, then emit one more \`{type: \"conversation_updated\", id, title}\` event over the open pipe. Frontend patches the conversations list in place via a new \`patchConversation\` helper on \`useConversations\`. No extra HTTP roundtrip, no per-conversation pub-sub channel, no polling.

   Cost: SSE response stays open ~1–3s longer than today. \`streaming\` still flips false on \`message_saved\`, so the input box re-enables and the user can fire the next turn immediately — that turn just queues behind title gen at the inference layer (a serialization that already existed at the single-slot model).

   Subtle correctness bits:
   - The existing \`title != \"New Conversation\"\` guard is preserved AND re-checked immediately before the DB write, so a manual rename arriving DURING the model call still wins.
   - Title-gen wrapped in try/except — failures can't abort the assistant response.
   - The existing \`[streaming] -> refresh()\` effect is kept for \`updated_at\` ordering; the SSE event arrives strictly after that refresh, so last-write-wins is correct.
   - Dropped unused \`threading\` import.

2. **Conversation checkbox hit area** — the 14px input inside the 28px slot meant clicks landing in the slot but not on the input fell through to the row's \`onSelect\` (selecting the conversation instead of toggling). Slot is now 32px and is itself the toggle target when the checkbox is shown.

## Files

- \`dashboard/chat/routes.py\` — \`_auto_generate_title\` refactored to sync + return-the-title; \`send_message.generate()\` calls it inline and yields the event.
- \`dashboard/frontend/src/services/sse.js\` — new \`onConversationUpdated\` handler.
- \`dashboard/frontend/src/hooks/useChat.js\` — accepts \`{ onConversationUpdated }\`, forwards to both \`streamChat\` calls.
- \`dashboard/frontend/src/hooks/useConversations.js\` — new \`patchConversation(id, patch)\`.
- \`dashboard/frontend/src/components/chat/ChatPage.jsx\` — wires \`patchConversation\` through \`useChat\`.
- \`dashboard/frontend/src/components/chat/ChatSidebar.jsx\` — enlarged checkbox slot + slot-as-target onClick.

## Test plan

- [ ] Restart the dashboard backend (route change).
- [ ] New conversation → send first message → sidebar entry flips from \"New Conversation\" to the model-generated title within ~1–3s of the assistant text rendering, no reload needed. DevTools Network: \`POST /api/chat/conversations/<id>/messages\` SSE stays open after the assistant text appears, then closes.
- [ ] Thinking model (Qwen reasoning service as \`main_service\`): title still generated (verifies the \`reasoning_content\` fallback).
- [ ] Manual rename mid-flight: rename the conversation while the assistant is still streaming the first turn — the manual title persists.
- [ ] Navigate away mid-flight: switch conversations the moment the assistant text fully renders. The DB row should still get the new title; the next \`refresh()\` shows it in the sidebar.
- [ ] Second message on the same conversation: no \`conversation_updated\` event fires (\`is_first\` is false server-side).
- [ ] Conversation checkbox: clicking anywhere in the slot toggles the checkbox; clicking the row outside the slot still navigates to the conversation.

### DIFF
--- a/dashboard/chat/routes.py
+++ b/dashboard/chat/routes.py
@@ -1,8 +1,8 @@
 import json
 import logging
 import os
-import threading
 import uuid
+from typing import Optional
 
 from flask import Blueprint, jsonify, request, current_app, Response, stream_with_context
 
@@ -215,42 +215,53 @@ def _stream_response(db: ChatDB, conv: Conversation, user_msg: Message, mcp_mana
         yield f"data: {json.dumps({'error': 'Stream ended unexpectedly'})}\n\n"
 
 
-def _auto_generate_title(app, db_path: str, conv_id: str, first_user_content: str, service_name: str):
-    """Background thread to auto-generate a conversation title."""
-    try:
-        from .db import ChatDB as _ChatDB
-        _db = _ChatDB(db_path)
-        conv = _db.get_conversation(conv_id)
-        if conv is None or conv.title != "New Conversation":
-            return
+def _auto_generate_title(db, conv_id: str, first_user_content: str, service_name: str) -> Optional[str]:
+    """Generate a 3–6 word title for a fresh conversation and persist it.
 
-        title_messages = [
-            {"role": "system", "content": "Generate a concise 3-6 word title for a conversation that starts with the following message. Return ONLY the title, nothing else. Do not explain or add anything."},
-            {"role": "user", "content": first_user_content},
-        ]
+    Runs synchronously inside the SSE response that delivered the assistant
+    reply — see send_message. Returns the new title on success so the caller
+    can emit a conversation_updated event to the open client, or None if
+    nothing should be pushed (manual rename already won, model produced
+    empty output, etc).
+    """
+    conv = db.get_conversation(conv_id)
+    if conv is None or conv.title != "New Conversation":
+        return None
 
-        full_content = ""
-        full_reasoning = ""
-        for event_type, data in stream_chat_completion(service_name, title_messages):
-            if event_type == "delta":
-                full_content += data.get("content", "")
-            elif event_type == "done":
-                full_content = data["content"]
-                full_reasoning = data.get("reasoning_content") or ""
-                break
+    title_messages = [
+        {"role": "system", "content": "Generate a concise 3-6 word title for a conversation that starts with the following message. Return ONLY the title, nothing else. Do not explain or add anything."},
+        {"role": "user", "content": first_user_content},
+    ]
 
-        # Some models put short responses in reasoning_content instead of content
-        title = full_content.strip() or full_reasoning.strip()
-        # Extract last non-empty line as the likely title if reasoning is verbose
-        if '\n' in title:
-            lines = [l.strip().strip('"\'').strip() for l in title.split('\n') if l.strip()]
-            title = lines[-1] if lines else title
-        title = title.strip('"\'*#').strip()
-        if title:
-            _db.update_conversation(conv_id, title=title[:100])
-            logger.info(f"Auto-generated title for {conv_id}: {title}")
-    except Exception:
-        logger.exception("Failed to auto-generate conversation title")
+    full_content = ""
+    full_reasoning = ""
+    for event_type, data in stream_chat_completion(service_name, title_messages):
+        if event_type == "delta":
+            full_content += data.get("content", "")
+        elif event_type == "done":
+            full_content = data["content"]
+            full_reasoning = data.get("reasoning_content") or ""
+            break
+
+    # Some models put short responses in reasoning_content instead of content
+    title = full_content.strip() or full_reasoning.strip()
+    if '\n' in title:
+        lines = [l.strip().strip('"\'').strip() for l in title.split('\n') if l.strip()]
+        title = lines[-1] if lines else title
+    title = title.strip('"\'*#').strip()
+    if not title:
+        return None
+
+    # Re-check before write: a rename arriving DURING the model call would
+    # otherwise be clobbered.
+    conv2 = db.get_conversation(conv_id)
+    if conv2 is None or conv2.title != "New Conversation":
+        return None
+
+    final = title[:100]
+    db.update_conversation(conv_id, title=final)
+    logger.info(f"Auto-generated title for {conv_id}: {final}")
+    return final
 
 
 @chat_bp.route("/api/chat/conversations/<conv_id>/messages", methods=["POST"])
@@ -288,15 +299,16 @@ def send_message(conv_id):
 
     def generate():
         yield from _stream_response(db, conv, user_msg, mcp_manager=mcp_manager)
-        # Generate title after the main response is done (same model, single slot)
+        # Generate the title inline so we can push it to the client over the
+        # same SSE pipe before closing — otherwise the sidebar shows
+        # "New Conversation" until the next manual refetch.
         if is_first:
-            db_path = db.db_path
-            t = threading.Thread(
-                target=_auto_generate_title,
-                args=(None, db_path, conv_id, content, conv.main_service),
-                daemon=True,
-            )
-            t.start()
+            try:
+                new_title = _auto_generate_title(db, conv_id, content, conv.main_service)
+                if new_title:
+                    yield f"data: {json.dumps({'type': 'conversation_updated', 'id': conv_id, 'title': new_title})}\n\n"
+            except Exception:
+                logger.exception("auto-title failed")
 
     return Response(stream_with_context(generate()), mimetype="text/event-stream",
                     headers={"Cache-Control": "no-cache", "X-Accel-Buffering": "no"})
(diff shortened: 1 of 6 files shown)


teo-mateo:
Review findings:

1. `dashboard/frontend/src/hooks/useChat.js:124` aborts the same SSE stream before the new `conversation_updated` event can arrive.

   `onMessageSaved` calls `loadConversation(conversation.id)`, and `loadConversation()` immediately aborts `abortRef.current` at lines 20-23. Since the server now emits `conversation_updated` only after `_stream_response()` completes and after title generation, the client will usually cancel the fetch right after `message_saved`, so the sidebar patch never runs. This defeats the main PR behavior. The reload/abort needs to happen after the SSE stream fully closes, or this particular reload path needs to avoid aborting the active stream.

2. `dashboard/frontend/src/components/chat/ChatSidebar.jsx:27` and line 49 still only toggle the slot when `checkboxShown` is true.

   The checkbox is also visible on hover via `group-hover/iconslot:opacity-100` while `checkboxShown` remains false. In that normal hover state, clicking the enlarged 32px slot outside the 14px input still bubbles to the row and selects the conversation instead of toggling. The hit-area fix should account for hover-visible checkboxes, not just select mode / already-selected rows.

I did not run the test suite; this was a read-only review.

teo-mateo:
Addressed both findings in cd064e4.

## Original review

> 1. `dashboard/frontend/src/hooks/useChat.js:124` aborts the same SSE stream before the new `conversation_updated` event can arrive.
>
>    `onMessageSaved` calls `loadConversation(conversation.id)`, and `loadConversation()` immediately aborts `abortRef.current` at lines 20-23. Since the server now emits `conversation_updated` only after `_stream_response()` completes and after title generation, the client will usually cancel the fetch right after `message_saved`, so the sidebar patch never runs. This defeats the main PR behavior. The reload/abort needs to happen after the SSE stream fully closes, or this particular reload path needs to avoid aborting the active stream.
>
> 2. `dashboard/frontend/src/components/chat/ChatSidebar.jsx:27` and line 49 still only toggle the slot when `checkboxShown` is true.
>
>    The checkbox is also visible on hover via `group-hover/iconslot:opacity-100` while `checkboxShown` remains false. In that normal hover state, clicking the enlarged 32px slot outside the 14px input still bubbles to the row and selects the conversation instead of toggling. The hit-area fix should account for hover-visible checkboxes, not just select mode / already-selected rows.

## Fixes

**1.** Extracted `refetchMessages(convId)` — server refetch only, no abort, no stream reset. Used from both `onMessageSaved` sites (sendMessage + editMessage). `loadConversation` keeps its full-reset semantics for the conversation-switch path (and the editMessage `onError` fallback where the stream is already over anyway). The trailing `conversation_updated` event can now arrive without the client cancelling itself.

**2.** Converted the iconslot `<div>` to a `<label>` wrapping the input. Native label-input pairing means any click in the 32px area toggles the wrapped checkbox regardless of `checkboxShown` — no manual `onToggleSelect` call needed, no conditional guard. `stopPropagation` on the label still prevents the row's select handler from also firing on the same click. Hover-only, selected, and selectMode cases all behave correctly.

Frontend rebuilt clean.

teo-mateo:
Follow-up review after cd064e4:

The two original findings look addressed: `onMessageSaved` no longer self-aborts the SSE stream, and the checkbox slot is now a label-backed hit area.

Remaining issue:

`dashboard/frontend/src/hooks/useChat.js:33` can still abort the title-generation SSE when the user switches conversations after `message_saved` but before `conversation_updated`.

`refetchMessages()` fixes the immediate self-abort in `onMessageSaved`, but `abortRef.current` still points at the still-open SSE while `_auto_generate_title()` runs after `_stream_response()` completes. Since `setStreaming(false)` runs at `message_saved`, the UI allows navigation. If the user switches conversations in that window, `loadConversation()` aborts the request, so the server may not finish title generation and the trailing `conversation_updated` event is definitely lost. That breaks the PR’s stated “navigate away mid-flight” test case.

The title work is still coupled to a client-owned stream that the UI intentionally aborts during navigation. Either title generation needs to survive client disconnect/navigation, or the client should not abort this post-message title phase if the server is expected to finish it.

teo-mateo:
Addressed in 6958092.

## Original follow-up finding

> `dashboard/frontend/src/hooks/useChat.js:33` can still abort the title-generation SSE when the user switches conversations after `message_saved` but before `conversation_updated`.
>
> `refetchMessages()` fixes the immediate self-abort in `onMessageSaved`, but `abortRef.current` still points at the still-open SSE while `_auto_generate_title()` runs after `_stream_response()` completes. Since `setStreaming(false)` runs at `message_saved`, the UI allows navigation. If the user switches conversations in that window, `loadConversation()` aborts the request, so the server may not finish title generation and the trailing `conversation_updated` event is definitely lost. That breaks the PR's stated "navigate away mid-flight" test case.
>
> The title work is still coupled to a client-owned stream that the UI intentionally aborts during navigation. Either title generation needs to survive client disconnect/navigation, or the client should not abort this post-message title phase if the server is expected to finish it.

## Fix: drain-aware abort

Added a `drainingRef` flag flipped true at `message_saved` and reset on every new send / loadConversation / stopStreaming. `loadConversation` now skips the abort while draining:

```js
if (abortRef.current) {
  if (!drainingRef.current) abortRef.current.abort()
  abortRef.current = null
}
drainingRef.current = false
```

The drained stream's callbacks remain bound via closures — when the trailing `conversation_updated` event arrives, `patchConversation` still updates the **original** conversation's title in the sidebar even though the user is now viewing a different one. The flag resets on the next user action so it can't get stuck on stale.

### Why this is safe across edge cases

- **Navigate to another conversation post-save**: drainingRef=true, abort skipped, stream finishes, title patches the original conv's sidebar entry. ✓
- **Send another message in the new conversation**: new `sendMessage` resets drainingRef=false and stores a new controller in abortRef. The old stream is no longer in abortRef but is still consumed by its async iterator; its callbacks still fire and patch the right conversation by ID. ✓
- **User clicks Stop during streaming**: streaming=true, drainingRef=false → normal abort. The Stop button is hidden post-`message_saved` (streaming=false), so the drain window is user-uninterruptible by design. ✓
- **User leaves /chat entirely**: ChatPage unmounts. The unmount cleanup (effect at line 240) is unchanged and still aborts. Title gen gets BrokenPiped server-side; sidebar picks up "New Conversation" until the next `listConversations` refresh. Acceptable — the user isn't watching the sidebar anymore.
- **Server crash mid-title-gen**: existing behavior — title stays "New Conversation" until next refresh.

### What this doesn't do (deliberately)

The reviewer offered two options: "title generation needs to survive client disconnect/navigation" OR "the client should not abort this post-message title phase". This PR takes the second route — strictly less invasive than detaching title-gen from the request lifecycle (which would need a separate broadcast channel or a job-queue style backend). The first option is a larger refactor and not warranted by this single delivery channel.

Frontend rebuilt clean.

teo-mateo:
Addressed in 7e38cb9.

## Original finding

> `dashboard/frontend/src/hooks/useChat.js:48` detaches the draining stream from `abortRef`, so the unmount cleanup can no longer abort it.
>
> The new drain-aware path fixes the prior "switch conversation after `message_saved`" case, but when `loadConversation()` skips the abort it still immediately sets `abortRef.current = null`. If the user switches conversations during the title-tail phase and then leaves `/chat` before `conversation_updated` arrives, the cleanup at unmount has no controller to abort. The old fetch keeps running after `ChatPage`/`useConversations` unmount, and its callbacks can still call `patchConversation` / React state setters on an unmounted tree. Either keep enough tracking to abort detached draining streams on unmount, or guard those callbacks/state updates with mounted state.

## Fix

Added `liveControllersRef` — a `Set<AbortController>` of every controller whose stream hasn't yet ended. Entries:
- added immediately after the `AbortController` is created in both `sendMessage` and `editMessage`
- removed in a `try { await streamChat(...) } finally { liveControllersRef.current.delete(controller) }` block when the stream resolves (normal end, error, or abort)

The unmount cleanup now sweeps the whole set, not just `abortRef`:

```js
return () => {
  for (const c of liveControllersRef.current) c.abort()
  liveControllersRef.current.clear()
  abortRef.current = null
}
```

This covers every path that leaves a controller live:
- active stream — always in both `abortRef` AND `liveControllersRef`.
- draining stream detached by `loadConversation` — gone from `abortRef`, still in `liveControllersRef` until its server side ends.
- both at once — user sends, message_saved, switches conv, sends again (new ctrl in `abortRef`, old ctrl still draining in `liveControllersRef`), then unmounts before either tail finishes. Both get aborted.

Frontend rebuilt clean.

PR #12, opened 2026-05-14
fix(chat): persist partial assistant response on client disconnect

teo-mateo:
## Summary

When the model is still generating an assistant reply and the user navigates away — to another conversation, to a different route, or by closing the tab — Flask raises \`GeneratorExit\` at the next yield in \`_stream_response\` (\`dashboard/chat/routes.py:133\`). Previously the function exited without writing anything; the assistant message was only persisted inside the \`event_type == \"done\"\` branch, so any disconnect before \"done\" arrived left the DB with only the user message. Coming back to the conversation, the user saw an empty response slot.

This PR is the **safety-net layer** (Option A): accumulate deltas, save whatever was generated on disconnect. A separate follow-up could keep the SSE stream alive across in-app navigations (Option B), but the floor needs to be \"no data loss even if the laptop closes.\"

## Approach

Wrap the streaming \`for\` loop in \`try / except GeneratorExit / finally\`. Inside, two new accumulators (\`accumulated_content\`, \`accumulated_reasoning\`) gather what the upstream deltas already expose (\`stream_chat_completion\` yields content/reasoning per-delta — see \`llm_proxy.py:131-135\`). \`_save_partial()\` writes the assistant message to DB on disconnect, gated by:

- \`saved_partial\` — single-save invariant.
- \`done\` — normal-completion path owns the save, don't double-write.
- \`had_error\` — server-side model error path explicitly opts out.
- Non-empty accumulators — don't write a zero-byte row.
- **Stale-stream guards** — \`db.get_message(user_msg.id) is None\` (user message edited away) or \`db.next_seq(conv.id) - 1 != user_msg.seq\` (newer messages landed). Either drops the partial silently so an old disconnected stream can't append after fresh state.

## Subtle correctness fixes in the same function

- **\`done = True\` is flipped IMMEDIATELY after \`db.add_message(assistant_msg)\`** in the done branch — before artifact saves and \`touch_conversation\`. If an artifact save throws after the assistant row is inserted, the finally clause's \`_save_partial\` would otherwise attempt a second insert with the same \`assistant_msg_id\`, hit a PK collision, and log a misleading \"Failed to save partial\".

- **\`had_error = True\` is set BEFORE yielding the error SSE.** \`finally\` always runs after the error branch's \`return\`. If the client disconnects DURING the error yield, \`GeneratorExit\` fires and \`finally\` runs — without setting the flag first, a server-side model failure with accumulated deltas would still write a misleading partial.

## What this doesn't do (deliberate)

- Frontend keep-alive across conversation switches (Option B). With this PR landed, that becomes optional — the partial is in DB regardless of the frontend's abort behavior.
- Truncation marker on partial messages. Partial sentences self-identify; we can add an indicator later if confusing.
- Resume from a saved partial (would need model-specific state).
- Tool-result reattachment. Partial tool_calls (no result) are excluded from the saved row to avoid showing half a transaction.

## Tests

\`dashboard/tests/test_partial_save.py\` — 5 cases, all passing locally:

1. \`test_partial_save_persists_accumulated_deltas_on_disconnect\` — drives the generator, consumes two deltas, calls \`gen.close()\` (Python's equivalent of GeneratorExit). DB ends up with user + partial assistant carrying the accumulated content.
2. \`test_close_before_any_delta_writes_nothing\` — generator advances to a tool_call yield (no content), then close. No assistant row.
3. \`test_error_event_does_not_save_partial\` — stub yields some deltas then an error. \`had_error\` gate fires; no assistant row.
4. \`test_stale_stream_skips_save_when_user_msg_deleted\` — user message deleted mid-stream (simulating an edit racing the disconnect). Save is skipped with a \"no longer exists\" info log.
5. \`test_normal_completion_does_not_double_save\` — happy path; the finally clause is a no-op when done=True.

The 100 other backend tests still pass.

## Manual smoke (post-merge, requires running dashboard)

- [ ] Start a long generation, click another conversation mid-stream, wait, click back — partial assistant present.
- [ ] Same, but navigate to \`/v2/services\` and back — partial present.
- [ ] Same, but close the tab mid-stream — partial present after reopen.
- [ ] Normal completion produces exactly one assistant row (no double-save).
- [ ] Point the conversation at a stopped service and send — error flows through, no partial row.
- [ ] Edit the user message in tab B while tab A's stream is mid-flight — backend log shows \"Skipping partial-save for ... user message no longer exists\"; the original conv's history reflects the edit, not a phantom partial.

### DIFF
--- a/dashboard/chat/routes.py
+++ b/dashboard/chat/routes.py
@@ -157,62 +157,142 @@ def _stream_response(db: ChatDB, conv: Conversation, user_msg: Message, mcp_mana
 
     assistant_msg_id = str(uuid.uuid4())
     done = False
+    had_error = False              # gate _save_partial against the error path
+    saved_partial = False          # avoid double-save
+    accumulated_content = ""       # per-delta accumulators for partial-save
+    accumulated_reasoning = ""
     collected_tool_calls = []
     collected_artifacts = []
 
-    for event_type, data in stream:
-        if event_type == "delta":
-            yield f"data: {data['raw']}\n\n"
-        elif event_type == "tool_call":
-            collected_tool_calls.append({
-                "name": data["name"],
-                "arguments": data["arguments"],
-                "server_id": data["server_id"],
-            })
-            yield f"data: {json.dumps({'type': 'tool_call', 'name': data['name'], 'arguments': data['arguments'], 'server_id': data['server_id']})}\n\n"
-        elif event_type == "tool_result":
-            for tc in reversed(collected_tool_calls):
-                if tc["name"] == data["name"] and "result" not in tc:
-                    tc["result"] = data["result"]
-                    break
-            yield f"data: {json.dumps({'type': 'tool_result', 'name': data['name'], 'result': data['result'], 'server_id': data['server_id']})}\n\n"
-        elif event_type == "artifact":
-            collected_artifacts.append(data)
-            # Send artifact to frontend (without full content to keep SSE lean)
-            yield f"data: {json.dumps({'type': 'artifact', 'artifact_type': data['type'], 'title': data.get('title'), 'content': data['content']}, default=str)}\n\n"
-        elif event_type == "done":
-            next_seq = db.next_seq(conv.id)
-            assistant_msg = Message(
+    def _save_partial():
+        """Persist whatever we accumulated when the client disconnects mid-stream.
+
+        Called from the GeneratorExit handler and the finally clause. No-op
+        when the stream completed normally, errored, already saved, or never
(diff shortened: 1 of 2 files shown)


teo-mateo:

I reviewed the PR metadata, diff, prior issue comments/reviews (none present), and the surrounding chat streaming, DB, tool-loop, and frontend SSE handling. I also ran the focused test file with the repo virtualenv:

`./venv/bin/python -m pytest dashboard/tests/test_partial_save.py`

Result: 5 passed. I do not see additional correctness, security, API-contract, or production-impact issues to raise for this iteration.

PR #17, opened 2026-05-14
feat(chat): add Render HTML MCP server (closes #15)

teo-mateo:
## Summary

Registers `render-html-mcp` as an optional stdio MCP server. With the toggle on, the model can render generated HTML or Markdown in a local desktop browser window through any of four tools: `render_html`, `render_html_from_file`, `render_html_from_markdown`, `render_html_from_markdown_file`.

- New `render-html` entry in `dashboard/chat/mcp_registry.py`, icon `fa-window-restore`.
- Command path resolves from `LLM_DOCK_RENDER_HTML_MCP_COMMAND` env var (defaults to the local checkout), so the absolute path isn't baked in for other hosts.
- Marked `external: True` so it picks up the existing executable-presence gate from PR #14 — on a host without the path, the toggle simply doesn't appear in the UI and no false `you-have-render-html` hint leaks into the system prompt.
- Search-side scope: all four tools are exposed (the registry has no per-tool filtering today). The `tool_hint` nudges the model toward the generated-content tools (`render_html`, `render_html_from_markdown`) and explicitly warns that the `*_from_file` variants are only for paths the user has named — never inferred or guessed. It also tells the model what to do if it sees a DISPLAY/WAYLAND error (i.e., stop trying).
- No frontend changes needed.

## Test plan

- [x] Registry lists `render-html` after the change.
- [x] `MCPClientManager.get_tools('render-html')` discovered all 4 tools cleanly over stdio.
- [x] Stdout stays clean JSON-RPC (no startup prints to stdout in the package).
- [ ] In the chat UI on a GUI session, enable Render HTML and ask for a small HTML/Markdown preview; confirm `tool_call` / `tool_result` events and a browser window opens.
- [ ] On a headless setup, confirm the tool returns a clean DISPLAY/WAYLAND error and the model surfaces it to the user instead of retrying.

PR #18, opened 2026-05-14
feat(chat): replace external render-html-mcp with built-in artifact server

teo-mateo:
## Summary

PR #17's external render-html-mcp opens a **native desktop browser** via `webbrowser.open()`. That doesn't work when the dashboard runs as a systemd service (no `DISPLAY` / `WAYLAND_DISPLAY` / `XAUTHORITY` in the process env), and even if it did, the `mcp` SDK's `get_default_environment()` only forwards `HOME, LOGNAME, PATH, SHELL, TERM, USER` to child processes — graphical-env vars get stripped before they reach the subprocess.

The user is already looking at a browser tab. So instead of opening *another* browser via subprocess, return the rendered HTML as a chat artifact and let `ArtifactRenderer.jsx` (which already supports `type: 'html'` with a sandboxed iframe) display it inline.

## Changes

- **New** `dashboard/chat/mcp_servers/render_html_server.py` — FastMCP server exposing:
  - `render_html(html, title)` — passes through full documents; wraps fragments in a minimal HTML5 shell.
  - `render_html_from_markdown(markdown, title)` — CommonMark + tables + strikethrough via the already-installed `markdown-it-py`, wrapped in a small default stylesheet so headings/code/tables look reasonable.
  - Both return the standard `{artifact: true, type: 'html', content, title, description}` JSON the existing pipeline recognizes (`mcp_client.py:134-145`).
- **Modified** `dashboard/chat/mcp_registry.py`:
  - Drops `_RENDER_HTML_MCP_COMMAND` / `LLM_DOCK_RENDER_HTML_MCP_COMMAND` indirection — server is in-tree now, under `sys.executable` like sympy/schemdraw.
  - Drops `external: True` from the entry. The `_server_available()` mechanism stays in place for the websearch entry.
  - Rewrites the tool hint around in-browser rendering. Drops the headless-error guidance and any mention of `*_from_file` (those tools are gone on purpose — local-file reads on the dashboard host had no value when rendering happens in the user's browser).
- **Modified** `dashboard/frontend/src/components/chat/ArtifactRenderer.jsx`:
  - Adds a "Pop out" button (only on `type: 'html'`) that opens the html in a new tab via `URL.createObjectURL` for users who want a full-window view. Trade-off: the popped-out tab is at a `blob:` origin and runs scripts with full privileges there; the inline iframe stays sandboxed. User has to click; same trust model as Download.

The external `ai-toolbox/mcp/render-html-mcp` package is untouched — anyone who wants the native-popup behavior outside llm-dock can still use it standalone.

## Test plan

- [x] Backend smoke: `MCPClientManager.get_tools('render-html')` returns 2 tools; `call_tool` for both returns an `html` artifact with content starting with `<!doctype html>`; markdown tool produces `<li>` and `<strong>` from list + bold input.
- [x] Frontend builds cleanly.
- [x] Restart dashboard, enable Render HTML on a conversation, ask the model for a small HTML report; expect an inline sandboxed iframe in the chat with Pop out + Download buttons.
- [x] Ask the model to render a markdown document; expect formatted output (headings, lists, code blocks) styled by the default CSS.
- [x] Reload the conversation; expect the artifact to re-render from the `artifacts` table.
- [x] Confirm "Pop out" opens a new tab with the same content full-width.

### DIFF
--- a/dashboard/chat/mcp_registry.py
+++ b/dashboard/chat/mcp_registry.py
@@ -10,11 +10,6 @@ _WEBSEARCH_MCP_ROOT = os.environ.get(
     "/github/teo-mateo/ai-toolbox/mcp/websearch-mcp",
 )
 
-_RENDER_HTML_MCP_COMMAND = os.environ.get(
-    "LLM_DOCK_RENDER_HTML_MCP_COMMAND",
-    "/github/teo-mateo/ai-toolbox/mcp/render-html-mcp/.venv/bin/render-html-mcp",
-)
-
 MCP_SERVERS = {
     "sympy-math": {
         "name": "SymPy Math",
@@ -151,11 +146,10 @@ The diagram will be rendered automatically as an artifact.""",
     },
     "render-html": {
         "name": "Render HTML",
-        "description": "Render HTML or Markdown in a local desktop browser window (requires GUI session)",
-        "command": [_RENDER_HTML_MCP_COMMAND],
+        "description": "Render HTML or Markdown as an artifact in the chat",
+        "command": [sys.executable, os.path.join(_SERVERS_DIR, "render_html_server.py")],
         "icon": "fa-window-restore",
-        "tool_hint": "You can render HTML or Markdown in a local desktop browser window when the user asks for a preview of a document, report, table, chart, or any formatted output. Prefer render_html for generated HTML you've just produced and render_html_from_markdown for generated markdown. The *_from_file variants open arbitrary local paths — only use them when the user has explicitly named a file to open, never on a path you inferred or guessed. The tool requires a graphical session; if it returns a DISPLAY/WAYLAND error, tell the user the dashboard is running headless and stop trying.",
-        "external": True,
+        "tool_hint": "You can render HTML or Markdown directly in this chat window. Use render_html when you've produced an HTML document or fragment the user should see formatted (reports, tables, dashboards, mock UI, charts, etc.) — pass a complete <!doctype html> document if you want full control over styling, or just a fragment if you don't care. Use render_html_from_markdown when the content is markdown and you want it styled nicely without writing HTML yourself. The output appears as an artifact panel attached to your message, inside a sandboxed iframe — the user can also pop it out to a full-window tab. There is NO file-system access: you cannot pass paths, only literal content.",
     },
 }
 
--- a/dashboard/chat/mcp_servers/render_html_server.py
+++ b/dashboard/chat/mcp_servers/render_html_server.py
@@ -1,0 +1,116 @@
+"""Render HTML MCP Server — turn model-generated HTML or Markdown into
+chat artifacts that the dashboard renders in a sandboxed iframe.
+
+No native browser, no DISPLAY, no subprocess. Just emits the existing
+artifact convention (mcp_client.py:134-145) so the dashboard's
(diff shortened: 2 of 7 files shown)


teo-mateo:

- [High] Keep popped-out HTML sandboxed. `ArtifactRenderer.jsx:20-22` creates a same-origin `blob:` HTML document and opens it as a top-level page. Unlike the inline iframe at `ArtifactRenderer.jsx:94-99`, that page is not sandboxed. Because dashboard auth is stored in `localStorage` (`dashboard/frontend/src/api.js:5-8`) and API calls use that bearer token, any script in model-generated HTML can read the token and/or make authenticated dashboard requests once the user clicks Pop out. `noopener` only removes the opener reference; it does not sandbox the blob document. Please keep the untrusted artifact inside a sandbox boundary for pop-out as well, for example by opening a static wrapper page/blob that contains a full-window sandboxed iframe with the artifact content escaped, or by omitting script privileges for popped-out content unless explicitly needed.

- [Medium] Declare `markdown-it-py` as a direct backend dependency. `dashboard/chat/mcp_servers/render_html_server.py:11` imports `markdown_it`, but `dashboard/requirements.txt` does not list `markdown-it-py`. It is currently available in this environment through transitive dependencies, but a direct import from application code should be pinned/listed directly; otherwise a clean install can lose the render-html MCP server when an unrelated dependency changes its dependency tree.

teo-mateo:

**Finding 1.1 — Pop-out HTML not sandboxed**: accepted.

Fixed in commit a9f36f5. The previous popout opened the model's HTML as a top-level \`blob:\` page. A blob URL inherits the creating document's origin, so it was same-origin with the dashboard and could read the bearer token from \`localStorage\` (see \`api.js\`) and make authenticated fetches against \`/api\`. \`noopener\` only removes the opener reference, not the origin boundary.

New behavior: \`popoutArtifact\` builds a tiny script-free wrapper page whose only content is a \`<iframe srcdoc="..." sandbox="allow-scripts">\` filling the viewport. The wrapper has no scripts; the model's HTML runs only inside the sandboxed iframe, where it gets an opaque origin and cannot reach \`parent.localStorage\` or send requests with the dashboard's session. Same security posture as the inline iframe. The artifact content is HTML-attribute-escaped (\`&\` → \`&amp;\`, \`"\` → \`&quot;\`) before being placed in \`srcdoc\`.

**Finding 1.2 — markdown-it-py not in requirements.txt**: accepted.

Fixed in commit a9f36f5. Added \`markdown-it-py>=3.0.0\` to \`dashboard/requirements.txt\`. It was previously available only transitively (likely through \`mcp[cli]\` → \`rich\`), so a clean install could lose it if the transitive chain changes upstream.

teo-mateo:

I reviewed the updated diff and prior discussion. The two iteration 1 findings appear addressed: the pop-out path now keeps model HTML inside a sandboxed iframe wrapper, and `markdown-it-py` is listed as a direct dependency.

I also checked the updated MCP server wiring and ran quick verification:
- `dashboard/venv/bin/python` can import and call the new render-html server tools.
- `MCPClientManager.get_tools('render-html')` discovers both `render_html` and `render_html_from_markdown`.
- `cd dashboard/frontend && npm run build` completes successfully.

No further blocking findings.

PR #22, opened 2026-05-14
Chat: streaming heartbeat + tool_call_pending events

### LINKED ISSUE #20: Chat: streaming UI gives no feedback during silent backend phases
## Problem

While the model is generating, the chat UI can sit silent for many seconds with no indication that the stream is still alive. Observed phases where this happens:

1. **Prompt prefill** — large prompts (e.g. a 5k-token CSV) take seconds to prefill before the first token; UI shows nothing.
2. **Reasoning before content** — Qwen 3 commonly emits a long reasoning block, then a tool call, with no `delta.content`. The reasoning panel only appears once reasoning tokens land.
3. **Tool-call assembly** — `dashboard/chat/llm_proxy.py:138` only emits the `tool_calls` SSE event when `finish_reason == "tool_calls"`, so the user sees nothing while the model is constructing the call.
4. **Between tool-call rounds** — multi-round flows have multi-second gaps between rounds where the UI looks frozen.

Reproduces reliably on conversation 4384d781-f444-43c4-ac9b-4964b3417250 (CSV → render_html, 4 tool calls).

## Goal

Continuous "the stream is alive" feedback so the user can tell the difference between *the model is working* and *the connection is dead*.

## Proposed approach (two layers)

1. **Heartbeat event**. Wrap the SSE generator in a queue-based pump that yields `("heartbeat", {elapsed_s})` whenever no upstream event arrived in the last ~3s. Works for *any* silence: prefill, reasoning, tool-call assembly, between-round gaps.
2. **`tool_call_pending` event**. As soon as the first delta carrying a tool name fragment arrives, emit `("tool_call_pending", {index, name})`. Frontend renders a placeholder pill that morphs into the full panel when the call completes (`finish_reason == "tool_calls"`).

## Frontend

Small pulsing dot + "Waiting for model… 4.5s" near the bottom-most streaming element. Resets the counter on any real event. Replaces the existing static "Thinking…" placeholder in `MessageList.jsx:135` with a heartbeat-driven counter.

## Out of scope

- Token-rate telemetry / progress bars.
- Phase classification beyond what's needed for the placeholder text.

## Files likely touched

- `dashboard/chat/routes.py` — wrap inner generator with heartbeat pump.
- `dashboard/chat/llm_proxy.py` — emit `tool_call_pending` on first name-fragment.
- `dashboard/chat/tool_loop.py` — pass through the new events.
- `dashboard/frontend/src/services/sse.js` — parse `heartbeat` / `tool_call_pending`.
- `dashboard/frontend/src/components/chat/MessageList.jsx` — render heartbeat indicator + pending pill.

teo-mateo:
Closes #20.

## What changed

Two new SSE events flow through the chat stream so the UI can show that the model is still working during phases that previously rendered as silence.

### Backend

- `llm_proxy.py` — when a streaming chunk first carries a name fragment for a given tool-call index, emit `tool_call_pending {index, name}`. The full `tool_calls` event still only fires at `finish_reason`, but the UI now has something to render in the gap.
- `routes.py` — new `_with_heartbeat()` helper wraps the inner stream in a queue-fed worker thread. The outer generator pops with a 3s timeout; on timeout it yields a `heartbeat {elapsed_s}` SSE event. The idle counter resets on any real upstream event. Daemon thread, so client disconnect leaves the model to finish its turn naturally (the same posture that backs partial-save in #12).
- `tool_loop.py` — passes `tool_call_pending` through transparently.

### Frontend

- `sse.js` — parses `heartbeat` and `tool_call_pending` into `onHeartbeat` / `onToolCallPending` callbacks.
- `useChat.js` — tracks `heartbeat` and `pendingToolCalls` state; clears both on any real event, `message_saved`, error, and `stopStreaming`.
- `MessageList.jsx` — dashed-border pill while a tool call is still assembling (replaced by the full panel when `finish_reason==tool_calls`); pulsing blue dot + "Waiting for model… 4.5s" replaces the static "Thinking…" fallback, and shows as a thin ribbon below partial content during multi-round gaps.

## Test plan

- [ ] Send a large-prompt (5k+ token) message and watch for "Waiting for model… Ns" counting up during prefill.
- [ ] Use a Qwen 3 reasoning model that emits a long reasoning block before a tool call — pending pill should appear before the full panel.
- [ ] Trigger a multi-round tool flow (CSV → render_html) and confirm the heartbeat ribbon shows during between-round silence.
- [ ] Disconnect the client mid-stream and confirm partial-save still works.
- [ ] Confirm no regression on a plain non-tool conversation (heartbeat appears only when truly idle for 3s).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,3 +86,6 @@ htmlcov/
 .codex/
 .codex
 opencode.json
+
+# test harness output
+tools/traces/
--- a/dashboard/chat/constants.py
+++ b/dashboard/chat/constants.py
@@ -33,7 +33,11 @@ Tool use
   reasoning. If you have nothing to add, write a brief direct answer rather than
   emitting empty output.
 - Hard cap: at most 5 tool calls per question. If you hit the cap without a confident
-  answer, answer with what you have and flag the uncertainty."""
+  answer, answer with what you have and flag the uncertainty.
+- When constructing web search queries about current state of anything (prices, products,
+  events, rankings, "best X in Y"), use the year from the `Current date:` line below —
+  not years from your training data. Do not append "2024" or "2025" to a query unless
+  the user specifically asked about that year."""
 
 DEFAULT_SIDEKICK_SYSTEM_PROMPT = (
     "You are a critical technical reviewer. When asked to critique a response, "
--- a/dashboard/chat/constants.py
+++ b/dashboard/chat/constants.py
@@ -6,8 +6,8 @@ of it.
 
 Style
 - Match length to the task. Short questions get short answers. No preamble ("Great question!",
-  "Sure, here's…"), no trailing summaries of what you just said.
-- Use markdown: fenced code blocks with a language tag, inline `code` for symbols and paths,
+  "Sure, here's…"), no trailing summary of what you just said.
+- Use markdown: fenced code blocks with a language tag, inline backticks for symbols and paths,
   tables when comparing, lists when enumerating. Plain prose otherwise.
 - For code, show the minimum that answers the question. Don't pad with imports, boilerplate,
   or "here's how to run it" unless that's the question.
@@ -19,11 +19,21 @@ Accuracy
 - When the user shows you code or output, read it carefully before answering. Quote the
   specific line you're reacting to.
 
-Tools
-- When tools are available (their usage is described below), prefer calling them over
-  describing what they would do. Always invoke tools by their full namespaced name
-  (`<server_id>__<tool_name>`).
-- Don't narrate tool calls before making them ("Let me search…"). Just call."""
+Tool use
+- Tools shown below are real and connected. When the task asks for current data
+  (web pages, prices, recent events, exact specs) and a relevant tool is available,
+  use it. Do not fabricate results from memory under the guise of "simulating" a search.
+- Call exactly ONE tool per turn. Wait for the result before deciding the next call.
+  Do not pre-plan a batch of calls — let each result guide what to do next.
+- Issue every tool call through the structured tool-calling interface. Never write
+  tool invocations as text, code blocks, or angle-bracket tags; never echo a tool's
+  raw schema back to the user.
+- When you have enough information to answer, stop calling tools and write the answer
+  as plain text. The user only sees that final text — not your tool calls, not your
+  reasoning. If you have nothing to add, write a brief direct answer rather than
+  emitting empty output.
+- Hard cap: at most 5 tool calls per question. If you hit the cap without a confident
+  answer, answer with what you have and flag the uncertainty."""
 
 DEFAULT_SIDEKICK_SYSTEM_PROMPT = (
     "You are a critical technical reviewer. When asked to critique a response, "
--- a/dashboard/chat/db.py
+++ b/dashboard/chat/db.py
@@ -99,6 +99,7 @@ class ChatDB:
             ("conversations", "selected_text", "ALTER TABLE conversations ADD COLUMN selected_text TEXT"),
             ("conversations", "mcp_servers_json", "ALTER TABLE conversations ADD COLUMN mcp_servers_json TEXT"),
             ("messages", "tool_calls_json", "ALTER TABLE messages ADD COLUMN tool_calls_json TEXT"),
+            ("messages", "parse_warning_json", "ALTER TABLE messages ADD COLUMN parse_warning_json TEXT"),
         ]
         for table, column, sql in migrations:
             try:
@@ -266,6 +267,7 @@ class ChatDB:
             model_service=row["model_service"],
             images_json=row["images_json"],
             tool_calls_json=row["tool_calls_json"],
+            parse_warning_json=row["parse_warning_json"] if "parse_warning_json" in row.keys() else None,
             seq=row["seq"],
             created_at=row["created_at"],
         )
@@ -313,10 +315,11 @@ class ChatDB:
         try:
             conn.execute(
                 """INSERT INTO messages
-                   (id, conversation_id, role, content, reasoning_content, model_service, images_json, tool_calls_json, seq)
-                   VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)""",
+                   (id, conversation_id, role, content, reasoning_content, model_service, images_json, tool_calls_json, parse_warning_json, seq)
+                   VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)""",
                 (msg.id, msg.conversation_id, msg.role, msg.content,
-                 msg.reasoning_content, msg.model_service, msg.images_json, msg.tool_calls_json, msg.seq),
+                 msg.reasoning_content, msg.model_service, msg.images_json, msg.tool_calls_json,
+                 msg.parse_warning_json, msg.seq),
             )
             conn.commit()
             return self.get_message(msg.id)
--- a/dashboard/chat/llm_proxy.py
+++ b/dashboard/chat/llm_proxy.py
@@ -1,9 +1,61 @@
 import json
 import logging
+import re
 
 import requests
 
 logger = logging.getLogger(__name__)
+
+
+# Known wrong-format markers Qwen3.6 (and similar) emit when the structured
+# tool-call format drifts. The qwen3_coder parser strips the canonical
+# `<tool_call>…</tool_call>` block, but if the model writes any of these
+# instead, the garbage leaks into `content` (or content ends up empty after
+# the parser tries to consume a malformed call). Detecting these gives the
+# UI a chip to surface the failure ad-hoc rather than rendering an empty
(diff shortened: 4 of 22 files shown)


teo-mateo:

- `dashboard/chat/routes.py:157` / `dashboard/chat/routes.py:169`: `_with_heartbeat()` starts an unbounded background producer, but nothing tells that producer to stop when the SSE generator is closed. When the client disconnects or the frontend aborts the fetch, `_stream_response()` exits from a `yield`, saves the partial, and drops the consumer side; the producer thread still keeps iterating `stream_chat_completion()` / `stream_with_tools()` and pushing into `q`. I reproduced this locally with a slow infinite iterator: after `g.close()`, the upstream counter kept advancing (`2 -> 9` after 80ms). In production this can continue burning an LLM request after Stop/navigation and, for tool-enabled conversations, can still reach `mcp_manager.call_tool(...)` in `tool_loop.py:74` after the client is gone. It also leaves an unbounded queue with no consumer, so a long completion after disconnect can grow memory until the upstream request finishes. Please add cancellation/backpressure to the wrapper, e.g. a stop event checked by the producer, explicit `it.close()` when the outer generator closes, and a bounded queue/put timeout so disconnected streams cannot keep accumulating events or execute later tool calls.

teo-mateo:

**Finding 1.1 — `_with_heartbeat()` leaks producer thread on client disconnect**: accepted.

Fixed in commit b79bfbb. Concretely:

- `queue.Queue` is now bounded (`HEARTBEAT_QUEUE_MAX=64`) — caps memory growth after disconnect even if the producer keeps running for a while.
- New `threading.Event` `stop`. The consumer's `finally` sets `stop` and drains the queue. The drain is necessary: a producer parked at `q.put()` on a full queue wouldn't notice `stop` until its 0.5s timeout fires, and would process its next event in the meantime.
- Producer checks `stop` between events and on every queue-full retry. Once it notices, it calls `it.close()` and exits.
- `it.close()` is what actually short-circuits further tool execution — it propagates `GeneratorExit` into `stream_with_tools` / `stream_chat_completion` at their next yield, unwinding the for-loop in `stream_with_tools` before it reaches the next `mcp_manager.call_tool()`.
- Closing is done from the **producer thread**: the generator protocol forbids closing a generator that's running on another thread, so the consumer's finally only signals stop and leaves the actual close to the producer's finally.

Documented one residual caveat in the docstring: cancellation is cooperative. A producer blocked inside `requests.iter_lines` on a socket read won't notice `stop` until the next chunk arrives, and an `mcp_manager.call_tool()` already in flight runs to completion (those subprocess calls aren't interruptible from here). True interruption would need a cancel token threaded through `stream_chat_completion` / `stream_with_tools`; that's a separate refactor I'll track as a follow-up.

Verified with the same slow-infinite-iterator pattern codex used: counter held at 2 after 600ms (vs codex's repro of 2 → 9 in 80ms on the old code), inner generator's `finally` ran, no leftover threads.

teo-mateo:

- `dashboard/chat/routes.py:213` / `dashboard/chat/routes.py:407`: `_with_heartbeat()` can drop its terminal sentinel when the queue is full, and `_stream_response()` relies on that sentinel because the `done` branch yields `[DONE]`/`message_saved` but does not return. A slow client or stalled WSGI send can let the producer fill all 64 queue slots after the consumer has yielded one event; if the final real event fills the queue, the producer's `q.put(sentinel, timeout=0.5)` hits `queue.Full` and silently `pass`es. The consumer later drains through `done`, saves the assistant message, then waits forever and emits heartbeat events after completion, so the fetch never resolves and the title-tail / controller cleanup can hang. I reproduced this locally with `_with_heartbeat()` by consuming one event, sleeping long enough for the producer to fill `HEARTBEAT_QUEUE_MAX`, then draining: `done` arrived but `StopIteration` never did, only `__heartbeat__` events. Please make termination non-droppable, e.g. always enqueue the sentinel by freeing a slot / using an out-of-band done flag, or return from `_stream_response()` immediately after yielding `message_saved` so completion does not depend on the wrapper's sentinel.

teo-mateo:

**Finding 2.1 — `_with_heartbeat()` can drop its terminal sentinel on a full queue, hanging the stream**: accepted.

Confirmed real. Reproduced the exact scenario you described: consume one event, stall the consumer long enough for the producer to fill all `HEARTBEAT_QUEUE_MAX` slots and run its `finally`, where `q.put(sentinel, timeout=...)` raises `queue.Full` and the `except: pass` drops the sentinel — the consumer then loops on `queue.Empty` emitting `__heartbeat__` forever; `[DONE]`/`message_saved` are already yielded but the generator never raises `StopIteration`, so the request never completes.

Fixed in commit 949bad0. Added an out-of-band `finished = threading.Event()`, set unconditionally in the producer's `finally` *before* the (still best-effort) sentinel put. The consumer's `queue.Empty` branch now returns when `finished` is set instead of yielding a heartbeat, so termination no longer depends on the droppable sentinel. The sentinel is kept purely as a fast-path for prompt shutdown. No events are lost: the `Empty` branch only fires once the queue is fully drained, so any real events enqueued before `finished` are still delivered first.

Verification: inner iterator emitting `HEARTBEAT_QUEUE_MAX + 5` events, consumer stalls 2s so the sentinel is provably dropped — generator now delivers all 69 events and terminates cleanly (StopIteration), no infinite heartbeats.

teo-mateo:

- `tools/test_no_tools.py:27` / `tools/test_search_iterative.py:34`: the new harness scripts commit the real local model-service API key (`llmd-cfc6b6ef75620adc289764238a831f10cb21`) into the repo. That key otherwise comes from ignored local config (`services.json` / generated compose) and is what protects the OpenAI-compatible model endpoints; publishing it means anyone who can reach those exposed ports can reuse the key, and the scripts also become non-portable for users with a different `LLM_DOCK_API_KEY`. Please remove the literal secret and read it from an env var/CLI arg instead, with no default secret value.

teo-mateo:

**Finding 3.1 — committed model-service API key in harness scripts**: accepted.

Fixed in commit 2899f1e:
- `tools/test_no_tools.py`: reads `LLM_DOCK_API_KEY` from the environment; `sys.exit`s with a clear message if unset. No default secret.
- `tools/test_search_iterative.py`: `DEFAULT_API_KEY = os.environ.get("LLM_DOCK_API_KEY")` (no literal), `--api-key` still overrides; `argparse` now errors out if neither is provided.

Verified both compile and that the literal no longer appears anywhere under `tools/`.

⚠️ Important caveat I want to flag rather than hide: removing the literal from the working tree does **not** scrub it from this branch's git history — it remains in commit `189e976` and will persist after a `--merge`. The key (`api_key` in `services.json`) should be **rotated** regardless of how this PR lands, since it was pushed to the remote. Rotating is the real remediation; the code change just prevents re-leaking it going forward. Let me know if you want a history rewrite (`git filter-repo`) on this branch before merge instead of/in addition to rotation.

teo-mateo:

- `dashboard/chat/routes.py:366` / `dashboard/chat/routes.py:397`: `_stream_response()` accumulates every reasoning delta in `accumulated_reasoning`, but the assistant row saved on `done` uses only `data["reasoning_content"]` from the final model round. In a tool-enabled turn, any reasoning emitted before a `finish_reason=tool_calls` event is streamed to the UI and, with this PR's continuous `streamingReasoning` display, shown as part of the assistant's thinking trace; as soon as `message_saved` triggers the DB refetch, that pre-tool reasoning is gone because it was never persisted. This loses history and makes the visible reasoning block shrink/disappear after save for multi-round tool flows. Please persist the accumulated reasoning for the final assistant message (or otherwise store the per-round reasoning) instead of only the final round's `done` payload.

teo-mateo:

**Finding 4.1 — only final-round reasoning persisted in multi-round tool flows**: accepted.

Correct. `_stream_response()` accumulates every reasoning delta across all rounds in `accumulated_reasoning`, but the `done` save used `data["reasoning_content"]`, which `llm_proxy` populates from only the last HTTP round. With this PR's continuous reasoning display, the pre-tool-call reasoning is shown live and then disappears when `message_saved` triggers the refetch.

Fixed in commit ce8c8ca: the `done` save now persists `(accumulated_reasoning or data["reasoning_content"]) or None`. `accumulated_reasoning` is the exact cross-round trace the user saw streamed; it falls back to the done payload for the single-round case where the two are identical, and to `None` when there is no reasoning at all. Content is unchanged — intermediate tool-call rounds emit no content deltas, so `data["content"]` already equals the full answer.

Compiles clean.

teo-mateo:

I reviewed the current head (`ce8c8ca`) against the PR diff and prior discussion, with the resolved iteration 1-4 items excluded. I do not have additional correctness, security, API-contract, or production-relevant performance findings to raise.

Local verification note: I attempted `pytest dashboard/tests/test_partial_save.py -q`, but this environment is missing Flask, so collection failed with `ModuleNotFoundError: No module named 'flask'` before tests could run.

PR #26, opened 2026-05-17
feat(security): one-click default API key rotation (#25)

### LINKED ISSUE #25: FEATURE: one-click default key rotation.
On a single click the default api key should be rotated; change should apply to ALL MODELS in services.json and in the docker-compose file. Running models should be STOPPED. A warning should be shown to the user IF there are models which are going to be stopped. 

Question: what do we need to do about open-webui?

teo-mateo:
Closes #25.

## What

One click rotates the shared **default API key** across the whole stack.

- Generates a new key, persists it to `.env` (`LLM_DOCK_API_KEY`), and applies it to **every** service in `services.json` + regenerated `docker-compose.yml`.
- **Stops** running affected containers — they keep the revoked key in memory until restarted, so they must be stopped. Restart is left to the user (per issue requirement).
- A pre-rotation **preview/warning** lists:
  - running models that will be stopped, and
  - Open WebUI-registered services that will need a manual key re-entry.

## Open WebUI decision (the issue's open question)

**Warn-only.** Open WebUI stores its own copy of each connection's API key in its sqlite DB; rotation does not touch it. Affected registered services are surfaced in both the pre-warning and the result so the user knows to update them under *Admin → Settings → Connections*.

## Changes

**Backend**
- `config.set_global_api_key()` — atomic `.env` writer + in-process update; `DOTENV_PATH` resolved next to the module.
- `key_rotation.rotate_keys_in_db()` — pure, unit-tested mutation of `services.json` + compose rebuild.
- `GET /api/default-api-key/rotation-preview` and `POST /api/default-api-key/rotate`.
- `GLOBAL_API_KEY` references switched to `config.GLOBAL_API_KEY` so the rotated value is visible without a process restart.

**Frontend**
- `RotateDefaultKeyModal` — impact warnings (red: will be stopped / amber: Open WebUI) and a copyable new key.
- "Rotate default key" button in the services table header.

**Drive-by fix**
- `TokenSparkline.test.jsx`: relaxed a stale assertion (label/value are separate spans spaced via CSS flex gap, so `textContent` has no literal space between them). This test was failing on `main`; unrelated to the feature but fixed here so the branch is green.

## Testing

- New `dashboard/tests/test_key_rotation.py` (4 tests).
- Full backend suite: **147 passed**.
- Frontend: **36 passed** (was 35 + 1 pre-existing failure, now fixed).
- Frontend builds clean.

## Notes

- A pre-existing unrelated local modification to `install-service.sh` was deliberately **not** included in this PR.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -86,6 +86,35 @@ entries — but it also means a **typo** in the flag name (e.g. `--gpu-mem`)
 ships to the container as-is and fails at startup, not at config time.
 Verify with `mgr.preview_service(name)` before bringing the container up.
 
+## Default API key rotation
+
+All model services share one **default API key**: `LLM_DOCK_API_KEY` in
+`dashboard/.env` (exposed as `config.GLOBAL_API_KEY`), mirrored into every
+service's `api_key` in `services.json`. The dashboard can rotate it in one
+click (Services table → "Rotate default key"):
+
+- `GET /api/default-api-key/rotation-preview` — impact only, mutates
+  nothing: counts services, lists running containers (will be stopped) and
+  Open WebUI-registered services (keys go stale).
+- `POST /api/default-api-key/rotate` — generates a new key, writes it to
+  `.env` via `config.set_global_api_key()`, rewrites every `services.json`
+  entry, rebuilds `docker-compose.yml`, then **stops** every running
+  affected container (they keep the revoked key in memory; they are not
+  auto-restarted — start them manually afterwards).
+
+Gotchas:
+
+- Reference `config.GLOBAL_API_KEY` (the module attribute), not a
+  `from config import GLOBAL_API_KEY` name binding — only the attribute
+  reflects an in-process rotation without a dashboard restart.
+- **Open WebUI is not touched.** It stores its own copy of each
+  connection's key in its sqlite DB; after rotation those are stale and
+  must be re-entered manually under Open WebUI → Admin → Settings →
+  Connections. The preview/rotate responses list the affected services.
+- Core rotation logic is `key_rotation.rotate_keys_in_db()` (pure, unit
+  tested in `dashboard/tests/test_key_rotation.py`); container stops and
+  preview live in `routes/services.py`.
+
 ## Adding an external MCP server
 
 Built-in MCP tools live in `dashboard/chat/mcp_registry.py` and ship with
--- a/dashboard/benchmarking/routes.py
+++ b/dashboard/benchmarking/routes.py
@@ -3,6 +3,7 @@ import os
 from flask import Blueprint, jsonify, request, current_app
 
 from auth import require_auth
+from db_lock import serialize_db
 from .db import BenchmarkDB
 from .executor import BenchmarkExecutor
 from .validators import validate_params, validate_service_name, BENCHMARK_ONLY_FLAGS
@@ -137,6 +138,7 @@ def delete_benchmark(run_id):
 
 @benchmarks_bp.route("/api/benchmarks/<run_id>/apply", methods=["PUT"])
 @require_auth
+@serialize_db
 def apply_benchmark(run_id):
     db = current_app.config["BENCHMARK_DB"]
     run = db.get_run(run_id)
--- a/dashboard/compose_manager.py
+++ b/dashboard/compose_manager.py
@@ -5,6 +5,7 @@ Docker Compose file manager with safe atomic updates.
 import os
 import shutil
 import subprocess
+import tempfile
 import fcntl
 import yaml
 import json
@@ -408,9 +409,27 @@ class ComposeManager:
             return {}
 
     def _save_services_db(self, services: Dict[str, Any]):
-        """Save services database to JSON file"""
-        with open(self.services_db_path, "w") as f:
-            json.dump(services, f, indent=2)
+        """Save services database to JSON file atomically.
+
+        Write to a same-directory temp file, fsync, then ``os.replace()`` so
+        an interrupted/failed write can never leave a truncated or partial
+        ``services.json`` (which ``_load_services_db`` would silently treat as
+        ``{}``, dropping every dynamic service on the next compose rebuild).
+        """
+        directory = self.services_db_path.parent
+        fd, tmp_path = tempfile.mkstemp(
+            prefix=".services.", suffix=".json.tmp", dir=str(directory)
+        )
+        try:
+            with os.fdopen(fd, "w") as f:
+                json.dump(services, f, indent=2)
+                f.flush()
+                os.fsync(f.fileno())
+            os.replace(tmp_path, self.services_db_path)
+        except Exception:
+            if os.path.exists(tmp_path):
+                os.unlink(tmp_path)
+            raise
 
     def save_services_db(self, services: Dict[str, Any]):
         """Replace the entire services database in one write.
--- a/dashboard/compose_manager.py
+++ b/dashboard/compose_manager.py
@@ -412,6 +412,16 @@ class ComposeManager:
         with open(self.services_db_path, "w") as f:
             json.dump(services, f, indent=2)
 
+    def save_services_db(self, services: Dict[str, Any]):
+        """Replace the entire services database in one write.
+
+        Used by atomic multi-service operations (e.g. default API key
+        rotation) that must commit all entries together rather than
+        per-service, so a mid-operation failure can't leave services.json
+        half-rotated.
+        """
+        self._save_services_db(services)
+
     def add_service_to_db(self, service_name: str, config: Dict[str, Any]):
         """Add service to database"""
         services = self._load_services_db()
--- a/dashboard/config.py
+++ b/dashboard/config.py
@@ -1,8 +1,10 @@
 import os
 import logging
-from dotenv import load_dotenv
+from dotenv import load_dotenv, set_key
 
-load_dotenv()
+# Resolve .env next to this file so it works regardless of CWD.
+DOTENV_PATH = os.path.join(os.path.dirname(os.path.abspath(__file__)), ".env")
+load_dotenv(DOTENV_PATH)
 
 DASHBOARD_PORT = int(os.getenv("DASHBOARD_PORT", 3305))
 DASHBOARD_HOST = os.getenv("DASHBOARD_HOST", "127.0.0.1")
@@ -11,6 +13,19 @@ LOG_LEVEL = os.getenv("LOG_LEVEL", "INFO")
 COMPOSE_FILE = os.getenv("COMPOSE_FILE", "../docker-compose.yml")
 COMPOSE_PROJECT = os.getenv("COMPOSE_PROJECT_NAME", "dockerized-models")
 GLOBAL_API_KEY = os.getenv("LLM_DOCK_API_KEY")
+
+
+def set_global_api_key(new_key: str, dotenv_path: str = DOTENV_PATH):
+    """Persist a new global API key to .env and update the in-process value.
+
+    Other modules must reference ``config.GLOBAL_API_KEY`` (not a name imported
+    via ``from config import GLOBAL_API_KEY``) to observe the rotated value
+    without a process restart.
+    """
+    global GLOBAL_API_KEY
+    set_key(dotenv_path, "LLM_DOCK_API_KEY", new_key)
+    GLOBAL_API_KEY = new_key
+    return new_key
 
 
 _config_initialized = False
--- a/dashboard/db_lock.py
+++ b/dashboard/db_lock.py
@@ -1,0 +1,33 @@
+"""Shared lock serializing every services.json read-modify-write + compose
+rebuild, across *all* blueprints (services routes and benchmarking routes).
+
+Flask's dev server runs threaded (``app.run`` defaults ``threaded=True``), so
+handlers in different blueprints that each do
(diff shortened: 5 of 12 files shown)


teo-mateo:

- `dashboard/routes/services.py:893` updates `.env` / `config.GLOBAL_API_KEY` before `rotate_keys_in_db()` rewrites `services.json` and validates/regenerates compose, while `dashboard/key_rotation.py:38` writes each service one at a time before the compose rebuild at line 44. If any later step fails (for example compose validation times out/fails, a malformed existing service cannot render, or a mid-loop DB write fails), the endpoint returns 500 after already changing at least the in-process/default key, and potentially a subset/all of `services.json`, while `docker-compose.yml` may have been rolled back and the old running containers were never stopped. That leaves the dashboard reporting the new default key while live services still accept the old key, and future starts/rebuilds can pick up different keys depending on which file they read. The rotation should be transaction-like: stage/validate the new service DB + compose first, then commit `.env` and service DB together, or roll back `.env`/`config.GLOBAL_API_KEY` and any partial DB writes on failure. A route-level failure test would catch this; the new tests only cover the pure helper and happy path persistence.

teo-mateo:

**Finding 1 — Non-atomic rotation leaves .env/key advertising a new key while services still accept the old one**: accepted.

Fixed in commit 00bf65f.

- `key_rotation.rotate_keys_in_db()` now writes `services.json` in a **single replace** (new `ComposeManager.save_services_db()`) rather than one service at a time, and **restores the pre-rotation `services.json`** if `rebuild_compose_file()` raises. Since `rebuild_compose_file()` already rolls `docker-compose.yml` back to its backup on failure, the two files stay mutually consistent and on the old key.
- `rotate_default_api_key` reordered so the riskiest, validated work (services.json + compose regeneration) is committed **first**; `.env` / `config.GLOBAL_API_KEY` is committed **only after** that succeeds. If the `.env` commit itself fails, the services.json + compose changes are rolled back from a pre-mutation snapshot. This eliminates the window where the dashboard reports the new key while live services still authenticate with the old one (and vice versa).
- Added tests: `test_rotate_rolls_back_services_json_on_compose_failure` (pure helper) and `test_route_does_not_commit_env_before_services_and_compose` (route-level — asserts a 500 on rebuild failure leaves `config.GLOBAL_API_KEY` and `services.json` untouched, and that the `.env` writer is never reached). Full backend suite: 149 passed.

teo-mateo:

- `dashboard/routes/services.py:891`-`908` performs the rotation as an unlocked sequence across three separate stores: `services.json`/compose via `rotate_keys_in_db()`, then `.env`/`config.GLOBAL_API_KEY` via `config.set_global_api_key()`. Two concurrent `POST /api/default-api-key/rotate` requests can interleave so request A writes services/compose with key A, request B writes services/compose with key B, then A writes `.env`/in-process key A after B has already committed the service files. The endpoint can return success with `.env`/dashboard reporting key A while `services.json` and `docker-compose.yml` contain key B, which breaks the core contract that the default key is mirrored everywhere. This can happen from a double-click/retry or two dashboard sessions, and the frontend disable only protects one component instance. Please serialize the entire rotate operation (for example a process-wide lock plus the existing file lock, or a file lock held from the first DB snapshot through `set_global_api_key`) and add a two-thread/barrier test for this interleaving.

- `dashboard/compose_manager.py:415`-`423` says `save_services_db()` replaces the services database for atomic multi-service operations, but it delegates to `_save_services_db()`, which opens `services.json` with `"w"` and streams `json.dump()` directly (`dashboard/compose_manager.py:410`-`413`). If that write is interrupted or raises after truncation, rotation leaves a corrupt/partial `services.json` while the old compose file may remain in place; `_load_services_db()` then treats invalid JSON as `{}`, so subsequent rebuilds can drop all dynamic services. This is the new commit point for key rotation, so it should use the same temp-file + `os.replace()` pattern as compose updates (ideally with fsync and same-directory temp), with a test that forces a mid-write failure and asserts the original DB bytes remain intact.

teo-mateo:

**Finding 2.1 — Concurrent rotations can interleave services/compose vs .env writes**: accepted. Fixed in commit 8be7a0c.

The entire rotate operation now runs under a process-wide `threading.Lock` (`_ROTATION_LOCK` in `routes/services.py`), held from the pre-rotation snapshot through the `.env` commit and the container-stop loop. The dashboard runs as a single process (systemd unit: `python app.py`), so a process-wide lock fully serializes rotation — there is no multi-worker deployment, so the heavier cross-process file-lock alternative isn't required. Added `test_concurrent_rotations_keep_stores_consistent`: two threads POST `/api/default-api-key/rotate` concurrently (with a widened interleave window between the services/compose commit and the `.env` commit); the test asserts `services.json`, `config.GLOBAL_API_KEY`, and `.env` all agree on the final key.

**Finding 2.2 — `save_services_db` truncate-then-write can corrupt services.json**: accepted. Fixed in commit 8be7a0c.

`_save_services_db` now writes to a same-directory temp file, `flush()` + `os.fsync()`, then `os.replace()` (the same atomic pattern compose updates already use). An interrupted/failed write leaves the original `services.json` byte-for-byte intact and cleans up the temp file. Added `test_save_services_db_atomic_on_write_failure`: forces `os.replace` to raise and asserts the original bytes are unchanged with no `.services.*.json.tmp` leftovers.

Full backend suite: 151 passed.

teo-mateo:

- `dashboard/key_rotation.py:45`-`56` rotates a stale snapshot of the entire services DB and writes it back wholesale, while the new route lock in `dashboard/routes/services.py:898` only excludes other rotate requests. Existing mutating endpoints like create (`dashboard/routes/services.py:263`-`264`), update (`335`-`338`), delete (`384`-`387`), and set-global-key (`823`-`825`) can still run concurrently under Flask’s threaded dev server. A realistic interleave is: rotate reads `{A}`; create adds `B` and rebuilds successfully; rotate then saves its old snapshot as `{A with new key}` and rebuilds, silently dropping `B` even though the create request returned success. The inverse can also leave a just-updated service with the old key after a successful rotation. Please serialize rotation with all service DB read-modify-write operations, ideally by holding one shared/file lock across the DB snapshot, DB replace, compose rebuild, and `.env` commit, and add a rotate-vs-create or rotate-vs-update regression test.

teo-mateo:

**Finding 3.1 — Rotation lock only excludes other rotations; concurrent create/update/delete/etc. can be dropped by rotation's wholesale snapshot write**: accepted. Fixed in commit e978adc.

Confirmed the premise: Flask `app.run()` defaults `threaded=True`, so the other mutating endpoints really can run concurrently with rotation under the dev server.

- Renamed `_ROTATION_LOCK` → `_SERVICES_DB_LOCK` and made it a re-entrant `threading.RLock` (a handler that nests guarded helpers on the same thread can't self-deadlock).
- Added a `_serialize_db` decorator, placed **below** `@require_auth` so unauthenticated requests are rejected without contending for the lock, and applied it to `create_service`, `update_service`, `delete_service`, `rename_service`, `set_public_port`, and `set_service_global_api_key`. `rotate_default_api_key` keeps its explicit `with _SERVICES_DB_LOCK:` block. All of these now serialize on the one shared lock across the DB snapshot → DB replace → compose rebuild (→ `.env` commit, for rotation).
- This also closes the pre-existing create-vs-create / update-vs-delete races in the same module, not just the rotation interleave.
- Added `test_rotation_does_not_drop_concurrent_create`: starts a rotation (critical section widened via a sleeping `set_global_api_key`), then fires a concurrent `POST /api/services`; asserts both the rotated pre-existing service and the concurrently-created service survive in `services.json`, and the pre-existing one mirrors `GLOBAL_API_KEY`.

Full backend suite: 152 passed.

teo-mateo:

- `dashboard/benchmarking/routes.py:149`-`170` still performs an unlocked read-modify-write of a full service config in `/api/benchmarks/<run_id>/apply`. The new `_SERVICES_DB_LOCK` only wraps handlers in `routes/services.py`, so this can still interleave with rotation: benchmark apply loads `service_config` with the old `api_key`, rotation commits the new key to `services.json`/compose/`.env`, then apply writes its stale full config at line 169 and rebuilds at line 170. Both requests can return success while that service is back on the old key and `config.GLOBAL_API_KEY`/`.env` hold the new key. This is the same class of race fixed for create/update/delete; the benchmark apply mutation needs to share the same lock (or the lock needs to live below both blueprints, e.g. around ComposeManager DB mutations), with a rotate-vs-apply regression test.

- `dashboard/routes/services.py:947`-`975` returns HTTP 200 with `"success": true` even when `control_service(name, "stop")` fails for a running affected container. In that case the security-critical part of rotation did not complete: the container keeps accepting the old key in memory, but API clients and the modal still get a successful “Default API Key Rotated” result plus the new key. Callers that key off `success` will treat the old key as revoked when it is not. Please make nonempty `stop_errors` a non-success/partial-failure response (or otherwise encode `success: false` for that path) and add a route test that forces a stop failure for a running affected service.

teo-mateo:

**Finding 4.1 — benchmark `apply` does an unlocked read-modify-write that can revert a concurrent rotation**: accepted. Fixed in commit 70e1969.

The lock was living in `routes/services.py`, so only that blueprint contended on it. Moved `SERVICES_DB_LOCK` + the `serialize_db` decorator into a new neutral module `db_lock.py` that sits below all blueprints. `routes/services.py` now imports them (aliased to `_SERVICES_DB_LOCK`/`_serialize_db` to avoid churn), and `benchmarking/routes.py` applies `@serialize_db` (below `@require_auth`) to `apply_benchmark`. Apply's `update_service_in_db` + `rebuild_compose_file` now serialize on the same lock as rotation/create/etc. Added `test_shared_lock_is_one_object_across_blueprints` (identity check) and `test_rotation_not_reverted_by_concurrent_benchmark_apply` (real two-thread rotate-vs-apply regression: asserts the service keeps the rotated key *and* the applied benchmark param).

**Finding 4.2 — rotation returned `success: true` / 200 even when a container stop failed**: accepted. Fixed in commit 70e1969.

When any running affected container fails to stop, the response is now `success: false`, `partial: true`, **HTTP 207**, with a message naming the containers that still accept the OLD key in memory. I used 207 (still 2xx) deliberately so the frontend `fetchAPI` doesn't throw and discard the body — the modal still surfaces the new key and the per-container stop errors, and its header switches to "Default API Key Rotated — Action Required" when `success === false`. Programmatic callers keying off `success` or a strict-200 check now correctly see this as a non-success/partial outcome. Added `test_rotate_partial_failure_when_container_stop_fails` (forces a stop failure, asserts 207 + `success false` + `partial true` + the key still rotated in files).

Backend suite: 155 passed. Frontend: 36 passed, build clean.

teo-mateo:

- `dashboard/routes/services.py:121` lets `/api/services/<name>/start` run outside the shared services DB/rotation lock, while `rotate_default_api_key()` snapshots `running` once at `dashboard/routes/services.py:903`-`930` and only stops names from that snapshot. A stopped service can therefore be started concurrently after the snapshot but before `rotate_keys_in_db()` replaces `docker-compose.yml`; `control_service(..., \"start\")` invokes `docker compose up` using the old compose file (`dashboard/docker_utils.py:225`-`226`), so the container comes up with the old API key in memory. Because it was not in the rotation snapshot, the stop loop never stops it, and the endpoint can return success even though a model is still serving the revoked key. Please serialize start/restart control with the same lock (or re-check and stop all affected running containers after the key/compose commit), and add a rotate-vs-start regression test for a previously stopped service.

teo-mateo:

**Finding 5.1 — `/api/services/<name>/start` runs outside the lock; a stopped service started mid-rotation comes up on the revoked key and is never stopped**: accepted. Fixed in commit 92242cf.

This is the same class as the create/apply races and the analysis is correct: `rotate` snapshots `running` once and only stops that set, while `control_service(name, "start")` shells out to `docker compose up` against whatever `docker-compose.yml` exists at exec time. A start interleaved between the snapshot and the compose replace boots the container on the old key and, not being in the snapshot, is never stopped.

Fix: applied `@_serialize_db` (below `@require_auth`) to `start_service` and `stop_service` (there is no separate restart route — the frontend "restart" is a stop+start pair, so both halves are now serialized). With start on the same lock as rotation, a concurrent start either:
- runs fully **before** rotation acquires the lock → it's then running, so rotation's `_affected_services_state()` snapshot includes it and the stop loop stops it; or
- runs fully **after** rotation releases the lock → `docker compose up` reads the already-rebuilt compose with the new key.

Either way no container is left serving the revoked key.

Added `test_rotation_vs_concurrent_start_no_stale_key_container`: models containers by the key they booted with, fires a start concurrently with a rotation (windows widened), and asserts the invariant *no running container is on a key other than the final rotated key*. It fails without the lock (start records the old key after the snapshot) and passes with it.

Backend suite: 156 passed.

teo-mateo:
## Codex review iteration 6 — No further findings

I reviewed the final diff, prior iteration comments/responses, and the checked-out PR branch. I focused on the rotation transaction boundary, shared service DB locking, benchmark apply interaction, start/stop interleavings, stop-failure reporting, and the frontend handling of partial rotation results.

I do not see any remaining correctness, security, API-contract, or production-performance findings that meet the bar for another defect report. The previously reported concurrency/atomicity issues appear addressed in the current branch.

Verification run locally: `dashboard/venv/bin/pytest dashboard/tests/test_key_rotation.py` — 13 passed.

PR #27, opened 2026-05-19
feat(chat): centered empty-state composer + model preselect

teo-mateo:
## Summary

Improves the `/v2/chat` first-run experience so a user can land on the page and immediately type.

- **Centered empty state**: with no conversation selected, `ChatArea` now renders a centered hero (icon, "Start a new conversation" title, subtitle naming the model) with a focused composer instead of the bare "Select or create a conversation" text.
- **Type-to-start**: sending from the empty-state composer creates a conversation with the first running model, navigates to it, and flushes the queued first message via `pendingMsgRef` once the conversation loads — no extra click.
- **Model preselect**: "New Conversation" defaults `main_service` to the first running service (replacing a stale hardcoded model name). When exactly one model is up it's preselected; when none are running it alerts / disables the composer (the backend requires a non-empty `main_service`).
- **Focus on conversation change**: `ChatInput` gained a `focusKey` so opening or creating a conversation lands the cursor in the composer (previously focus only moved on the `disabled` transition).

## Testing

- `npm run lint` — clean (also removed a pre-existing unused `convsLoading` binding in the touched file).
- `npm run build` — succeeds; bundle output to the gitignored `dashboard/frontend/dist/` that `app.py` serves `/v2` from.

### DIFF
--- a/dashboard/frontend/src/components/chat/ChatArea.jsx
+++ b/dashboard/frontend/src/components/chat/ChatArea.jsx
@@ -15,6 +15,7 @@ let spinoffCounter = 0
 
 export default function ChatArea({
   conversation,
+  awaitingConversation,
   defaultModelName,
   onCreateAndSend,
   messages,
@@ -91,6 +92,21 @@ export default function ChatArea({
     topZRef.current++
     setSpinoffs(prev => prev.map(s => s.id === id ? { ...s, zIndex: topZRef.current } : s))
   }, [])
+
+  // A conversation is in the URL but its fetch hasn't returned yet. Show a
+  // loading state — NOT the create composer — otherwise typing here would
+  // spawn a different new conversation instead of waiting for the requested
+  // one (codex iteration 1, P1).
+  if (!conversation && awaitingConversation) {
+    return (
+      <div className="flex-1 flex items-center justify-center text-gray-500">
+        <div className="text-center">
+          <i className="fa-solid fa-spinner fa-spin text-3xl mb-3 block opacity-40"></i>
+          <p>Loading conversation…</p>
+        </div>
+      </div>
+    )
+  }
 
   if (!conversation) {
     return (
--- a/dashboard/frontend/src/components/chat/ChatArea.jsx
+++ b/dashboard/frontend/src/components/chat/ChatArea.jsx
@@ -15,6 +15,8 @@ let spinoffCounter = 0
 
 export default function ChatArea({
   conversation,
+  defaultModelName,
+  onCreateAndSend,
   messages,
   critiques,
   setCritiques,
@@ -92,10 +94,22 @@ export default function ChatArea({
 
   if (!conversation) {
     return (
-      <div className="flex-1 flex items-center justify-center text-gray-500">
-        <div className="text-center">
-          <i className="fa-solid fa-comments text-5xl mb-4 block opacity-20"></i>
-          <p>Select or create a conversation</p>
+      <div className="flex-1 flex flex-col items-center justify-center px-4">
+        <div className="w-full max-w-2xl flex flex-col items-center">
+          <i className="fa-solid fa-comments text-6xl mb-5 text-blue-500/40"></i>
+          <h2 className="text-2xl font-semibold text-gray-200 mb-2">Start a new conversation</h2>
+          <p className="text-sm text-gray-500 mb-6 text-center">
+            {defaultModelName
+              ? <>Type a message below — a chat will be created with <span className="text-gray-400">{defaultModelName}</span>.</>
+              : 'No running model services available. Start a model first.'}
+          </p>
+          <div className="w-full">
+            <ChatInput
+              focusKey="empty-state"
+              onSend={(msg, images) => onCreateAndSend?.(msg, images)}
+              disabled={!defaultModelName}
+            />
+          </div>
         </div>
       </div>
     )
@@ -193,6 +207,7 @@ export default function ChatArea({
         {/* Input */}
         <ChatInput
           ref={composerRef}
+          focusKey={conversation.id}
           onSend={(msg, images) => { onSend(msg, images); setPendingInserts([]) }}
           disabled={streaming || !conversation.main_service}
           pendingInserts={pendingInserts}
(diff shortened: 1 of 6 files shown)


teo-mateo:

- **P1: Don’t show the active empty-state composer while an existing conversation is still loading.** In `dashboard/frontend/src/components/chat/ChatArea.jsx:95`, `!conversation` now always renders a focused composer that creates a new conversation. But `conversation` is also `null` during the async `loadConversation(convId)` path in `dashboard/frontend/src/components/chat/ChatPage.jsx:45-50`, such as opening `/v2/chat/<id>` directly or selecting a conversation before its fetch returns. During that window, the UI invites typing and Enter will call `handleCreateAndSend`, creating and navigating to a different new conversation instead of waiting for the requested one. Please gate this empty composer to the true no-conversation route/state, or pass an explicit loading/hasRouteConversation flag and render a loading/disabled state while `convId` is present and the conversation has not loaded yet. A regression test for direct-loading `/chat/:conversationId` would cover this path.

teo-mateo:

**Finding P1 — Empty-state composer renders while a URL conversation is still loading**: accepted.

Fixed in commit 4308a8e. `ChatPage` now passes `awaitingConversation={!!convId && (!conversation || conversation.id !== convId)}` to `ChatArea`. When a conversation id is in the route but its fetch hasn't returned, `ChatArea` renders a "Loading conversation…" state instead of the create composer, so typing/Enter can no longer spawn a different new conversation. The truly-empty route (`/v2/chat`, no id) still shows the focused create composer. Added `ChatArea.test.jsx` covering both branches (composer when no route conversation; loading + no textarea while awaiting). Lint clean, build OK, tests pass.

teo-mateo:

- **P1: Block the stale conversation composer while switching between existing conversations.** `ChatPage` correctly computes `awaitingConversation` when the URL id differs from the loaded conversation (`dashboard/frontend/src/components/chat/ChatPage.jsx:140`), but `ChatArea` only renders the loading state when `!conversation && awaitingConversation` (`dashboard/frontend/src/components/chat/ChatArea.jsx:100`). `useChat.loadConversation()` does not clear the previous `conversation` before awaiting the fetch, so selecting conversation B while conversation A is displayed leaves A's header/messages/composer active until B returns. During that window the sidebar/URL indicate B, but pressing Enter submits through A's `sendMessage` closure and appends the message to the wrong chat. Please treat `awaitingConversation` as authoritative for the main chat UI (or clear `conversation` before refetching) so the composer is disabled/hidden whenever the route id and loaded conversation id do not match. The new tests cover the direct-load `conversation=null` case, but should also cover switching from an already-loaded conversation to a different route id.

teo-mateo:

**Finding P1 — Stale conversation composer while switching between existing conversations**: accepted.

Fixed in commit d218e76. `ChatArea` now treats `awaitingConversation` as authoritative: the gate changed from `!conversation && awaitingConversation` to just `awaitingConversation`. Since `ChatPage` already computes `awaitingConversation = !!convId && (!conversation || conversation.id !== convId)`, switching from a loaded conversation A to route B now renders the "Loading conversation…" state (no header/messages/composer) until B resolves, so an Enter can no longer submit into A via the stale `sendMessage` closure. Direct-load and truly-empty (`/v2/chat`) paths are unaffected. Added a regression test covering switch-from-loaded (`conversation={id:'conv-A'}`, `awaitingConversation=true` → loading, no textarea). 3/3 tests pass, lint clean, build OK.

teo-mateo:

- **P1: Don’t flush the queued first message after the user has navigated away from the newly-created conversation.** `pendingMsgRef` is flushed whenever `conversation.id === pending.convId` and `!streaming` (`dashboard/frontend/src/components/chat/ChatPage.jsx:53-58`), but it does not also verify that the current route still points at that pending conversation. Since `loadConversation()` does not cancel or sequence-guard its `getConversation` request (`dashboard/frontend/src/hooks/useChat.js:62-70`), this can happen: submit from the empty-state composer, navigate to the new conversation C, immediately select existing conversation B before C’s fetch returns, then C’s slower/late fetch sets `conversation` to C while the URL is already B. The flush condition passes and sends the queued message to C in the background, even though the active route is B; that stream can then continue to mutate the shared chat state and refetch C while the UI is trying to show B. Please gate the flush on the route as well, e.g. require `convId === pending.convId`, and clear/cancel the pending message when the user leaves that route (or add request sequencing in `loadConversation`) so stale fetches cannot trigger sends for a non-active chat.

Checked: `npm test -- ChatArea.test.jsx` passes. Full `npm run lint` currently fails on pre-existing files outside this PR’s changed set, so I did not count that as a changed-code finding.

teo-mateo:

**Finding P1 — Queued first message can flush into a non-active conversation after navigation**: accepted.

Fixed in commit 2174328. The flush effect now gates on the live route as well: it abandons `pendingMsgRef` the moment `convId !== pending.convId`, so a late `getConversation` for the just-created chat C resolving after the user navigated to B can no longer trigger a background send into C. The decision (`wait` / `abandon` / `send`) was extracted to a pure module `pendingFlush.js` (also satisfies `react-refresh/only-export-components`) with unit tests covering: no pending, not-yet-loaded, streaming, match→send, and navigated-away→abandon (including `convId=null`).

Scope note on the broader `loadConversation()` request-sequencing point: that pre-existing race (no cancellation/sequence guard on `getConversation` in `useChat`) is out of scope for this PR — it predates the empty-state feature and affects normal A→B switching independent of the pending message. The route gate fully closes the wrong-send path introduced by the new `pendingMsgRef` mechanism, which is what this PR is responsible for.

8/8 frontend tests pass, ESLint exit 0 on changed files, build OK.

teo-mateo:

No further findings after reviewing the current diff and prior iteration comments. I checked the empty-state composer flow, route/loading gates, queued first-message flush logic, changed tests, and surrounding chat hooks/API contracts.

Local checks run:
- `npm test -- ChatArea.test.jsx pendingFlush.test.js` — 8/8 passed
- `npx eslint src/components/chat/ChatArea.jsx src/components/chat/ChatArea.test.jsx src/components/chat/ChatInput.jsx src/components/chat/ChatPage.jsx src/components/chat/pendingFlush.js src/components/chat/pendingFlush.test.js` — passed
- `npm run build` — passed, with the existing Vite chunk-size warning

Issue #29, opened 2026-05-19
Chat visual redesign — adopted decisions (typography, palette, transcript-as-log, collapsible nav)


teo-mateo:
## Summary

Adopted decisions from the chat-window redesign exploration ("Drydock" mock — see `design/chat-redesign/mock.html`, `DESIGN-NOTES.md`, `DECISIONS.md`). This issue is **only the decisions we kept**; rejected ideas are listed so implementation is unambiguous. The big Projects + Artifacts feature is tracked separately in #28 and is **out of scope here** (this issue only owns the *affordance* where they intersect).

The guiding principle: **distinctive typography on a refined-but-familiar dark blue palette** — not a wholesale visual reinvention.

### 1. Vocabulary — conventional terms only (D1)
No metaphor/jargon. Labels stay conversation-based: "Conversation", "Spin-off", "Critique panel", "Composer", "You/User". Labels/copy only; does not by itself change any visual treatment.

### 2. Collapsible global app nav (D2) — `components/Sidebar.jsx`
Currently a fixed 224px rail (logo + text labels Services/Chat/Tools + user block). Make it collapsible:
- Pin/chevron toggle. Collapsed = **icons only** with hover tooltips; expanded = icons + labels.
- State persisted in `localStorage`.
- Applies on **all routes**. **Default collapsed.**
- Reclaims horizontal space (esp. for chat).

### 3. Keep the active-conversation vertical bar (D3) — `ChatSidebar.jsx`
Retain the colored left vertical bar marking the active conversation row. Accent color follows the palette (blue per §7), not the mock's amber.

### 4. Transcript-as-log (D7) — `MessageList.jsx`, `MessageBubble.jsx`
- **Keep:** a timeline spine down the left gutter with a per-turn tick **node**.
- **Keep:** turns rendered as **hairline-bordered blocks with a colored left accent spine**, *not* filled rounded bubbles.
- **Keep:** assistant prose in the **serif** (Newsreader per §6); user input stays **mono**.
- **Drop:** the monospace `ROLE · model · seq · time · tok` meta line — keep a **light/minimal role label** like today.
- **Alignment:** user blocks stay **right-aligned**, assistant blocks **left-aligned**; the spine + nodes run down the left gutter (it is *not* collapsed to a single left column).

### 5. Composer updates (D8) — `ChatInput.jsx`
[...]

Issue #28, opened 2026-05-19
Projects + Conversation Artifacts (unified artifact store)


teo-mateo:
## Summary

Introduce two linked concepts:

1. **Project** — an opt-in container that owns multiple conversations and a shared artifact store, plus project-level defaults (model, system prompt, MCP toggles).
2. **Conversation/Project Artifacts** — a unified store of files/images that can be **uploaded by the user** or **generated by the model**, persisted in the DB, organized in a tree under the conversation/project in the sidebar, referenceable from chat, deletable by the user, and managed by the model through built-in tools.

This unifies the **existing** per-message render outputs (the current `artifacts` table — code/HTML/SVG shown by `ArtifactRenderer.jsx`) into the same store rather than keeping two unrelated "artifact" notions.

> Design discussion + full rationale: `design/chat-redesign/SPEC-conversation-artifacts.md` and `design/chat-redesign/DECISIONS.md` (D4–D6).

## Decisions (resolved in discussion)

- **A — Unify.** One artifact store. Existing per-message render outputs become a subtype (`source='render'`). Single term across the app: **Artifacts**.
- **B — Storage.** DB blobs in a **separate** `artifact_blobs` table (kept off the hot conversation/message rows). **10 MB/file** cap, MIME allowlist (images, text, pdf, json, csv, common code) — a configurable constant.
- **C — Tree UI.** Nested under the conversation/project in the **left sidebar**, grouped by origin: **Uploaded** / **Generated** (Generated = model `create_artifact` + migrated render outputs).
- **D — Spin-offs.** No special rule — they resolve to their lineage's scope (read + reference; create lands on the shared owner).
- **E — Project (now).** Introduce the Project entity in this work; artifacts are project-scoped when a conversation is in a project.
- **F — Projects optional.** `conversations.project_id` is **nullable**; projects are opt-in grouping.
- **G — Polymorphic owner.** An artifact belongs to exactly one of `{project, conversation-root}`.

## Data model

**`projects`**
`id`, `name`, `created_at`, `archived_at` (nullable), `default_main_service`, `default_sidekick_service`, `default_system_prompt`, `default_mcp_servers` (defaults nullable; new conversations inherit).

**`conversations`** — add `project_id` (nullable FK). Spin-offs inherit their parent's `project_id` (enforced equal: a spin-off shares its parent's project, or shares its parent's loose state).

**`artifacts`**
[...]

PR #33, opened 2026-05-19
feat(chat): composer attach affordance + restyled chips — issue #29 phase 4

teo-mateo:
## Phase 4 of issue #29 — Chat visual redesign

Implements §5 (composer updates) in `ChatInput.jsx`. Built on the phase-1 tokens.

### Kept (per §5)
- **Attach-files button + drag-and-drop** with allowlist validation (images / text / pdf / json / csv / code). One shared `ingestFiles` path for the button, drop and paste; depth-counted drag enter/leave with a drop hint. Images flow through the existing image pipeline; other allowlisted files become attachment chips. **Affordance only** — transport/storage of non-image attachments is owned by #28, so they are cleared on send.
- **Restyled critique pending-insert chips** to the refined visual language: hairline block + left `critique` accent + mono labels, using the phase-1 semantic tokens.
- New attachment chips share the same refined hairline + mono style.
- **Empty-state composer parity**: `ChatArea` renders the same `<ChatInput>` for both the empty-state (line 125) and active (line 226) composers, so the affordance + restyle apply there automatically — verified in the browser.

### Dropped (per §5)
- No command-console frame (prompt glyph / focus glow / mono hint line).
- No reference-an-existing-artifact picker (deferred to #28).

### Verification
- `npm run build` ✓
- Playwright against local `vite preview`, on the `/chat` empty-state composer:
  - attach button + multi `<input type=file>` present; `accept` includes `application/pdf` and code extensions ✓
  - drag-drop a `.py` + `.csv` + `.exe`: `.py`/`.csv` accepted as chips, `.exe` rejected by the allowlist ✓
  - placeholder updated; **0 console errors** ✓

No backend changes.

### DIFF
--- a/dashboard/frontend/src/components/chat/ChatInput.jsx
+++ b/dashboard/frontend/src/components/chat/ChatInput.jsx
@@ -111,15 +111,14 @@ const ChatInput = forwardRef(function ChatInput({ onSend, disabled, pendingInser
   }
 
   function handlePaste(e) {
-    const items = e.clipboardData?.items
-    if (!items) return
     if (disabled) return
-    for (const item of items) {
-      if (item.type.startsWith('image/')) {
-        e.preventDefault()
-        const file = item.getAsFile()
-        if (file) addImageFile(file)
-      }
+    // Route pasted files (images and allowlisted non-images, e.g. copied
+    // from the OS file manager) through the same ingest path as the
+    // attach button and drag-and-drop. Text paste falls through.
+    const files = e.clipboardData?.files
+    if (files && files.length > 0) {
+      e.preventDefault()
+      ingestFiles(files)
     }
   }
 
--- a/dashboard/frontend/src/components/chat/ChatInput.jsx
+++ b/dashboard/frontend/src/components/chat/ChatInput.jsx
@@ -1,9 +1,53 @@
 import { useState, useRef, useEffect, forwardRef, useImperativeHandle } from 'react'
 
+// Attach affordance allowlist (issue #29 §5). This is the affordance only —
+// where non-image attachments are stored/used is feature #28.
+const ALLOWED_EXT = new Set([
+  // text / docs
+  'txt', 'md', 'markdown', 'pdf', 'json', 'csv', 'tsv', 'log',
+  // code
+  'js', 'jsx', 'ts', 'tsx', 'py', 'rb', 'go', 'rs', 'java', 'c', 'h',
+  'cpp', 'cc', 'hpp', 'cs', 'php', 'swift', 'kt', 'sh', 'bash', 'zsh',
+  'sql', 'html', 'css', 'scss', 'yml', 'yaml', 'toml', 'ini', 'xml',
+])
+
+function fileExt(name) {
+  const i = name.lastIndexOf('.')
+  return i >= 0 ? name.slice(i + 1).toLowerCase() : ''
+}
+
+function isImage(file) {
+  return file.type.startsWith('image/')
+}
+
+function isAllowed(file) {
+  if (isImage(file)) return true
+  if (
+    file.type.startsWith('text/') ||
+    file.type === 'application/pdf' ||
+    file.type === 'application/json' ||
+    file.type === 'text/csv'
+  ) {
+    return true
+  }
+  return ALLOWED_EXT.has(fileExt(file.name))
+}
+
+function formatSize(bytes) {
+  if (bytes == null) return ''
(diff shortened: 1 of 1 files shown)


teo-mateo:

- P2 `dashboard/frontend/src/components/chat/ChatInput.jsx:198`: the drag/drop handlers remain active even when the composer is disabled. The attach button and textarea honor `disabled`, but dropping files on the composer still reaches `handleDrop`, calls `ingestFiles`, and for images reads full data URLs into component state. This means a disabled composer during streaming, or the empty-state composer with no model selected, can still accumulate attachments/images and potentially send them later once re-enabled. Please gate the drag enter/over/drop path and file ingestion on `disabled`, or avoid registering the drop handlers while disabled.

Verification: `npm run build` passes. `npm run lint` currently fails only on existing files outside this PR diff; `ChatInput.jsx` is not reported.

teo-mateo:

**Finding P2 — drag/drop active while composer disabled**: accepted.

Fixed in commit c318e8c. `ingestFiles()` now short-circuits when `disabled` (defense in depth), and `handleDragEnter` / `handleDrop` / `handlePaste` bail early when disabled — so a disabled composer (during streaming, or the empty-state with no model selected) neither shows the drop hint nor accumulates images/attachments. `handleDragOver`/`handleDrop` still call `preventDefault()` so the browser doesn't navigate to a dropped file, but no state is mutated. Build passes.

teo-mateo:

- P2 `dashboard/frontend/src/components/chat/ChatInput.jsx:113`: the paste path still only handles `image/*` clipboard items and bypasses `ingestFiles()`, so allowlisted non-image files copied from the OS/file manager and pasted into the composer are silently ignored instead of becoming attachment chips. The PR description calls out one shared ingest path for button, drop, and paste, and this gap is independent of the deferred #28 transport/storage work. Please route clipboard file items/files through `ingestFiles()` as well, while keeping the existing image `preventDefault()` behavior.

Verification: `npm run build` passes. `npm run lint` still fails on existing issues outside this PR's changed file; `ChatInput.jsx` is not reported.

teo-mateo:

**Finding P2 — paste path bypasses the shared ingest**: accepted.

Fixed in commit da3f3b3. `handlePaste` now routes `e.clipboardData.files` (images **and** allowlisted non-image files copied from the OS/file manager) through the shared `ingestFiles()` with `preventDefault()`; plain text paste still falls through unchanged, and the `disabled` early-return from iteration 1 is preserved. Button / drop / paste now genuinely share one ingest path.

Note on verification: this offline environment has no backend, so the empty-state composer renders `disabled` (no model service) — which, correctly, makes the iteration-1 disabled-gate suppress a synthetic paste. The paste path is now identical to the drag-drop path (both call `ingestFiles`), which was exercised live earlier (`.py`/`.csv` accepted, `.exe` rejected). Build passes.

teo-mateo:

- P3 `dashboard/frontend/src/components/chat/ChatInput.jsx:137`: the wrapper drag/drop handlers now call `preventDefault()` unconditionally, so dragging normal text into the composer textarea is swallowed instead of being inserted by the browser. The feature only needs to capture file drags; for non-file drags/drops the handlers should return without canceling the default textarea behavior (for example, gate `preventDefault()`/`setDragOver()`/`ingestFiles()` on `dataTransfer.types` containing `Files`, and let text drops fall through).

Verification: `npm run build` passes. `npm run lint` still fails on existing issues outside this PR's changed file; `ChatInput.jsx` is not reported.

teo-mateo:

**Finding P3 — unconditional preventDefault swallows text drag**: accepted.

Fixed in commit (see branch head). Added an `isFileDrag(e)` guard (`dataTransfer.types` includes `'Files'`) to all four drag handlers — `handleDragEnter/Over/Leave/Drop` now return early (no `preventDefault`, no state change) for non-file drags, so dragging selected text into the textarea uses the browser's native drop again. File drags still show the drop hint and run `ingestFiles`. Build passes.

(Live re-exercise of file-drop remains blocked by the offline composer being `disabled` per the iteration-1 gate; logic is unchanged from the previously-verified drop path apart from the new non-file fall-through.)

teo-mateo:

I reviewed the current PR branch after the iteration 1–3 fixes, including the full ChatInput.jsx change and the ChatArea.jsx call sites. I did not find any new correctness, security, API-contract, or production-performance issues that should block this PR.

Verification: npm run build passes. npm run lint still fails on pre-existing issues outside this PR changed file; ChatInput.jsx is not reported.

Issue #23, opened 2026-05-14
Move tool-specific guidance out of the default system prompt into MCP tool_hints


teo-mateo:
## Problem

Two related design issues with how the chat system prompt is structured today:

### 1. The default system prompt is hardcoded in Python

`DEFAULT_MAIN_SYSTEM_PROMPT` lives in \`dashboard/chat/constants.py\` as a Python string constant. Iterating on it requires a code edit + service restart, and the value isn't visible to the user without reading the source. New conversations get this baseline; existing conversations carry their own copy in the DB.

### 2. Tool-specific guidance is in the base prompt instead of the tools

Recent work (commit on \`feature/streaming-feedback\`) added a substantial \"Web research\" / \"Reporting on web research\" section to the default prompt — guidance about the search→fetch workflow, how to handle 403/404, source quality, citation honesty, etc.

This guidance is only relevant when the websearch / web fetch MCP server is enabled. Loading it for every conversation regardless wastes tokens (~500 tokens on a fairly basic chat) and isn't where a user would look to understand or modify the tool's behavior.

### A. Make the default system prompt editable from the dashboard

- Surface it on a settings / system-prompts page (probably alongside or near the Tools page added in PR #19).
- Store it in a config file or a singleton settings row in the DB. Loaded at conversation creation, same as today.
- Reset-to-built-in button so the user can recover.
- Out of scope: per-user prompts, prompt versioning. Single global default is enough.

### B. Move tool-specific guidance into MCP \`tool_hint\` fields

- Tool hints already get appended to the system prompt at runtime when the corresponding MCP server is enabled (\`dashboard/chat/mcp_registry.py:get_tool_hints\`).
- Rewrite the Web research section as content of the \`websearch\` MCP server's \`tool_hint\` in \`mcp_servers.json\` (external file, machine-local — already gitignored).
- Similarly for web_fetch if it's a separate server.
- Result: the guidance loads if-and-only-if the user has those tools enabled for the conversation, and it lives next to the tools it describes.

### C. Strip tool-specific guidance from the default prompt

After (A) and (B), the default prompt should only contain the generic style/accuracy/tools-posture guidance. Anything specific to web research, sympy, schemdraw, render-html, etc. lives in the corresponding tool_hint.

## Acceptance criteria

- [ ] Editing the global default prompt is possible from the dashboard UI (no code edit + service restart).
[...]

Issue #1, opened 2026-02-14
Changelog


teo-mateo:
@claude add to the changelog in the main README.MD, BE THOROUGH

PR #41, opened 2026-05-20
feat(chat): editable default main_system_prompt — backend (issue #23 phase 1)

teo-mateo:
Phase 1 of issue #23. Adds the persistence + API surface for an editable global default `main_system_prompt`; the UI lands in phase 2.

## What

- **`chat/settings_store.py`** — JSON-file singleton with atomic tempfile+fsync+`os.replace` write, in-process lock for concurrent setters, `LLM_DOCK_CHAT_SETTINGS_FILE` env override. Empty string and the literal built-in both count as uncustomized so the indicator stays honest.
- **`chat/routes.py`** — GET / PUT / DELETE on `/api/chat/settings/main-system-prompt`. Response shape: `{current, builtin, customized}`. `create_conversation` reads `settings_store.get_main_system_prompt()` when the client did not pass a prompt explicitly.
- **`chat/constants.py`** — drop the last remaining tool-specific bullet (web search query year) from the default prompt. It moves into the websearch MCP server's `tool_hint` in phase 3.
- **Tests** — 29 new tests across `test_settings_store.py` (atomic write, concurrent writers, env override, malformed file fallback, idempotent reset) and `test_chat_settings_routes.py` (auth on all three verbs, validation, GET-after-PUT consistency, plus a regression test for acceptance criterion #2 — the built-in baseline must contain no tool-specific keywords).

## Acceptance criterion mapping

- **#1** Editing the global default prompt is possible without a code edit + restart → **API in place**; UI follows in phase 2.
- **#2** No-MCP conversation has no tool-specific text → **passing** and locked in by `test_builtin_prompt_has_no_tool_specific_guidance`.
- **#3** Websearch MCP enabled gets its hint appended → already true via `routes._stream_response` augmentation; phase 3 wires the example content.
- **#4** Existing conversations unaffected → guaranteed by reading from `settings_store` only on `create_conversation` for the no-explicit-prompt path; stored conversations keep their own `main_system_prompt`.

## Tests

Full backend suite: **185 passed**.

This is phase 1 of 3 per the per-phase delivery workflow. Phase 2 (frontend editor on the Tools page) and phase 3 (docs + websearch example) follow on separate branches.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,9 @@ docker-compose.lock
 # Machine-local MCP registry (declared external servers)
 dashboard/mcp_servers.json
 
+# Machine-local chat settings (user-customized default system prompt)
+dashboard/chat_settings.json
+
 # Logs
 *.log
 dashboard/dashboard.log
--- a/dashboard/chat/constants.py
+++ b/dashboard/chat/constants.py
@@ -33,11 +33,7 @@ Tool use
   reasoning. If you have nothing to add, write a brief direct answer rather than
   emitting empty output.
 - Hard cap: at most 5 tool calls per question. If you hit the cap without a confident
-  answer, answer with what you have and flag the uncertainty.
-- When constructing web search queries about current state of anything (prices, products,
-  events, rankings, "best X in Y"), use the year from the `Current date:` line below —
-  not years from your training data. Do not append "2024" or "2025" to a query unless
-  the user specifically asked about that year."""
+  answer, answer with what you have and flag the uncertainty."""
 
 DEFAULT_SIDEKICK_SYSTEM_PROMPT = (
     "You are a critical technical reviewer. When asked to critique a response, "
--- a/dashboard/chat/routes.py
+++ b/dashboard/chat/routes.py
@@ -17,6 +17,7 @@ from .constants import DEFAULT_MAIN_SYSTEM_PROMPT, DEFAULT_SIDEKICK_SYSTEM_PROMP
 from .llm_proxy import stream_chat_completion, build_messages_array, resolve_service
 from .critique import build_critique_context, request_critique, validate_annotations
 from .mcp_registry import list_available_servers, get_tool_hints
+from . import settings_store
 from .tool_loop import stream_with_tools
 
 logger = logging.getLogger(__name__)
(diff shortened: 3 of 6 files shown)


teo-mateo:

- `dashboard/chat/routes.py:84`: `PUT /api/chat/settings/main-system-prompt` handles missing JSON and object payloads, but it assumes any successfully parsed JSON is a dict before calling `data.get(\"content\")`. A non-object JSON body such as `[1]`, `\"x\"`, or `123` currently raises `AttributeError` and returns a 500 instead of the endpoint's documented/implemented 400 validation response. I reproduced this against the checked-out branch with Flask's test client. Please guard `isinstance(data, dict)` before `.get()` and add a route test for a non-object JSON payload.

teo-mateo:

**Finding 1.1 — non-object JSON body 500s instead of 400**: accepted.

Fixed in commit 73574f4. `request.get_json(silent=True)` returns the parsed value as-is, so a non-object body (`[1]`, `"x"`, `123`, `true`) was truthy, survived the `or {}` fallback, and reached `data.get()` → `AttributeError` → 500. Added an `isinstance(data, dict)` guard before `.get()`, returning the documented 400. New parametrized route test `test_put_rejects_non_object_json_body` covers list/string/number/bool/null bodies; it also asserts no customization is recorded on the rejected request.

Phase-1 test suite: 34 passed.

teo-mateo:

- `dashboard/chat/settings_store.py:31`: the new persistent settings file defaults to `dashboard/chat_settings.json`, but `.gitignore` does not ignore that generated machine-local file. After a real `PUT /api/chat/settings/main-system-prompt`, the repo will show an untracked JSON file containing the user's custom default system prompt, which can easily include private operational instructions or other sensitive prompt text. This is the same class of local dashboard state as `services.json` and `dashboard/mcp_servers.json`, both already ignored. Please add `dashboard/chat_settings.json` to the generated/machine-local ignores so it cannot be accidentally committed.

- `dashboard/chat/routes.py:124`: the new route tests cover the settings endpoints, but they do not exercise the integration point that actually consumes the setting: `POST /api/chat/conversations` without an explicit `main_system_prompt`. A regression that switched this line back to `DEFAULT_MAIN_SYSTEM_PROMPT` would still leave the new tests green while breaking the main acceptance criterion. Please add a route-level test with an in-memory `ChatDB` that sets a custom default, creates a conversation without `main_system_prompt`, and asserts the stored conversation uses the custom value; ideally also assert that an explicit `main_system_prompt` still overrides the global default.

teo-mateo:

**Finding 2.1 — `chat_settings.json` not gitignored**: accepted.

Fixed in commit 78eed02. Added `dashboard/chat_settings.json` to `.gitignore` alongside the existing `dashboard/mcp_servers.json` machine-local ignore. Verified with `git check-ignore`.

**Finding 2.2 — no test for the conversation-creation integration point**: accepted.

Fixed in commit 78eed02. Wired an in-memory `ChatDB` into the test fixture and added 4 tests against `POST /api/chat/conversations`:
- `test_new_conversation_uses_custom_default_prompt` — custom default flows through
- `test_new_conversation_uses_builtin_when_no_customization` — built-in used when uncustomized
- `test_explicit_prompt_overrides_global_default` — explicit body prompt still wins
- `test_explicit_empty_prompt_is_honored_not_replaced` — explicit `""` is honored (the key-presence branch must not substitute the global default)

Phase-1 test suite: 38 passed.

teo-mateo:

I reviewed the latest diff and prior comments, including the fixes for the non-object JSON validation, gitignore coverage for the local settings file, and conversation-creation integration tests. I also checked the full touched files locally and ran the focused backend tests with the dashboard venv:

```
venv/bin/python -m pytest tests/test_settings_store.py tests/test_chat_settings_routes.py
# 38 passed
```

No further findings.

PR #44, opened 2026-05-21
fix(chat): make the per-conversation system prompt actually editable

teo-mateo:
## Problem

The chat page's "System Prompts" panel was effectively read-only. The textareas were controlled components bound straight to the persisted `conversation.main_system_prompt` / `sidekick_system_prompt` props, while `onChange` only fired an async server PUT and never touched local state. Every keystroke re-rendered the textarea back to the stale prop value, so the box looked frozen — and each keystroke also fired a full PUT request.

This is a pre-existing bug, separate from issue #23 (which made the *global default* prompt editable on the Tools page).

## Fix

- **`SystemPromptEditor.jsx`** — the Main Model prompt is now backed by **local state** and persisted once, **on blur**. A `savedValue` tracks the last-persisted text so the "Unsaved" indicator is correct without the parent reloading the conversation. Save errors surface inline; a failed save keeps the editor dirty.
- **Removed the Critic Model (sidekick) textarea** — the sidekick prompt isn't tuned per-conversation. The `sidekick_system_prompt` DB field and its default are untouched; only the per-conversation editor is dropped.
- **`ChatArea.jsx`** — keys `SystemPromptEditor` by `conversation.id` so local edit state resets cleanly on conversation switch; passes the renamed `onSaveMain`.

## Tests

New `SystemPromptEditor.test.jsx` (8 tests): collapsed-by-default, single textarea (no critic), editability, save-on-blur, no-op blur when unchanged, Unsaved-indicator lifecycle, error surfacing, no double-save. Full frontend suite: **66 passed**. Lint + `vite build` clean.

### DIFF
--- a/dashboard/frontend/src/components/chat/ChatArea.jsx
+++ b/dashboard/frontend/src/components/chat/ChatArea.jsx
@@ -43,6 +43,9 @@ export default function ChatArea({
   const [spinoffs, setSpinoffs] = useState([]) // [{id, text, minimized, zIndex}]
   const topZRef = useRef(100)
   const composerRef = useRef(null)
+  // Tracks an in-flight system-prompt save so a fast send can't POST a
+  // message before the prompt PUT lands (SystemPromptEditor saves on blur).
+  const pendingPromptSaveRef = useRef(null)
 
   const quoteSelection = useCallback((text) => {
     if (!text) return
@@ -153,7 +156,31 @@ export default function ChatArea({
   }
 
   async function handleSystemPromptChange(field, value) {
-    await updateConversation(conversation.id, { [field]: value })
+    const save = updateConversation(conversation.id, { [field]: value })
+    pendingPromptSaveRef.current = save
+    try {
+      await save
+    } finally {
+      // Only clear if a newer save hasn't replaced this one.
+      if (pendingPromptSaveRef.current === save) pendingPromptSaveRef.current = null
+    }
+  }
+
+  async function handleSend(msg, images) {
+    // SystemPromptEditor persists on blur; focusing the composer blurs it,
+    // so a save may be in flight right as the user hits send. Wait it out
+    // so the backend reads the new prompt when it generates the reply.
+    const pending = pendingPromptSaveRef.current
+    if (pending) {
+      try {
+        await pending
+      } catch {
+        // The failure is already surfaced in SystemPromptEditor; don't
+        // block the send on it (the prompt just stays at its old value).
+      }
+    }
+    onSend(msg, images)
+    setPendingInserts([])
   }
 
   return (
@@ -226,7 +253,7 @@ export default function ChatArea({
         <ChatInput
           ref={composerRef}
           focusKey={conversation.id}
-          onSend={(msg, images) => { onSend(msg, images); setPendingInserts([]) }}
+          onSend={handleSend}
           disabled={streaming || !conversation.main_service}
           pendingInserts={pendingInserts}
           onClearInsert={(idx) => setPendingInserts(prev => prev.filter((_, i) => i !== idx))}
--- a/dashboard/frontend/src/components/chat/ChatArea.jsx
+++ b/dashboard/frontend/src/components/chat/ChatArea.jsx
@@ -43,9 +43,9 @@ export default function ChatArea({
   const [spinoffs, setSpinoffs] = useState([]) // [{id, text, minimized, zIndex}]
   const topZRef = useRef(100)
   const composerRef = useRef(null)
-  // Tracks an in-flight system-prompt save so a fast send can't POST a
-  // message before the prompt PUT lands (SystemPromptEditor saves on blur).
-  const pendingPromptSaveRef = useRef(null)
+  // Lets handleSend flush a just-edited system prompt to the server before
+  // the message is posted (SystemPromptEditor persists on blur).
+  const promptEditorRef = useRef(null)
 
   const quoteSelection = useCallback((text) => {
     if (!text) return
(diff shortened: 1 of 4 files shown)


teo-mateo:

- dashboard/frontend/src/components/chat/SystemPromptEditor.jsx:56 / dashboard/frontend/src/components/chat/ChatArea.jsx:155: Saving the prompt only from the textarea blur handler leaves a race with sending the next message. A normal flow is edit the system prompt, click/focus the composer, then immediately press Enter/send; `onBlur` starts `updateConversation(...)`, but `ChatInput` remains enabled and `send_message` reads `conv = db.get_conversation(conv_id)` before streaming. If the POST to `/messages` reaches the backend before the PUT completes, that first response is generated with the old `main_system_prompt` even though the UI says the prompt is saved when you click away. The new tests cover the component calling `onSaveMain` on blur, but not that sends are gated on the pending save. Consider tracking the pending prompt save in `ChatArea` and disabling/awaiting send while it is in flight, or otherwise ensuring the current local prompt is persisted before starting a message stream.

teo-mateo:

**Finding 1 — save-on-blur races with sending the next message**: accepted.

Real race. SystemPromptEditor persists on blur, and focusing the composer is exactly what blurs it — so the prompt PUT can be in flight when the user hits send, and `POST /messages` could reach the backend first, generating the first reply with the stale `main_system_prompt`.

Fixed in commit 9f4862d. `ChatArea` now tracks the in-flight prompt save in `pendingPromptSaveRef` (declared above the early `return`s to satisfy the rules of hooks — `handleSystemPromptChange` runs after them). The send path (`handleSend`) awaits that promise before calling `onSend`. A failed save does **not** block the send: the error is already surfaced in `SystemPromptEditor` and the prompt just stays at its old value, so swallowing it in `handleSend` is intentional.

Added 3 ChatArea regression tests: send waits for an in-flight save, sends immediately when none is pending, and still sends if the save fails. Full frontend suite: 69 passed.

teo-mateo:

- dashboard/frontend/src/components/chat/SystemPromptEditor.jsx:24-35: `commit()` drops any blur while `saving` is true, so edits made during an in-flight save can be left dirty with no later persistence attempt. Repro: edit prompt to A and blur, the PUT starts; before it resolves, focus the textarea again, change it to B, and blur. That second blur returns at line 25 because `saving` is still true. When the first request resolves, `savedValue` is set to A and `saving` becomes false, but B was never sent and the user is already blurred away. The UI still shows Unsaved, but the advertised "Saved when you click away" contract is broken, and `ChatArea` has no pending save to await for B before the next message. Please queue/redo a save when `value !== savedValue` after the in-flight save settles, or otherwise disable editing while saving so a blur cannot be ignored. A regression test for edit-while-saving would cover this gap.

teo-mateo:

**Finding 2 — a blur during an in-flight save is dropped**: accepted.

Real bug. `commit()`'s `if (saving) return` guard meant an edit made while an earlier PUT was still running never got its own save — when the first request resolved, the later value was left unpersisted while the editor still showed "Unsaved".

Fixed in commit 60d542b. `commit()` now loops: it re-reads the latest textarea value via `valueRef` (a ref mirroring `value`, so the read survives an `await`) and keeps issuing saves until what's on disk matches the textarea. A blur dropped by the `saving` guard is now harmless — the already-running loop picks the edit up. On the last iteration `savedValue` is set to the persisted value, clearing the "Unsaved" indicator.

Added the requested regression test — `persists an edit made while an earlier save is still in flight`: edit→blur (deferred save A), edit→blur (B, dropped by guard), resolve A, assert the loop then persists B and clears "Unsaved". Full frontend suite: 70 passed.

teo-mateo:

- dashboard/frontend/src/components/chat/ChatArea.jsx:173-182 / dashboard/frontend/src/components/chat/SystemPromptEditor.jsx:44-47: `handleSend` only awaits the pending save promise that exists when send starts, but `SystemPromptEditor` can enqueue another save immediately after that promise resolves. Repro: edit to A and blur so save A starts; while A is in flight, edit to B and blur (the editor's `saving` guard returns, relying on the loop); press Enter before A resolves. `handleSend` captures/awaits save A. When A resolves, the editor loop observes B and calls `onSaveMain(B)`, which replaces `pendingPromptSaveRef.current`, but `handleSend` does not re-check the ref after awaiting A and proceeds to `onSend` while save B is still in flight. That means the message can still be generated with prompt A instead of the visible prompt B, so the iteration-1 race remains for the edit-while-saving path. A focused regression test should cover this sequence; the send gate likely needs to wait until the pending-save ref is clear/stable, not just await the first captured promise.

teo-mateo:

**Finding 3 — send gate only awaited the first save, not the chained follow-up**: accepted.

Correct — the iteration-1 `pendingPromptSaveRef` captured a single promise at send time. When an edit landed during an in-flight save, `SystemPromptEditor`'s commit loop issued a follow-up PUT *after* the first resolved, and `handleSend` proceeded to `onSend` without re-checking, so the reply could still use the intermediate prompt.

Fixed in commit b19c87a by replacing the single-promise gate with an imperative **`flush()`**:

- `SystemPromptEditor` is now a `forwardRef` component exposing `flush()` via `useImperativeHandle`. `commit()` tracks its in-flight loop in `commitRunningRef`; `flush()` kicks off a save if the value is dirty and awaits the loop to completion. Because the loop re-reads `valueRef` every iteration, `flush()` drains the **entire** save chain, not just the first request.
- `ChatArea` drops `pendingPromptSaveRef`; `handleSend` awaits `promptEditorRef.current.flush()` before `onSend`. `handleSystemPromptChange` is back to a plain `updateConversation` call.

Tests:
- `ChatArea` — new **`waits for the whole save chain when an edit lands during a save`**: edit A → blur, edit B → blur, send while A in flight; asserts the send stays held after A resolves *and* the follow-up PUT for B is issued, and only fires once B lands. `sends immediately` relaxed to async (`flush()` is always awaited now).
- `SystemPromptEditor` — `flush()` persists an un-blurred edit and resolves only after it lands; `flush()` is a no-op when nothing is dirty.

Full frontend suite: 73 passed. Lint + build clean.

teo-mateo:

- dashboard/frontend/src/components/chat/ChatArea.jsx:162-170 / dashboard/frontend/src/components/chat/ChatInput.jsx:187-193: `handleSend` now awaits `promptEditorRef.current?.flush()` before calling `onSend`, but the composer remains enabled and `streaming` is still false during that wait. A user can submit once while a prompt save is in flight, then type/submit again before the save resolves; both `handleSend` calls await the same flush and then both call the `useChat.sendMessage` callback created from the pre-streaming render. Because that callback closes over `streaming === false`, both sends can start concurrent SSE requests for the same conversation instead of the second being blocked by the normal streaming guard. The current regression tests only assert one held send. Please add a local "send pending"/flush gate in `ChatArea` or disable/await the composer submission path while the flush is pending, and cover the double-submit case.

teo-mateo:

**Finding 4 — double-submit during the flush await opens concurrent streams**: accepted.

Real, and my PR widened the window: `handleSend` awaits `flush()` before `onSend`, and the composer stays enabled with `streaming` still false for that whole PUT round-trip. Two submits in that window both await the flush, then both reach the pre-streaming `sendMessage` closure (whose `if (streaming) return` guard sees `false` in both) → two concurrent SSE streams.

Fixed in commit 15ee0b6. Added a `sendInFlightRef` re-entrancy guard in `ChatArea`: `handleSend` drops a second call while the first is still preparing (awaiting `flush()`). Once `onSend` runs, `sendMessage` flips `streaming` true synchronously and the composer's own `disabled` guard takes over — so the ref is cleared in a `finally` without leaving a gap.

Test: new ChatArea case **`ignores a re-entrant submit while a prompt save is still flushing`** — two Enter presses during an in-flight save resolve to exactly one `onSend` call. Full frontend suite: 74 passed. Lint + build clean.

teo-mateo:

I reviewed the current diff, the full relevant files (`ChatArea.jsx`, `SystemPromptEditor.jsx`, `ChatInput.jsx`, and the `useChat` send path), and the prior PR discussion. The previously reported prompt-save/send races appear addressed by the current `flush()` loop plus the local re-entry guard, with regression coverage for the chained-save and double-submit cases.

Focused verification: `npm test -- --run src/components/chat/SystemPromptEditor.test.jsx src/components/chat/ChatArea.test.jsx` passed (2 files, 19 tests).

PR #48, opened 2026-05-23
feat(chat): shift-click range selection in conversation sidebar

teo-mateo:
## Summary

Adds shift-click range selection to the chat sidebar's conversation checkboxes.

- Plain click on a checkbox: toggle that row, anchor moves to it (same as before, just now with an explicit anchor tracked).
- Shift-click another checkbox: select every conversation between the anchor and the clicked row in render order — roots and their spin-offs interleaved depth-first.
- Anchor stays put across consecutive shift-clicks (Finder / Gmail behavior), so you can fan a range out from one origin.
- Anchor resets when the selection is cleared.

## Notes

- The range is built from `flatOrder`, a memoized depth-first walk of the visible tree, so spin-offs are correctly included when the anchor and clicked checkbox span them.
- Native `<input type="checkbox">` toggle is suppressed (`onChange` no-op, `onClick` handles the toggle ourselves) so we can read `shiftKey` off the click event — `onChange` doesn't expose modifiers.
- Anchor lives in a ref, not state, because it never needs to trigger a re-render.

## Test plan

- [x] `npm test` — 75/75 (no regressions; existing select-mode flow unchanged for plain clicks).
- [x] `npm run build` — clean.
- [ ] Manual: click checkbox A, shift-click checkbox B several rows down → all rows between A and B selected.
- [ ] Manual: anchor at root A, shift-click checkbox on a spinoff of root C → range includes A, all rows in between (including A and C's roots/spinoffs in order), up to the clicked spinoff.
- [ ] Manual: Clear → next click starts a fresh anchor.

### DIFF
--- a/dashboard/frontend/src/components/chat/ChatSidebar.jsx
+++ b/dashboard/frontend/src/components/chat/ChatSidebar.jsx
@@ -22,10 +22,19 @@ function ConversationItem({ conv, activeId, depth, selectMode, selected, onToggl
       }`}
       style={{ paddingLeft: `${12 + depth * 16}px`, paddingRight: 12 }}
     >
+      {/* Drive the toggle from the label's click rather than the nested
+          <input>'s. The label is a 32px hit target wrapping a 14px
+          checkbox; clicks on the surrounding padding reach the input
+          only via the browser's label-activation synthetic click, whose
+          modifier-key state is not guaranteed to mirror the original
+          mouse event. Reading shiftKey from the label click — and
+          preventDefault-ing so the synthetic input toggle never fires —
+          keeps the full hit target consistent with the visible checkbox
+          (PR #48 codex iter 2). */}
       {isSpinoff ? (
         <label
           className="group/iconslot relative w-8 h-8 -m-2 flex items-center justify-center flex-shrink-0 cursor-pointer"
-          onClick={e => e.stopPropagation()}
+          onClick={e => { e.preventDefault(); e.stopPropagation(); onToggleSelect(conv.id, e.shiftKey) }}
         >
           <i
             className={`fa-solid fa-code-branch text-[10px] opacity-50 text-purple-400 absolute inset-0 flex items-center justify-center transition-opacity pointer-events-none ${spinoffIconVisibilityClass}`}
@@ -34,7 +43,6 @@ function ConversationItem({ conv, activeId, depth, selectMode, selected, onToggl
             type="checkbox"
             checked={selected}
             onChange={() => {}}
-            onClick={e => { e.stopPropagation(); onToggleSelect(conv.id, e.shiftKey) }}
             className={`w-3.5 h-3.5 accent-accent cursor-pointer transition-opacity ${spinoffCheckboxVisibilityClass}`}
             title="Select"
           />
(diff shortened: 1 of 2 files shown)


teo-mateo:

- [low] The new shift-click selection path has no direct regression coverage. The changed behavior lives in `dashboard/frontend/src/components/chat/ChatSidebar.jsx:125` and depends on several easy-to-break details: the anchor ref staying fixed across consecutive shift-clicks, `flatOrder` matching rendered DFS order across spinoffs, and `clearSelection()` resetting the anchor. The current suite still passes, but there are no `ChatSidebar` tests, so a regression in this new behavior would not be caught. A focused RTL test can render roots plus spinoffs, click one checkbox, shift-click a later spinoff, assert every id in the rendered range is selected, then clear and verify the next shift-click does not reuse the old anchor.

Checked prior comments: none. Verified locally: `npm test` (75/75) and `npm run build` both pass; build emits the existing large chunk warning.

teo-mateo:

**Finding 1.1 — Missing regression coverage for shift-click selection**: accepted. Fixed in commit a4f07de — added `ChatSidebar.test.jsx` (6 tests) covering:

- Plain checkbox click toggles a single row and sets the anchor
- Shift-click without an anchor falls back to a single toggle
- Shift-click selects the inclusive depth-first range (including interleaved spinoffs) in the forward direction
- Same in the reverse direction (clicked above anchor)
- Anchor persists across consecutive shift-clicks, extending from the same origin
- Clear resets the anchor — a subsequent shift-click no longer extends the old range

Full suite 81/81.

teo-mateo:

- [low] Shift-click only works when the pointer lands directly on the small `<input>`, not the full checkbox hit target. The row uses a 32px label wrapper as the clickable checkbox area (`ChatSidebar.jsx:26-28`, `:43-45`), but the range logic reads `e.shiftKey` only from the input click handler (`ChatSidebar.jsx:37`, `:51`). When a user shift-clicks the label padding, the browser activates the nested checkbox by dispatching a synthetic input click, and that synthetic click does not carry the original mouse modifier, so `toggleSelect(..., shiftKey)` sees `false` and falls back to a single toggle. The new tests call `fireEvent.click()` on the input itself, so they miss this path. Handling the toggle from the label click with the original event, or otherwise preserving the modifier before label activation reaches the input, would make the expanded hit target behave like the visible checkbox control.

Verified: prior iteration 1 requested coverage and was addressed in `a4f07de`; I did not re-raise it. I also ran `npm test -- ChatSidebar.test.jsx --run` locally and the 6 focused tests pass.

teo-mateo:

**Finding 2.1 — Shift-click hit-target inconsistency on label padding**: accepted. Fixed in commit 5ecab9a.

Moved the toggle handler from the inner `<input>` onto the wrapping `<label>`:

```jsx
<label
  onClick={e => { e.preventDefault(); e.stopPropagation(); onToggleSelect(conv.id, e.shiftKey) }}
>
  <input type="checkbox" checked={selected} onChange={() => {}} ... />
</label>
```

Now whether the user clicks the inner 14px input or anywhere in the 32px label padding, the label's onClick is the single source of truth for the toggle, and it reads `shiftKey` directly off the original mouse event. `preventDefault` suppresses the browser's default label-activation toggle so the input never double-fires.

Updated the test helper to target the label (the actual click handler owner). Added a code-comment noting that the direct-input-click path can't be tested under jsdom — its label-activation model doesn't propagate input clicks back up to the wrapping label's onClick the way real browsers do — and called it out in the PR test plan instead.

`npm test` 81/81, `npm run build` clean.

teo-mateo:

I reviewed the current diff and prior comments, including the iteration 1 coverage request and iteration 2 label hit-target issue, both of which have been addressed in the PR branch. I did not find any further correctness, security, API contract, performance, or test-gap issues in the changed files.

Verified locally:
- `npm test -- ChatSidebar.test.jsx --run` passes: 6/6
- `npm test` passes: 81/81
- `npm run build` passes with the existing large chunk warning

Note: `npm run lint` currently fails on pre-existing unrelated files outside this PR's changed paths, so I did not treat that as a finding for PR #48.

PR #49, opened 2026-05-23
feat(mcp): support remote MCP servers over HTTP

### LINKED ISSUE #45: Support remote MCP servers over HTTPS
## Summary

Add support for connecting to MCP servers over HTTPS, in addition to the
current local-subprocess (stdio) transport.

## Current state

External MCP servers are declared in `dashboard/mcp_servers.json` and are
always spawned as **local subprocesses**:

- `dashboard/chat/mcp_client.py` connects only via `mcp.client.stdio`
  (`StdioServerParameters` / `stdio_client`).
- Each entry requires an absolute `command` path plus `args` — the server
  has to live on the same machine, with its own venv and files.
- See `dashboard/mcp_servers.example.json` and the "Adding an external MCP
  server" section of `CLAUDE.md`.

This rules out hosted / remote MCP servers entirely.

## Proposal

Allow an MCP server entry to declare an HTTP(S) transport instead of a
local command. The MCP SDK provides `streamablehttp_client` (and the older
`sse_client`) for this.

Sketch of a possible config shape:

```json
{
  "remote-tool": {
    "enabled": true,
    "name": "Remote Tool",
    "description": "...",
    "transport": "http",
    "url": "https://mcp.example.com/mcp",
    "headers": { "Authorization": "Bearer ${SOME_ENV_VAR}" },
    "icon": "fa-cloud",
    "tool_hint": "..."
  }
}
```

`transport` defaults to `stdio` for backward compatibility (existing
entries with `command`/`args` keep working unchanged).

## Work items

- [ ] Add a `transport` field to the MCP server schema; keep `stdio` as
      the default so existing configs are untouched.
- [ ] Branch in `mcp_client.py` between `stdio_client` and
      `streamablehttp_client` based on `transport`.
- [ ] Load-time validation for HTTP entries: require `url`, validate the
      scheme; `command`/`args` become optional (and irrelevant) for HTTP.
- [ ] Decide how auth headers / secrets are supplied (env-var
      interpolation vs. plain values — secrets should not be committed).
- [ ] Update the `/tools` page UI and JSON validation for the new shape.
- [ ] Update the "Adding an external MCP server" docs in `CLAUDE.md` and
      `mcp_servers.example.json`.

## Notes / open questions

- Connection lifecycle / timeouts differ from stdio — remote servers can
  be slow or flaky; current code uses a 30s default timeout.
- Tool-name namespacing (`server_id__tool_name`) is transport-agnostic and
  needs no change.

teo-mateo:
Closes #45.

## Summary

External MCP server entries in `dashboard/mcp_servers.json` may now declare a remote HTTP(S) transport instead of a local subprocess command. The existing stdio path is unchanged and remains the default — all in-tree configs keep working as-is.

```json
{
  "ragflow": {
    "enabled": true,
    "name": "RagFlow",
    "description": "Query the local RagFlow knowledge base over HTTP",
    "transport": "http",
    "url": "http://localhost:9382/mcp",
    "headers": { "Authorization": "Bearer ..." },
    "icon": "fa-cloud",
    "tool_hint": "..."
  }
}
```

\`type: \"remote\"\` is accepted as a synonym for \`transport: \"http\"\` so configs from other harnesses (e.g. Claude Code) can be pasted in unchanged.

## Implementation

- **\`mcp_config.py\`**: extended \`_validate_entry\` and \`_normalize_external\` with a \`transport\` field (default \`\"stdio\"\`). HTTP entries require a \`url\` with an http(s) scheme; \`headers\` is an optional string→string map. \`_chat_available\` now considers HTTP entries available the moment they're enabled with a URL — reachability is left to call time so an unreachable remote can't block the dashboard.
- **\`mcp_client.py\`**: factored a transport-agnostic \`_open_streams\` async context manager. Both \`_discover_tools\` and \`_execute_tool\` go through it, so the chat path is otherwise unchanged.
- **\`mcp_admin_routes.py\`**: registry-view JSON now exposes \`transport\`, \`url\`, and a redacted view of \`headers\` (keys only) for HTTP entries.
- **\`ToolsPage.jsx\`**: renders an HTTP-transport block (transport / url / header-key list) for remote entries.
- **\`mcp_servers.example.json\` + \`CLAUDE.md\`**: ragflow example + docs covering availability semantics and the secrets caveat (no env-var interpolation yet — don't commit production tokens).
- **\`tests/test_mcp_config.py\`**: 25 unit tests covering both transports' validation, normalization, the \`type: \"remote\"\` alias, chat-availability rules, and file-load round-trip.

## Verification

- \`pytest dashboard/tests/\` — **219/219** (25 new).
- \`npm test\` — **81/81**, \`npm run build\` clean.
- **Live**: I ran the new \`_open_streams\` against \`http://localhost:9382/mcp\` and got a successful \`initialize\` (server \`ragflow-mcp-server\` v1.19.0) plus a tool list (\`ragflow_retrieval\`).

## Test plan

- [ ] Restart the dashboard.
- [ ] Add the ragflow entry above to \`dashboard/mcp_servers.json\`.
- [ ] Reload the Tools page — entry should show up with the http transport block (transport: http, url: …).
- [ ] Enable ragflow in a chat conversation; ask a question that should trigger \`ragflow_retrieval\`; confirm the tool fires and returns chunks.
- [ ] Existing stdio entries (\`websearch\`) keep working unchanged.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -123,8 +123,12 @@ config live outside this repo) are declared in a JSON file on disk,
 default `dashboard/mcp_servers.json` (gitignored). Override via
 `LLM_DOCK_MCP_SERVERS_FILE`.
 
-Shape — one entry per server id, top-level object keyed by id. See
-`dashboard/mcp_servers.example.json` for a working example.
+Shape — one entry per server id, top-level object keyed by id. Two
+transports are supported: **stdio** (local subprocess, the default) and
+**http** (remote MCP over the SDK's streamable-HTTP transport). See
+`dashboard/mcp_servers.example.json` for working examples of both.
+
+Stdio (local subprocess):
 
 ```json
 {
@@ -140,15 +144,48 @@ Shape — one entry per server id, top-level object keyed by id. See
 }
 ```
 
+HTTP (remote MCP):
+
+```json
+{
+  "ragflow": {
+    "enabled": true,
+    "name": "RagFlow",
+    "description": "Query the local RagFlow knowledge base over HTTP",
+    "transport": "http",
+    "url": "http://localhost:9382/mcp",
+    "headers": { "Authorization": "Bearer ..." },
+    "icon": "fa-cloud",
+    "tool_hint": "..."
+  }
+}
+```
+
 Rules enforced at load time:
 
-- `command` must be an **absolute path**. No `shell=True`, no PATH lookup.
 - `id` must not contain `__` (collides with `server_id__tool_name`
   namespacing in `mcp_client.py`).
 - `id` must not collide with a built-in (`sympy-math`, `schemdraw-circuits`,
   `render-html`). Built-ins always win.
-- All of `name`, `description`, `command`, `args`, `icon`, `tool_hint` are
-  required; `enabled` defaults to `true`.
+- `name`, `description`, `icon`, `tool_hint` are always required.
+- `enabled` defaults to `true`.
+- `transport` defaults to `"stdio"`. Accepted values: `"stdio"`, `"http"`.
+  `"type": "remote"` (Claude Code's shape) is accepted as a synonym for
+  `transport: "http"`, so configs from other harnesses can be pasted in
+  unchanged.
+- For **stdio** entries: `command` (absolute path; no `shell=True`, no
+  PATH lookup) and `args` (list of strings) are required.
+- For **http** entries: `url` (must start with `http://` or `https://`)
+  is required. `headers` is an optional object of string→string and is
+  sent on every request — values are plain strings (no env-var
+  interpolation yet, so don't commit production secrets via this file).
+
+Chat-side availability differs by transport: stdio entries are hidden
+from chat until the command path exists on disk; http entries are
+considered available as soon as they're enabled (the remote is not
+probed at startup). A misconfigured remote therefore surfaces as a
+tool-call error at chat time rather than as a missing entry on the
+Tools page.
 
 Editing the file:
 
--- a/dashboard/chat/mcp_admin_routes.py
+++ b/dashboard/chat/mcp_admin_routes.py
@@ -51,11 +51,22 @@ def _registry_view(state: dict) -> dict:
             "enabled": bool(cfg.get("enabled", True)),
         }
         if not is_builtin:
-            cmd = cfg["command"]
-            entry["command"] = cmd[0]
-            entry["args"] = cmd[1:]
-            entry["command_exists"] = bool(cmd) and os.path.exists(cmd[0])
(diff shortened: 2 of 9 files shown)


teo-mateo:

- `dashboard/chat/mcp_admin_routes.py:58` returns `dict(cfg.get("headers") or {})` from `GET /api/chat/mcp-registry`, so HTTP MCP header values such as `Authorization: Bearer ...` are included in the registry payload. The PR description says this view is redacted to header keys only, and the route module comment says these handlers do not echo secrets back to the client. The frontend only needs `Object.keys(server.headers)`, so please return a redacted shape here, for example `{key: true}`/`{key: "redacted"}` or a `header_keys` list, and add a test that a configured header value is absent from `_registry_view`/the endpoint response.

teo-mateo:

**Finding 1.1 — Header values leaked through /api/chat/mcp-registry**: accepted. Fixed in commit 3d7fe24.

The registry view now exposes only the sorted list of header NAMES as `header_keys`. The `headers` value-map never leaves disk:

```python
entry["header_keys"] = sorted(list((cfg.get("headers") or {}).keys()))
```

Tools page now renders \`headers: Authorization, X-Tenant (2, values not shown)\` instead of the previous full-value dict.

Regression test in `tests/test_mcp_admin_routes.py::test_registry_view_redacts_http_header_values` configures \`Authorization: Bearer super-secret-token-do-not-leak\` and asserts:

- the literal secret value is not in the response body anywhere,
- the raw \`"headers":\` field is not present in the public shape,
- only the header key NAMES come through, sorted.

Backend tests 222/222 (3 new), frontend 81/81, build clean.

teo-mateo:

- `dashboard/requirements.txt:10` still allows `mcp[cli]>=1.0.0`, but this PR imports `mcp.client.streamable_http` at module import time in `dashboard/chat/mcp_client.py:11`. A deployment with an existing MCP SDK version satisfying the current requirement but older than the streamable HTTP client will fail to import the chat subsystem before any config is loaded. I verified this with a clean env: `mcp==1.0.0` raises `ModuleNotFoundError: No module named "mcp.client.streamable_http"`; `mcp==1.8.0` has the module. Please bump the lower bound to the first supported SDK version, or otherwise guard the import so stdio-only installs do not break.

teo-mateo:

I reviewed the current PR state, including the full diff, prior Codex iterations and author responses, and the checked-out branch. The earlier header redaction issue is addressed with `header_keys`, and the MCP SDK lower bound is now bumped to cover the unconditional `mcp.client.streamable_http` import.

I did not find additional correctness, security, API contract, production-performance, or test-gap issues worth blocking on.

Verification run locally:

- `cd dashboard && venv/bin/pytest tests/` — 222 passed

PR #51, opened 2026-05-23
fix(chat): inline text-file attachments into outgoing message (#50)

### LINKED ISSUE #50: Attaching text files does not work
the text files are not sent to the model. I tried with a .md file.

teo-mateo:
## Summary
- Fixes #50 — attaching a `.md` (or any other text file) no longer silently drops the content. `ChatInput` now reads text-file contents client-side via `FileReader.readAsText` and appends each attachment to the outgoing message as a fenced code block with a filename header and language hint derived from the extension.
- 512 KB per-file inline cap with a visible \`(truncated to …)\` marker; guards against accidentally pasting a multi-MB log into the prompt.
- Drops PDF from the picker / accept list — PDFs need a binary parser the unified artifact store in #28 will own. Until then we'd rather not show an attach chip that pretends to work.
- The previous "affordance only" path that #28 would have replaced is gone; the composer can now send a message containing only attachments (no text, no image).

## Test plan
- [x] `npm test -- --run` — 81/81 frontend tests pass
- [x] `npm run build` — clean
- [ ] Manual: attach a `.md`, hit send, verify the fenced block reaches the model
- [ ] Manual: drag-and-drop a `.py` and a `.json` together
- [ ] Manual: attach a >512 KB file, verify the truncation marker
- [ ] Manual: paste a copied file from the OS file manager

### DIFF
--- a/dashboard/frontend/src/components/chat/ChatInput.jsx
+++ b/dashboard/frontend/src/components/chat/ChatInput.jsx
@@ -46,6 +46,20 @@ function langHintFromName(name) {
   return ext
 }
 
+// Pick a fence that won't be closed early by backtick runs inside `content`.
+// CommonMark allows fences of any length ≥ 3, and the closing fence must be at
+// least as long as the opening one — so we pick one strictly longer than the
+// longest internal run.
+export function pickFence(content) {
+  let longest = 0
+  const re = /`+/g
+  let m
+  while ((m = re.exec(content)) !== null) {
+    if (m[0].length > longest) longest = m[0].length
+  }
+  return '`'.repeat(Math.max(3, longest + 1))
+}
+
 function formatSize(bytes) {
   if (bytes == null) return ''
   if (bytes < 1024) return `${bytes} B`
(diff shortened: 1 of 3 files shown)


teo-mateo:

- `dashboard/frontend/src/components/chat/ChatInput.jsx:95` / `dashboard/frontend/src/components/chat/ChatInput.jsx:222`: text attachments are only added to `attachments` inside `FileReader.onload`, but submit does not know a read is pending. If the user has already typed text, selects a file, and immediately presses Enter/clicks send before the async read completes, `buildMessage()` sends the typed text without the attachment; then the later `onload` can add the file chip back into the now-cleared composer. This is still a silent drop path for the exact feature this PR fixes. Track pending reads in state and disable/hold submit until they settle, or add a placeholder attachment before starting the read and only allow send once every attachment has content/error state. This needs a regression test that controls/defer-resolves `FileReader` and submits before `onload`.

- `dashboard/frontend/src/components/chat/ChatInput.jsx:215`: wrapping every attachment in a fixed triple-backtick fence breaks common `.md` and code attachments that themselves contain ``` fences. The embedded fence closes the wrapper early, so the generated prompt no longer has a faithful single attached-file block and following content can be interpreted as regular prompt text instead of file content. Use a fence length longer than the longest backtick run in the content, or another escaping/serialization format, and add a test with an attached markdown file containing a fenced code block.

teo-mateo:

**Finding 1 — Pending FileReader race (silent drop on send-before-onload)**: accepted. Fixed in 7218089. `addTextFile` now pushes a placeholder synchronously with `pending: true`; `canSend` and `handleSubmit` both refuse while any attachment is still reading. The chip shows a spinner during pending. `onerror` drops the placeholder so a failed read doesn't wedge the composer. Regression test in `ChatInput.test.jsx` uses a controllable `FileReader` mock and asserts that Enter/click are no-ops until `onload` resolves, and that the eventually-sent message contains the file body.

**Finding 2 — Triple-backtick collision in attachment fences**: accepted. Fixed in 7218089. Added `pickFence(content)` which scans for the longest backtick run in the body and picks an outer fence one longer (min 3). The attachment-build path uses it for both the opening and closing fence. Test covers a markdown body containing a `\`\`\`` block, asserting the wrapper is `\`\`\`\`markdown` and that the inner fence survives verbatim. Unit cases cover 0/3/4-tick bodies and isolated single backticks.

Separately, 0cc4252 fixes an unrelated UI bug surfaced while reviewing: `@tailwindcss/typography` was rendering inline `<code>` as `` `text` `` (literal backticks via `::before`/`::after`); overridden to `content: none`.

Issue #58, opened 2026-06-21
Rearchitect chat runtime for background runs and mode-independent chat sessions


teo-mateo:
## Summary

Rearchitect the chat subsystem so a chat turn is not owned by the Flask SSE response or by the mounted React chat page.

The goal is to make slow chat turns robust and make future chat modes easier to implement:

- A model/tool run should continue in the background when the browser navigates away or the SSE client disconnects.
- The completed assistant reply, tool calls, artifacts, and parse warnings should be persisted normally for persisted conversations.
- The UI should be able to reattach to an active run or simply refetch the completed conversation later.
- Future modes such as Ghost Chats (#57) should reuse the same runtime while swapping the persistence policy.

This is an architecture issue, not a single bug fix. Please split the implementation into small PRs following the phases below.

---

## Current Problem

Today the backend chat execution path is coupled to the SSE generator in `dashboard/chat/routes.py`.

`_stream_response()` currently does all of these jobs in one function:

- persists the user message
- builds the prompt message array
- appends tool hints and the current-date line
- decides whether to call the plain model stream or MCP tool loop
- forwards stream events as SSE frames
- accumulates assistant content and reasoning
- persists assistant messages
- persists tool calls, artifacts, and parse warnings
- handles partial-save-on-disconnect

Because the Flask generator owns the model stream, a client disconnect can stop the server-side work. This makes slow models and multi-round tool loops fragile.

The frontend has similar coupling:

- `useChat` assumes a mounted chat page owns the active stream.
- `ChatArea` mixes reusable chat UI with persisted-conversation features such as `updateConversation`, `SystemPromptEditor`, `McpToggle`, edit, critique, and reload callbacks.
- `McpToggle` always persists changes to a conversation, which makes in-memory modes like Ghost Chat harder.

---

## Target Architecture

Introduce a backend chat runtime layer that can run independently of any single HTTP response.

Suggested backend modules/files:

- `dashboard/chat/runtime.py`
  - `ChatTurnRequest`
  - `ChatRuntimeEvent`
  - `ChatRunner`
- `dashboard/chat/prompt_builder.py`
  - builds OpenAI-compatible message arrays
  - applies main system prompt
  - applies MCP tool hints
  - applies current-date line
  - supports image content
- `dashboard/chat/event_codec.py`
  - converts runtime events to SSE payloads
  - keeps event formatting consistent across persisted chat, spinoff, ghost chat, and run reattachment
- `dashboard/chat/runs.py`
[...]

PR #61, opened 2026-06-21
Phase 2 (#58): chat_runs storage without background execution

teo-mateo:
## Phase 2 of #58 — `chat_runs` storage (no background execution)

Adds run tracking that lives independently of any HTTP response. **No background runner yet** — that's Phase 3+. This phase is storage + conversation-payload enrichment only.

### New / changed

- **`chat/runs.py`** — `ChatRunStatus` vocabulary (`queued`/`running`/`completed`/`failed`/`cancelled`), `ACTIVE_STATUSES` / `TERMINAL_STATUSES` sets, `is_active()`. Kept import-free so both the DB layer and the future runner can share it without cycles.
- **`models.py`** — `ChatRun` dataclass (`to_dict` + trimmed `active_run_dict`). `Conversation` gains a **non-persisted** `active_run` field, populated by the DB layer and included in `to_dict` (`null` when none).
- **`db.py`**:
  - `chat_runs` table via `CREATE TABLE IF NOT EXISTS` in `SCHEMA_SQL` — idempotent on existing DBs, no `_migrate` change needed for a new table. FK to `conversations` (CASCADE) and `messages` (SET NULL).
  - Methods: `create_chat_run`, `get_chat_run`, `get_active_run_for_conversation`, `list_active_runs`, `update_chat_run_status`, `complete/fail/cancel_chat_run`.
  - `update_chat_run_status` stamps the matching lifecycle timestamp (`running`→`started_at` via `COALESCE` so re-running doesn't move it; `completed`/`failed`→`completed_at`; `cancelled`→`cancelled_at`) and only writes `active_step`/`error` when provided.
  - `list_conversations` (batch, one query) and `get_conversation` enriched with `active_run`.
  - Deletion **explicitly** clears `chat_runs` for the removed conversations — the FK is `ON DELETE CASCADE`, but the in-memory persistent test connection doesn't enable `PRAGMA foreign_keys`, so cleanup can't rely on it.

### Acceptance

- DB migration idempotent (re-running `_init_db` is a no-op for the table). ✅
- Conversation deletion cleans up related runs. ✅
- Existing conversation APIs backward compatible (additive `active_run` key). ✅
- Conversation list includes `active_run: null` when no run is active. ✅

### Tests

```
cd dashboard
pytest tests/test_chat_runs.py tests/test_chat_settings_routes.py   # 40 passed
pytest tests/   # full suite: 256 passed
```

`test_chat_runs.py` covers create/get/update, started_at COALESCE, the complete/fail/cancel helpers, active-run lookup returning only queued/running, terminal runs not active (parametrized), `list_active_runs` across conversations, the `active_run` payload shape (null + populated), and deletion cleanup.

Refs #58 (Phase 2).

### DIFF
--- a/dashboard/chat/db.py
+++ b/dashboard/chat/db.py
@@ -4,7 +4,7 @@ import logging
 from typing import Optional, List, Tuple
 
 from .models import Conversation, Message, Critique, Artifact, ChatRun
-from .runs import ChatRunStatus, ACTIVE_STATUSES, ALL_STATUSES
+from .runs import ChatRunStatus, ACTIVE_STATUSES, TERMINAL_STATUSES, ALL_STATUSES
 
 logger = logging.getLogger(__name__)
 
@@ -491,10 +491,13 @@ class ChatDB:
         list enrichment without reopening.
         """
         placeholders = ",".join("?" for _ in ACTIVE_STATUSES)
+        # rowid (not id) is the chronological tiebreaker: created_at is only
+        # second-precision and ids are random UUIDs, so within one second only
+        # insertion order (rowid) reflects which run is actually newer.
         row = conn.execute(
             f"""SELECT * FROM chat_runs
                 WHERE conversation_id = ? AND status IN ({placeholders})
-                ORDER BY created_at DESC, id DESC LIMIT 1""",
+                ORDER BY created_at DESC, rowid DESC LIMIT 1""",
             (conv_id, *sorted(ACTIVE_STATUSES)),
         ).fetchone()
         return self._row_to_chat_run(row) if row else None
@@ -509,10 +512,11 @@ class ChatDB:
         rows = conn.execute(
             f"""SELECT * FROM chat_runs
                 WHERE status IN ({status_ph}) AND conversation_id IN ({conv_ph})
-                ORDER BY created_at ASC, id ASC""",
+                ORDER BY created_at ASC, rowid ASC""",
             (*sorted(ACTIVE_STATUSES), *conv_ids),
         ).fetchall()
-        # ASC order means the last write per conversation is the most recent run.
+        # ASC by insertion order (rowid) means the last write per conversation
+        # is the most recent run — chronological even within a one-second tie.
         latest = {}
         for row in rows:
             run = self._row_to_chat_run(row)
@@ -561,7 +565,7 @@ class ChatDB:
             placeholders = ",".join("?" for _ in ACTIVE_STATUSES)
             rows = conn.execute(
                 f"""SELECT * FROM chat_runs WHERE status IN ({placeholders})
-                    ORDER BY created_at ASC, id ASC""",
+                    ORDER BY created_at ASC, rowid ASC""",
                 tuple(sorted(ACTIVE_STATUSES)),
             ).fetchall()
             return [self._row_to_chat_run(row) for row in rows]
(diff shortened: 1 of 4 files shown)


teo-mateo:

- [P2] Use a chronological tie-breaker for active runs — `dashboard/chat/db.py:493` / `dashboard/chat/db.py:509`
  Both the single-conversation lookup and the batch enrichment resolve same-second `created_at` ties by text `id` ordering (`id DESC` or last write after `id ASC`). Since `created_at` is only second precision and future run IDs are UUID-like, `id` is deterministic but not chronological. I reproduced this by inserting `zzz-old` and then `aaa-new` in the same second; both `get_active_run_for_conversation()` and `list_conversations()` surfaced `zzz-old` as the active run. That can leave the sidebar attached to a stale queued/running run whenever two runs are created within one second. Please use a real creation-order tie-breaker, for example higher-precision timestamps or SQLite `rowid`, consistently in both queries. The current `test_active_run_picks_most_recent` also misses this because it accepts either run.

- [P2] Guard terminal statuses from being reactivated — `dashboard/chat/db.py:580`
  `update_chat_run_status()` validates only that the requested status is in `ALL_STATUSES`, so a completed/failed/cancelled run can later be set back to `running` and immediately reappear as `active_run`, while keeping terminal timestamps such as `completed_at`. I reproduced `queued -> completed -> running`, after which `get_active_run_for_conversation()` returned that run. This conflicts with the lifecycle contract that terminal states never reactivate and creates an easy stale-worker/cancel race for the next phase. Please add transition validation or a guarded `WHERE` condition so terminal rows cannot move back to active states.

Verification: `./venv/bin/pytest tests/` passes locally (`256 passed`).

teo-mateo:

**Finding 1.1 — chronological tiebreaker for active runs**: accepted. Fixed in commit f609833. Both `_active_run_for` and `_attach_active_runs` (plus `list_active_runs` for consistency) now order by `rowid` instead of text `id` — `created_at DESC, rowid DESC` for the single lookup, `created_at ASC, rowid ASC` (last-wins) for the batch. rowid is monotonic with insertion, so same-second ties resolve chronologically; random-UUID ids never reflect creation order. `test_active_run_picks_most_recent_by_insertion_not_id` now inserts the newer run with an id (`aaa-new`) that sorts *earlier* than the older one (`zzz-old`), so only the rowid tiebreaker returns the correct run; asserts it via both `get_active_run_for_conversation` and the conversation-list `active_run`.

**Finding 1.2 — guard terminal statuses from reactivation**: accepted. Fixed in commit f609833. `update_chat_run_status` now appends `WHERE id = ? AND status NOT IN (<terminal>)`, so a completed/failed/cancelled row is frozen — it can never move back to running and its terminal timestamp is preserved. The call is a harmless no-op that returns the run unchanged (this also gives Phase 5 its "cancel an already-completed run = no-op" semantics for free). Added `test_terminal_run_cannot_be_reactivated`, parametrized over all three terminal states, asserting status/active_step/terminal-timestamp are unchanged and the run is not active afterward.

Verification: `cd dashboard && venv/bin/python -m pytest tests/test_chat_runs.py` → 19 passed; `pytest tests/` → 259 passed.

teo-mateo:

Verified the iteration 1 fixes:
- `get_active_run_for_conversation` and conversation-list enrichment now resolve same-second active-run ties with `created_at` plus `rowid`, and the regression test covers the UUID/id-order case that failed before.
- `update_chat_run_status` now leaves already-terminal rows unchanged via the guarded `WHERE`, preserving terminal timestamps and keeping those runs out of `active_run` / `list_active_runs`.

I did not find new correctness, security, or concrete test-gap findings in this storage-only phase.

Verification:
- `cd dashboard && venv/bin/python -m pytest tests/test_chat_runs.py` -> 19 passed
- `cd dashboard && venv/bin/python -m pytest tests/` -> 259 passed (with an existing logging cleanup warning after pytest exit)

PR #62, opened 2026-06-21
Phase 3 (#58): background ChatRunner callable without Flask

teo-mateo:
## Phase 3 of #58 — Background `ChatRunner` callable without Flask

Adds the job-runner core that executes one model/tool turn and persists the result, fully decoupled from any SSE response. **Additive** — `routes._stream_response` is untouched; Phase 4 wires the runner into the send route.

### New modules

- **`chat/event_bus.py`** — `EventBus`, an in-memory per-run pub/sub for live observers. Thread-safe (worker publishes, request threads drain). **NOT a source of truth** — the DB is. `publish()` is a no-op when nobody is subscribed, so a disconnected observer never affects the run.
- **`chat/runtime.py`** — `ChatRuntimeEvent`, `ChatTurnRequest`, and `ChatRunner`:
  - `run()` marks the run `running` (`active_step="generating"`), builds the prompt from persisted state via `prompt_builder`, drives the plain stream or the MCP tool loop, accumulates content/reasoning/tool-calls/artifacts/parse-warning, and on `done` persists the assistant message + artifacts and marks the run `completed`.
  - Produces the **same final assistant message shape** as `_stream_response` (including the `accumulated_reasoning or done.reasoning_content` rule and the matched-back tool-call results), minus SSE framing and the now-obsolete partial-save-on-disconnect.
  - Model/tool `error`, a stream that ends without `done`, or any exception → run marked `failed` with useful error text, **no completed assistant message** persisted.
  - Runtime events published to the optional event bus.

The user message is assumed already persisted (the caller creates it + the run before invoking the runner); the runner builds messages from `db.get_messages()`.

### Acceptance

- Runner invocable directly from a unit test, no live SSE client. ✅
- Same final assistant message shape; artifacts + parse warnings persisted. ✅
- Failed runs recorded with useful error text; no completed assistant on failure. ✅

### Tests

`tests/test_chat_job_runner.py` uses a **temp file-backed `ChatDB`** (not `:memory:`) per the Phase-3/4 decision — the runner runs on a background thread in Phase 4 and the shared `:memory:` connection (`check_same_thread=True`) would break. Covers: plain success (+ reasoning rule + event order), tool run (calls/results/artifacts/routing), parse-warning persistence, model error → failed, no-`done` → failed, exception → failed, and the optional-bus path.

```
cd dashboard
pytest tests/test_chat_job_runner.py tests/test_partial_save.py   # 12 passed
pytest tests/   # full suite: 266 passed
```

Refs #58 (Phase 3).

### DIFF
--- a/dashboard/chat/db.py
+++ b/dashboard/chat/db.py
@@ -334,8 +334,13 @@ class ChatDB:
             if should_close:
                 self._close_conn(conn)
 
-    def add_message(self, msg: Message) -> Message:
-        conn = self._get_conn()
+    def add_message(self, msg: Message, conn: sqlite3.Connection = None) -> Message:
+        """Insert a message. When `conn` is passed, the caller owns the
+        transaction (no commit/close here) so the insert can be part of a
+        larger atomic unit; otherwise this commits on its own."""
+        should_close = conn is None
+        if conn is None:
+            conn = self._get_conn()
         try:
             conn.execute(
                 """INSERT INTO messages
@@ -345,10 +350,15 @@ class ChatDB:
                  msg.reasoning_content, msg.model_service, msg.images_json, msg.tool_calls_json,
                  msg.parse_warning_json, msg.seq),
             )
-            conn.commit()
-            return self.get_message(msg.id)
+            if should_close:
+                conn.commit()
+                return self.get_message(msg.id)
+            # Shared transaction not yet committed — a re-fetch on a fresh
+            # connection wouldn't see the row, so return the object as-is.
+            return msg
         finally:
-            self._close_conn(conn)
+            if should_close:
+                self._close_conn(conn)
 
     def delete_messages_from_seq(self, conv_id: str, from_seq: int) -> int:
         conn = self._get_conn()
@@ -433,8 +443,12 @@ class ChatDB:
             created_at=row["created_at"],
         )
 
-    def save_artifact(self, artifact: Artifact) -> Artifact:
-        conn = self._get_conn()
+    def save_artifact(self, artifact: Artifact, conn: sqlite3.Connection = None) -> Artifact:
+        """Insert an artifact. When `conn` is passed, the caller owns the
+        transaction (no commit/close) so it can be batched atomically."""
+        should_close = conn is None
+        if conn is None:
+            conn = self._get_conn()
         try:
             conn.execute(
                 """INSERT INTO artifacts
@@ -443,10 +457,12 @@ class ChatDB:
                 (artifact.id, artifact.message_id, artifact.artifact_type,
                  artifact.content, artifact.title, artifact.language),
             )
-            conn.commit()
+            if should_close:
+                conn.commit()
             return artifact
         finally:
-            self._close_conn(conn)
+            if should_close:
+                self._close_conn(conn)
 
     def get_artifacts_for_conversation(self, conv_id: str) -> dict:
         """Returns a dict of message_id -> [Artifact] for all messages in a conversation."""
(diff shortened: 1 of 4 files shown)


teo-mateo:

1. `dashboard/chat/runtime.py:90` / `dashboard/chat/runtime.py:128`: `ChatRunner` ignores whether its run-status transitions actually won before continuing and before persisting the final assistant message. `ChatDB.update_chat_run_status()` intentionally no-ops once a run is terminal, so a run that is cancelled before the worker starts, or cancelled while the stream is still running, still writes the assistant message on `done` and emits `run_completed`; the DB row remains `cancelled`. This will bite the Phase 4 background-thread path as soon as another request thread can cancel/fail a run while the worker is active. The runner needs to stop without writing a completed assistant message unless the run is still the active/running run and the final transition to completed succeeds.

2. `dashboard/chat/runtime.py:142` / `dashboard/chat/runtime.py:151`: exceptions during `_persist_completion()` can leave a completed assistant message behind on a failed run. The method commits the assistant row via `db.add_message()` before saving artifacts and touching the conversation; if `save_artifact()`, `touch_conversation()`, or a later completion step raises, the outer `except` marks the run failed but the assistant message is already persisted. That violates the stated failure contract of “exception -> failed run with no completed assistant message”. The tests cover a stream exception before `done`, but not a persistence exception after the assistant insert; this probably needs an atomic transaction or explicit cleanup/rollback around the assistant+artifact persistence block.

teo-mateo:

Both findings accepted — they're exactly the Phase-4 concurrency hazards worth fixing now. Fixed in commit a363e29.

**Finding 1 — runner ignored whether status transitions won.** Two guards added:
- *Start guard*: `run()` now checks the result of `update_chat_run_status(RUNNING)`. Since that no-ops on a terminal row, a run cancelled before pickup fails the transition → the runner builds no stream, calls no model, writes no message, and emits `run_<status>` (e.g. `run_cancelled`).
- *Completion guard*: see below — a run cancelled mid-stream no longer produces a completed assistant message or `run_completed`.

**Finding 2 — non-atomic persistence left an orphan completed message on failure.** Replaced `_persist_completion` (which committed `add_message` before artifacts/touch) with a single atomic DB method:

`ChatDB.complete_run_with_message(run_id, assistant_msg, artifacts)` — in **one transaction**: assigns seq, inserts the message + artifacts, flips the run to completed via a guarded `UPDATE ... WHERE id=? AND status NOT IN (terminal)`, and touches the conversation. Then:
- If the guarded UPDATE changes 0 rows (run went terminal/cancelled mid-stream) → `conn.rollback()`, return None → **no assistant message written**. Runner emits `run_cancelled`.
- If any insert raises → `conn.rollback()` and re-raise → the runner's outer `except` marks the run failed → **no orphan completed message**.

To let the inserts share one transaction, `add_message` and `save_artifact` gained an optional `conn` param (same pattern as the existing `next_seq`); standalone callers are unchanged (`add_message` still re-fetches via `get_message` on the self-commit path).

New tests:
- `test_persistence_exception_rolls_back_no_assistant` — `save_artifact` throws after the message insert; asserts only the user message remains and the run is `failed`.
- `test_cancelled_before_start_does_no_work` — run cancelled pre-pickup; asserts the stream stub was never consumed and no message written.
- `test_cancelled_mid_stream_writes_no_completed_message` — stub cancels the run between deltas; asserts no completed message, run stays `cancelled`, last event is `run_cancelled`.

Note (in scope for Phase 5): this makes a mid-stream cancel *discard* the result rather than *promptly stop* the model — prompt cooperative cancellation (a token checked between rounds) is Phase 5. The correctness contract (cancelled run never persists a completed message) holds now.

Verification: `cd dashboard && venv/bin/python -m pytest tests/test_chat_job_runner.py tests/test_chat_runs.py tests/test_partial_save.py` → green; `pytest tests/` → 269 passed.

teo-mateo:

I reviewed the current PR branch, the full changed files, the PR diff, and the prior issue comments. The iteration-1 fixes look correct and complete: the runner now checks the start transition before building the stream, and `ChatDB.complete_run_with_message()` keeps message insertion, artifact insertion, terminal-guarded completion, and conversation touch in one sqlite transaction with rollback on a lost terminal guard or any persistence exception.

I also verified the sqlite3 `cursor.rowcount` behavior used by the completion guard: the guarded `UPDATE` returns `1` for a running row completed by this transaction and `0` when the row is already terminal. The `add_message` / `save_artifact` optional-connection refactor preserves standalone behavior; the self-owned `add_message` path still commits and re-fetches via `get_message()`.

Verification run:
- `cd dashboard && venv/bin/python -m pytest tests/test_chat_job_runner.py tests/test_chat_runs.py tests/test_partial_save.py` -> 34 passed
- `cd dashboard && venv/bin/python -m pytest tests/` -> 269 passed

PR #63, opened 2026-06-21
Phase 4 (#58): send/edit start background runs; SSE is an observer

teo-mateo:
## Phase 4 of #58 — Send/edit start background runs; SSE becomes an observer

The `POST .../messages` and `PUT .../messages/<id>` routes no longer own the model work. They persist the user message, create a `chat_run`, start the run on a worker pool, and return an SSE response that **only observes** the run via the event bus. Closing the response (navigation, tab close) detaches the observer; **the run keeps going and the assistant reply is persisted regardless.**

### Changes

- **`chat/run_manager.py`** (new) — `ChatRunManager`:
  - owns a `ThreadPoolExecutor`; `start()` submits a `ChatRunner` turn off the request thread
  - post-completion **auto-title** runs here (published to the bus), so a first-message title is generated + persisted even if the client navigated away
  - **startup recovery**: `recover_interrupted_runs()` marks any `queued`/`running` run from a dead process as `failed` ("Interrupted by dashboard restart")
  - the **SSE observer** translates runtime events back into the *legacy wire format* (via `event_codec`), so the existing frontend is unchanged. A `STREAM_END` sentinel marks end-of-stream separately from `run_completed`, so the trailing `conversation_updated` still gets delivered. Heartbeats injected on idle.
- **`chat/runtime.py`** — `auto_generate_title` moved here (it's a post-run step now, not an SSE-response step).
- **`routes.py`** — removed `_stream_response`, `_with_heartbeat`, the partial-save machinery, and inline title gen (**−270 net lines**). `send_message`/`edit_message` share `_start_run_response`. **One active run per conversation** → a second send/edit while a run is active returns **409**. `init_chat` wires the `EventBus` + `ChatRunManager` and recovers stale runs on startup.

### Acceptance

- Disconnecting from SSE does not stop the run. ✅ (manager-level + HTTP-level tests)
- Existing send-message API usable by the current frontend (identical SSE frames). ✅
- Only one active run per conversation (409). ✅
- Stale `queued`/`running` runs recovered on startup. ✅

**Cancellation note:** a mid-stream cancel currently *discards* the result (terminal-guarded completion from Phase 3) rather than *promptly stopping* the model. Explicit Stop/Cancel APIs + cooperative cancellation are **Phase 5**.

### Tests

- `test_partial_save.py` **retired** — partial-save-on-disconnect no longer exists (runs complete fully in the background). This was the planned Phase-4 retirement.
- `test_chat_stream_events.py` **rewritten** — it previously drove the removed `_stream_response`; now asserts the **same SSE frame contract** through the real send route + observer (raw delta passthrough, ordering, `[DONE]`→`message_saved`, tool frames, artifact, parse_warning, error, first-message `conversation_updated` tail).
- `test_chat_background_runs.py` (new) — run completes after observer detaches mid-stream (manager + HTTP), 409 on concurrent send, completed/failed clears `active_run` + surfaces error, startup recovery.

```
cd dashboard
pytest tests/test_chat_background_runs.py tests/test_chat_stream_events.py   # 11 passed
pytest tests/   # full suite: 269 passed
```

Refs #58 (Phase 4).


teo-mateo:

1. **High: the one-active-run guard is racy.** `send_message` / `edit_message` check `get_active_run_for_conversation()` before the run exists (`dashboard/chat/routes.py:262` and `dashboard/chat/routes.py:306`), but the user message and queued `chat_run` are inserted later in `_start_run_response()` (`dashboard/chat/routes.py:224`-`232`). Two concurrent requests can both observe no active run, compute `next_seq`, then both insert user messages/runs and start workers. I reproduced that interleaving locally by pausing both requests immediately after the active-run check: both POSTs returned 200, two runs were created, and the two user messages both had `seq=1`. That violates the advertised 409 invariant and can corrupt message ordering. The check needs to be enforced atomically with message+run creation, e.g. one DB transaction/conditional insert or a partial unique index for active statuses, with the user-message insert in the same transaction.

2. **Medium: a live but non-draining SSE observer can grow memory without bound.** `EventBus.subscribe()` creates an unbounded `queue.Queue()` (`dashboard/chat/event_bus.py:24`), and `publish()` unconditionally enqueues every event for every subscriber (`dashboard/chat/event_bus.py:50`). If the HTTP client/proxy stops reading but the generator has not closed yet, `observe()` can be blocked yielding a previous frame and no longer drains its queue, while the background worker keeps publishing token deltas. A long response can therefore accumulate one queued object per runtime event until the run ends. Phase 4 intentionally lets the run continue after observer detachment, but live slow observers still need a bounded/drop/coalescing policy or explicit slow-observer detachment.

Verification: `cd dashboard && venv/bin/python -m pytest tests/test_chat_background_runs.py tests/test_chat_stream_events.py tests/test_chat_job_runner.py tests/test_chat_runs.py` passed (40 tests).

teo-mateo:

Both findings accepted. Fixed in commit f6351ee.

**Finding 1 (High) — racy one-active-run guard.** Replaced the TOCTOU pre-check with an atomic DB operation: `ChatDB.create_run_with_user_message(user_msg, run, delete_from_seq=None)` performs, in **one transaction**: (optional edit truncation) → assign seq + insert the user message → `INSERT INTO chat_runs ... SELECT ... WHERE NOT EXISTS (SELECT 1 FROM chat_runs WHERE conversation_id=? AND status IN (active))`. If the conditional insert affects 0 rows, the transaction is rolled back and the method returns None → the route returns 409, with the user message **and any edit truncation** rolled back (so an edit that loses the race never destroys the conversation tail). `send_message`/`edit_message` no longer pre-check. Added `PRAGMA busy_timeout=5000` so the two writers serialize on the write lock instead of erroring `SQLITE_BUSY`.

I chose a conditional insert over a partial unique index deliberately: the index would reject the Phase-2 tests that legitimately create two active runs directly via `create_chat_run` for tiebreaker coverage; the conditional insert enforces the invariant only at the send/edit entry point.

New tests: success; reject-when-active (user message rolled back); **edit-truncation rollback** (3 messages survive a rejected edit); and a **2-thread race** on a file-backed DB asserting exactly one winner, one active run, one user message.

**Finding 2 (Medium) — unbounded observer queue.** `subscribe()` now uses a bounded `queue.Queue(maxsize=2048)`; `publish()` is non-blocking and **evicts the oldest event on overflow**. This caps memory for a live-but-non-draining client while preserving the newest events — crucially the terminal `stream_end`, so the observer still terminates. Dropped middle deltas don't lose data (the DB has the full reply on refetch). The worker thread (publisher) never blocks.

New tests: basic pubsub, publish-to-no-subscribers no-op, unsubscribe stops delivery, and bounded-evict-oldest (publish > maxsize → qsize capped, newest retained, oldest evicted).

Verification: `cd dashboard && venv/bin/python -m pytest tests/` → 277 passed.

teo-mateo:

1. **Medium: `stream_end` can still be dropped when overflow races with a drainer.** In `EventBus.publish()`, after `put_nowait()` raises `Full`, the code evicts one item and retries, but if the observer drains the queue between the failed put and the eviction `get_nowait()`, `get_nowait()` raises `Empty` and the handler breaks without enqueueing the current event (`dashboard/chat/event_bus.py:67`). If that current event is the terminal `STREAM_END`, the observer remains subscribed and `observe()` keeps heartbeating indefinitely instead of closing the SSE stream. `Empty` means there is now room, so this path should retry `put_nowait()` rather than break. I reproduced this interleaving with a full queue and a drainer between the `Full` and eviction steps; the event was absent afterward.

Verification: `cd dashboard && venv/bin/python -m pytest tests/` passed (`277 passed`; pytest emitted a post-run logging teardown error after completion). The new DB race test passed 50 consecutive runs, and the conditional insert plus rollback path appears atomic, including edit truncation rollback.

teo-mateo:

**Finding 1 (Medium) — stream_end dropped when overflow races a drainer**: accepted, real bug. Fixed in commit 244c0e7.

In the evict loop, after `put_nowait` raised `Full`, an `Empty` from the eviction `get_nowait` meant a concurrent drainer had freed space — but the handler `break`ed without enqueuing, dropping the event (fatal if it's `stream_end` → observer never closes). Changed the `Empty` branch from `break` to `pass` so the loop **retries the put**. With a single publisher per run, the retry always succeeds (either the eviction freed a slot or the drainer did), so it can't spin.

Added `test_event_bus_overflow_racing_drainer_still_delivers`: fills the queue to capacity, monkeypatches `get_nowait` so the first eviction empties the queue and raises `Empty` (reproducing exactly your interleaving), then asserts the terminal event is still present after publish. With the old `break` it was absent.

Thanks for the 50-run race-test confirmation and the atomic-insert/rollback verification on finding 1.

Verification: `cd dashboard && venv/bin/python -m pytest tests/` → 278 passed.

teo-mateo:

1. **High: an accepted concurrent send can still persist a stale/duplicate `seq`.** In `create_run_with_user_message`, `user_msg.seq = self.next_seq(...)` runs before the send path owns a write transaction/lock (`dashboard/chat/db.py:661`); the first write happens afterward when the user message is inserted, and the active-run guard is checked after that (`dashboard/chat/db.py:662`-`674`). A competing request can read `next_seq == 1`, get paused, another request can insert its user message at `seq=1` and complete with an assistant at `seq=2`, then the paused request resumes. Because there is no active run by the conditional insert time, it succeeds but commits its user message with the stale `seq=1`. I reproduced this interleaving against current head by pausing one thread immediately after `next_seq`: the final messages were `first-user seq=1`, `second-user seq=1`, `first-assistant seq=2`, with `second-run` active. That corrupts conversation ordering and can build the accepted second run from an ambiguous message order. Acquire the SQLite write lock before sequence assignment (for example `BEGIN IMMEDIATE` at the top of `create_run_with_user_message`, before optional truncation and `next_seq`) or otherwise allocate the sequence under the same writer lock as the guard.

Verification:
- The latest `event_bus.py` retry-on-`Empty` fix looks correct; `publish()` now retries the put instead of dropping the current event.
- `cd dashboard && venv/bin/python -m pytest tests/test_chat_background_runs.py tests/test_chat_stream_events.py tests/test_chat_runs.py tests/test_event_codec.py` passed (`45 passed`).

teo-mateo:

**Finding 1 (High) — stale/duplicate seq under concurrent send**: accepted, real race. Fixed in commit f5f0f54.

Root cause: `create_run_with_user_message` read `next_seq` under only a read lock (deferred transaction), so a paused request could commit a stale seq after a competing request advanced the tail and its run went terminal — the active-run guard doesn't catch that ordering.

Fix: allocate the sequence under the **write** lock. The method now drives the transaction manually (`isolation_level=None`, restored in `finally` since the in-memory test DB reuses one persistent connection) and issues `BEGIN IMMEDIATE` **before** the optional truncation and `next_seq`, so seq allocation, the `WHERE NOT EXISTS` active-run guard, and the inserts all hold the same writer lock. With `PRAGMA busy_timeout=5000`, a second writer waits for the first instead of erroring, then reads the advanced `next_seq`.

Regression test `test_concurrent_create_complete_cycles_keep_seqs_unique`: 16 threads race full create→complete turns (retrying past 409s); asserts the final message seqs are exactly `1..2N`, unique and contiguous. Without `BEGIN IMMEDIATE` this produces duplicate seqs. Ran it repeatedly (5×) — stable.

Verification: `cd dashboard && venv/bin/python -m pytest tests/` → 279 passed.

teo-mateo:

I reviewed the latest head (`f5f0f54`) against the prior Phase 4 findings and the pushed fix.

The manual transaction in `create_run_with_user_message` looks correct: it switches to `isolation_level=None`, takes `BEGIN IMMEDIATE` before truncation/`next_seq`, uses the shared connection helpers without nested commit/autocommit, explicitly commits or rolls back on every path, returns `None` only after `ROLLBACK`, and restores the previous isolation level in `finally` (including the persistent `:memory:` connection case). The post-commit `get_chat_run()` call while still in autocommit mode is safe, and I do not see a leaked transaction path.

I also do not see a deadlock/lock-leak issue from `BEGIN IMMEDIATE` plus `busy_timeout`; the write-locked section is short and contains only SQLite statements, while the worker/model work remains outside it. The bounded observer queue and retry-on-`Empty` fix also look correct.

Verification run:
- `cd dashboard && venv/bin/python -m pytest tests/test_chat_background_runs.py tests/test_chat_stream_events.py tests/test_chat_runs.py tests/test_event_codec.py` -> 46 passed
- Re-ran `tests/test_chat_runs.py::test_concurrent_create_complete_cycles_keep_seqs_unique` 10x -> passed each time

PR #64, opened 2026-06-21
Phase 5 (#58): run observation + cancellation APIs

teo-mateo:
## Phase 5 of #58 — Run observation + cancellation APIs

Adds navigation-independent run APIs and **cooperative mid-stream cancellation** so Stop actually halts the model — not just a frontend `AbortController.abort()` that leaves the run going.

### Routes

- **`GET /api/chat/runs/<id>`** — run status + metadata (404 if missing).
- **`GET /api/chat/runs/<id>/stream`** — reattach to a run. A terminal run emits a single `run_status` frame and closes (nothing live to replay — the conversation refetch carries the saved content); an active run subscribes to the bus and observes like the send path.
- **`POST /api/chat/runs/<id>/cancel`** — cancel server work. Cancelling an already-terminal run is a harmless **200 no-op** with the existing status (terminal-guarded in the DB). *(Chose 200-no-op over 409, per the issue's "choose one".)*

### Mechanism

- **`ChatRunManager.request_cancel`** sets a per-run cooperative cancel flag (a `threading.Event` registered for the worker's lifetime) **and** marks the run cancelled in the DB. Even if the worker doesn't check in time, the existing terminal-guarded completion (Phase 3) guarantees no assistant message is persisted for a cancelled run.
- **`ChatRunner.run(cancel_check=...)`** polls between stream events; on cancel it `stream.close()`s the upstream generator (GeneratorExit stops `stream_chat_completion`/`stream_with_tools` and the tool loop at the next yield), marks the run cancelled, and persists nothing. Cancellation is **cooperative** — a fully-silent generation is interrupted only once the next chunk/event arrives (same caveat the old `_with_heartbeat` had; true interruption would need a cancel token inside the HTTP read).
- **`observe()` DB-terminal backstop** — a reattaching observer that subscribed *after* `STREAM_END` was published would otherwise heartbeat forever; it now closes on the durable run status when idle.

### Acceptance

- Stop calls cancel (backend), not just `abort()`. ✅
- Cancelling marks the run `cancelled`; it clears `active_run` (no stuck sidebar). ✅
- Cancelling an already-completed run is harmless (200 no-op). ✅

### Tests (`test_chat_run_routes.py`)

auth on all three routes; GET metadata / 404; cancel active → cancelled (+ clears `active_run`); cancel completed → 200 no-op; cancel 404; **cooperative mid-stream cancel** stops the model and writes no assistant message; terminal-stream `run_status` frame; observe backstop closes on terminal.

```
cd dashboard
pytest tests/test_chat_run_routes.py tests/test_chat_background_runs.py   # green
pytest tests/   # full suite: 289 passed
```

Refs #58 (Phase 5).

### DIFF
--- a/dashboard/chat/llm_proxy.py
+++ b/dashboard/chat/llm_proxy.py
@@ -134,6 +134,7 @@ def stream_chat_completion(service_name: str, messages_array: list, tools: list 
     pending_emitted = set()    # tool-call indices we've already announced via tool_call_pending
     finish_reason = None
 
+    resp = None
     try:
         resp = requests.post(url, json=payload, headers=headers, stream=True, timeout=300)
         resp.encoding = "utf-8"
@@ -228,3 +229,10 @@ def stream_chat_completion(service_name: str, messages_array: list, tools: list 
     except Exception as e:
         logger.exception("Unexpected error during streaming")
         yield ("error", {"message": f"Unexpected error: {str(e)}"})
+    finally:
+        # Always release the upstream socket — including when the generator is
+        # closed early (cooperative cancellation calls stream.close(), which
+        # raises GeneratorExit here). Without this, a cancelled run leaves the
+        # model-server connection alive until garbage collection.
+        if resp is not None:
+            resp.close()
--- a/dashboard/chat/routes.py
+++ b/dashboard/chat/routes.py
@@ -12,10 +12,10 @@ from .constants import DEFAULT_MAIN_SYSTEM_PROMPT, DEFAULT_SIDEKICK_SYSTEM_PROMP
 from .llm_proxy import stream_chat_completion
 from .critique import build_critique_context, request_critique, validate_annotations
 from .mcp_registry import list_available_servers
-from .event_codec import DONE, encode_sse, encode_sse_delta
+from .event_codec import DONE, encode_sse, encode_sse_event, encode_sse_delta
 from .event_bus import EventBus
 from .run_manager import ChatRunManager
-from .runs import ChatRunStatus
+from .runs import ChatRunStatus, TERMINAL_STATUSES
 from . import settings_store
 
 logger = logging.getLogger(__name__)
@@ -324,6 +324,61 @@ def edit_message(conv_id, msg_id):
     # replaces an existing turn, so it never triggers auto-title.
     return _start_run_response(db, conv, new_msg, mcp_mgr, is_first=False,
                                first_user_content=content, delete_from_seq=msg.seq)
+
+
+# -- Runs (observation + cancellation, issue #58) --
+
+
+@chat_bp.route("/api/chat/runs/<run_id>", methods=["GET"])
+@require_auth
+def get_run(run_id):
+    run = _get_db().get_chat_run(run_id)
+    if run is None:
+        return jsonify({"error": "Run not found"}), 404
+    return jsonify(run.to_dict())
+
+
+@chat_bp.route("/api/chat/runs/<run_id>/stream", methods=["GET"])
+@require_auth
+def stream_run(run_id):
+    """Reattach to a run's live event stream.
+
+    If the run is already terminal, emit a single run_status frame and close
+    (there is nothing live to replay — the conversation refetch carries the
+    saved content). Otherwise subscribe to the bus and observe like the send
+    path; the observer's DB backstop closes the stream even if the run finishes
+    between this check and the subscribe.
+    """
+    db = _get_db()
+    manager = _get_run_manager()
+    run = db.get_chat_run(run_id)
+    if run is None:
+        return jsonify({"error": "Run not found"}), 404
+
+    headers = {"Cache-Control": "no-cache", "X-Accel-Buffering": "no"}
+
+    if run.status in TERMINAL_STATUSES:
+        def terminal():
+            yield encode_sse_event("run_status", {"status": run.status, "error": run.error})
+        return Response(terminal(), mimetype="text/event-stream", headers=headers)
+
+    q = manager.subscribe(run_id)
+    return Response(stream_with_context(manager.observe(run_id, q)),
+                    mimetype="text/event-stream", headers=headers)
+
+
+@chat_bp.route("/api/chat/runs/<run_id>/cancel", methods=["POST"])
+@require_auth
+def cancel_run(run_id):
+    """Cancel a run's server-side work, independent of any SSE observer.
+
+    Cancelling an already-terminal run is a harmless no-op that returns 200
+    with the run's existing status (terminal-guarded in the DB).
+    """
+    run = _get_run_manager().request_cancel(run_id)
+    if run is None:
+        return jsonify({"error": "Run not found"}), 404
+    return jsonify(run.to_dict())
 
 
 # -- Critique --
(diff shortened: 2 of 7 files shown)


teo-mateo:

- [P1] Stop still only aborts the observer; it never calls the new cancel API. The backend route exists, but the React Stop path remains `abortRef.current.abort()` and local state cleanup at `dashboard/frontend/src/hooks/useChat.js:342`. `streamChat` also never receives/handles a run id, while `_sse_frames_for` explicitly drops `run_started`, so the client has no id to post to `/api/chat/runs/<id>/cancel`. Repro: start a long chat in `/v2`, click Stop; the SSE fetch closes but the background run keeps going and can still persist the assistant message. This misses the Phase 5 "Stop calls cancel" behavior. Please wire the current run id through the send/observe stream and POST cancel before or alongside aborting the observer.

- [P2] Cancellation does not deterministically close the upstream HTTP stream. `ChatRunner.run()` calls `stream.close()` on cancellation (`dashboard/chat/runtime.py:166`), but the actual streaming `requests` response opened in `stream_chat_completion` is never closed in a `finally`/context manager (`dashboard/chat/llm_proxy.py:138`, iterated at `dashboard/chat/llm_proxy.py:145`). Generator close exits the loop, but no code path calls `resp.close()`, so a cancelled run can leave the model-server socket/request alive until object finalization. Wrap the streaming request with `with requests.post(...) as resp:` or a `finally: resp.close()` so cooperative cancel reliably tears down the upstream connection.

Checked: `gh pr view 64`, `gh pr diff 64`, issue/review comments, checked out the branch, read the full changed files plus DB/event helpers, and ran `source dashboard/venv/bin/activate && pytest tests/test_chat_run_routes.py tests/test_chat_background_runs.py` (16 passed).

teo-mateo:

**Finding 2 (P2) — upstream response not closed on cancel**: accepted, real leak. Fixed in commit cb26ae1. `stream_chat_completion` now initializes `resp = None` and closes it in a `finally`, so the model-server socket is released on every exit path — including the cooperative-cancel path where `stream.close()` raises GeneratorExit into the generator. New `test_llm_proxy_stream.py`: asserts `resp.close()` is called on normal completion and on early generator close (monkeypatched `requests.post` returning a fake response with a close spy).

**Finding 1 (P1) — Stop doesn't call cancel**: backend half accepted, frontend half is Phase 6.
- *Accepted (Phase 5)*: you're right the client had no run id — `run_started` was dropped by `_sse_frames_for`. The observer now emits a `run_started` SSE frame carrying `run_id` as the **first frame** of the stream. The current frontend safely ignores unknown SSE `type`s (it dispatches on known types and otherwise looks for `choices[].delta`), so this doesn't disturb the existing UI. Added a test asserting run_started leads the stream with a valid run id.
- *Push back (Phase 6)*: wiring the React Stop button to capture that id and POST `/runs/<id>/cancel` belongs to Phase 6 ("Frontend Session Refactor"), whose acceptance explicitly lists the frontend test *"Stop button calls the cancel API for active persisted runs."* Phase 5 is the backend API + exposing the run id; Phase 6 owns the `useChat`/`usePersistedChatSession` refactor where that handler lives. Doing the hook wiring here would be rewritten by Phase 6's refactor. The backend now exposes everything Phase 6 needs (cancel API + run id on the wire).

Verification: `cd dashboard && venv/bin/python -m pytest tests/` → 292 passed.

teo-mateo:

- [P2] Guarantee `run_started` before any heartbeat on queued send streams. The PR now publishes `run_started` from inside the worker (`ChatRunner.run()`), but the HTTP observer starts immediately after `ThreadPoolExecutor.submit()`. If all chat workers are occupied for longer than `HEARTBEAT_INTERVAL_S`, `observe()` hits the empty-queue heartbeat path before the queued worker ever publishes `run_started`, so the first SSE frame is `heartbeat` rather than the promised run id. I reproduced this with `max_workers=1`: start one blocking run, start a second run, then `next(manager.observe(second_run_id, q))` returns `data: {"type": "heartbeat", ...}`. Since the run id already exists before submit, the send-path stream should synthesize/emit `run_started` before entering the bus observer, or otherwise publish it synchronously before the worker queue can delay it, while avoiding a duplicate when the worker starts. This keeps the Phase 5 backend contract true under normal worker saturation and gives Phase 6 a run id immediately for queued-run cancellation.

Checked `gh pr view 64`, `gh pr diff 64`, issue comments, checked out the branch, read the changed files in full plus the run DB/event helpers, and ran `dashboard/venv/bin/python -m pytest dashboard/tests/test_chat_run_routes.py dashboard/tests/test_chat_background_runs.py dashboard/tests/test_chat_stream_events.py dashboard/tests/test_llm_proxy_stream.py` (24 passed). The `llm_proxy` `resp = None` / `finally: resp.close()` fix looks correct for normal completion, yielded errors, early `GeneratorExit`, and `requests.post` raising before a response exists.

teo-mateo:

**Finding (P2) — run_started not guaranteed before heartbeat under worker saturation**: accepted, real gap. Fixed in commit 3f2020e.

The run id is known at submit time, so I moved run_started emission out of the worker and into the observer: `observe()` now `yield`s a synthesized `run_started` frame (with the run id) **before the loop and any heartbeat**, and `_sse_frames_for` suppresses the worker's own `run_started` to avoid a duplicate. The client now always receives the run id as the first frame — even for a still-queued run whose worker hasn't started — so Phase 6 can also cancel queued runs.

Test `test_run_started_emitted_first_even_when_workers_saturated`: `max_workers=1`, block the single worker with run A, then assert the queued run B's `observe()` yields `run_started` (with B's id) as its first frame rather than a heartbeat. Reproduces your interleaving deterministically. Updated the backstop test for the new leading frame.

Thanks for confirming the `resp.close()` finally is correct across all paths.

Verification: `cd dashboard && venv/bin/python -m pytest tests/` → 293 passed.

teo-mateo:

No new correctness, race, security, or test-gap findings in the Phase 5 backend changes at `3f2020e`.

I verified the latest fix against the actual code: `observe()` yields the synthetic `run_started` frame before queue polling or heartbeats, `_sse_frames_for()` suppresses the worker-published `run_started`, so there is no duplicate client-visible `run_started`. The send and active-reattach paths use that observer, and the already-terminal reattach route still emits a single `run_status` frame and closes. The saturated-worker test is sound for the previous race: with `max_workers=1`, run A blocks the only worker while run B is still queued, and `observe(run_b.id, q)` must return `run_started` first.

Checked with `gh pr view 64`, `gh pr diff 64`, `gh api repos/teo-mateo/llm-dock/issues/64/comments --paginate`, checked out the branch, and read the changed files plus DB/event helpers. Focused verification: `dashboard/venv/bin/python -m pytest dashboard/tests/test_chat_run_routes.py dashboard/tests/test_chat_background_runs.py dashboard/tests/test_chat_stream_events.py dashboard/tests/test_llm_proxy_stream.py dashboard/tests/test_chat_job_runner.py dashboard/tests/test_chat_runs.py` → 59 passed. I also ran `dashboard/venv/bin/python -m pytest dashboard/tests`; it got 291 passed / 2 failed, both in `test_services_sse.py` because this local checkout cannot read `../docker-compose.yml`, which is outside the Phase 5 chat changes.

Issue #57, opened 2026-06-21
Ghost chats: zero-trace ephemeral conversations


teo-mateo:
## Summary

Add **Ghost Chats** — conversations that leave absolutely no trace: never persisted on the server (no SQLite writes), never cached on disk in the browser (no localStorage, no HTTP cache, no service worker). The idea is to provide a truly ephemeral, deniable chat mode for sensitive topics where the user wants no record of the conversation surviving anywhere after the tab closes.

---

### 1. Backend: Ghost Chat Endpoint (no persistence)

**New endpoint:** `POST /api/chat/ghost`

- Same SSE streaming as `send_message` and `/api/chat/spinoff`, but **never calls** `db.add_message()`, `db.save_artifact()`, or `db.touch_conversation()`.
- No conversation ID is needed — the endpoint accepts a `messages_array` directly (similar to `/api/chat/spinoff`) or a single user message + service name.
- Supports MCP tools if provided, but never persists tool results.
- Response is a pure SSE stream with the same event types (`delta`, `tool_call`, `tool_result`, `artifact`, `done`, `error`) but no `message_saved` or `conversation_updated` events.

**Relevant file:** `dashboard/chat/routes.py`

A new route handler in the `chat_bp` blueprint:

```python
@chat_bp.route("/api/chat/ghost", methods=["POST"])
@require_auth
def ghost_chat():
    """Stateless ghost chat — no persistence, no trace on disk."""
    data = request.get_json() or {}
    service_name = data.get("service_name")
    messages_array = data.get("messages", [])
    # ... SSE stream without DB writes
```

The existing `_stream_response()` helper in `routes.py` always persists. A new helper `_stream_ghost()` is needed that yields the same SSE events but skips all `db.*` calls. The `_auto_generate_title` logic is also skipped.

### 2. Backend: Ensure no caching headers

**Add to the SSE response:**

- `Cache-Control: no-store, max-age=0, must-revalidate`
- `Pragma: no-cache`
- `Expires: 0`

This prevents the browser and any intermediate proxies from caching the SSE response body on disk.

**Relevant file:** `dashboard/chat/routes.py`

### 3. Frontend: Ghost Chat Hook (`useGhostChat`)

**New file:** `dashboard/frontend/src/hooks/useGhostChat.js`

A new React hook similar to `useChat` but:

- Uses the `/api/chat/ghost` endpoint instead of `/chat/conversations/<id>/messages`
[...]

PR #65, opened 2026-06-21
Phase 6 (#58): frontend session refactor — background-run aware UI

teo-mateo:
Part of #58 (rearchitect chat runtime for background runs).

Decouples the chat UI from the SSE response so it **observes** a server-side run rather than owning it. Builds on the Phase 5 run-observation/cancellation APIs.

## Changes
- **services/sse.js** — dispatch the `run_started` frame via `onRunStarted`, so the client learns the backend run id.
- **services/chat.js** — `getRun()` / `cancelRun()` against the Phase 5 endpoints.
- **hooks/useChat.js** — capture the run id from `run_started`; `stopStreaming` now cancels the **server** run (`POST /runs/<id>/cancel`) before aborting the fetch. Navigation (`loadConversation` / unmount) deliberately does **not** cancel — the run finishes and persists in the background.
- **ChatSidebar.jsx** — spinner badge on rows whose conversation has a queued/running `active_run` (data already supplied by `_attach_active_runs`).
- **McpToggle.jsx** — controlled mode: with an `onChange` prop the toggle returns the next server list in-memory and persists nothing (clears the way for non-persisted modes like Ghost Chat #57); otherwise falls back to the persisted PUT path.

## Tests
- `ChatSidebar.test.jsx` — active-run badge for running / queued / none.
- `McpToggle.test.jsx` — controlled (`onChange`) vs persisted (`updateConversation`) modes.
- `useChat.test.jsx` — Stop cancels the captured run; switching conversation does **not** cancel; no-run-id Stop is a cancel no-op.

97 frontend tests pass; `npm run build` clean. Lint unchanged (8 pre-existing errors on base, 0 new).

## Out of scope (Phase 7)
Live sidebar refresh of `active_run` and failed-run error surfacing after refetch.

### DIFF
--- a/dashboard/chat/routes.py
+++ b/dashboard/chat/routes.py
@@ -381,6 +381,26 @@ def cancel_run(run_id):
     return jsonify(run.to_dict())
 
 
+@chat_bp.route("/api/chat/conversations/<conv_id>/cancel-active-run", methods=["POST"])
+@require_auth
+def cancel_active_run(conv_id):
+    """Cancel whatever run is active for a conversation, by conversation id.
+
+    The Stop button uses this so cancellation never depends on the client
+    having captured the run id from the run_started frame: the server creates
+    the run when the turn starts, so it is always findable from the
+    conversation. Returns {"run": <run>} when a run was cancelled, or
+    {"run": null} (still 200) when the conversation has no active run — a
+    harmless no-op (e.g. it already finished). 404 only if the conversation
+    does not exist.
+    """
+    db = _get_db()
+    if db.get_conversation(conv_id) is None:
+        return jsonify({"error": "Conversation not found"}), 404
+    run = _get_run_manager().request_cancel_for_conversation(conv_id)
+    return jsonify({"run": run.to_dict() if run is not None else None})
+
+
 # -- Critique --
 
 @chat_bp.route("/api/chat/messages/<msg_id>/critique", methods=["POST"])
--- a/dashboard/chat/routes.py
+++ b/dashboard/chat/routes.py
@@ -389,15 +389,25 @@ def cancel_active_run(conv_id):
     The Stop button uses this so cancellation never depends on the client
     having captured the run id from the run_started frame: the server creates
     the run when the turn starts, so it is always findable from the
-    conversation. Returns {"run": <run>} when a run was cancelled, or
-    {"run": null} (still 200) when the conversation has no active run — a
-    harmless no-op (e.g. it already finished). 404 only if the conversation
+    conversation.
+
+    An optional {"expected_run_id": "<id>"} body guards against a stale Stop:
+    if the run the client meant to stop has already finished and a newer run is
+    active in this conversation, the cancel is a no-op rather than killing the
+    newer run.
+
+    Returns {"run": <run>} when a run was cancelled, or {"run": null} (still
+    200) when the conversation has no active run, or its active run does not
+    match expected_run_id — both harmless no-ops. 404 only if the conversation
     does not exist.
     """
     db = _get_db()
     if db.get_conversation(conv_id) is None:
         return jsonify({"error": "Conversation not found"}), 404
-    run = _get_run_manager().request_cancel_for_conversation(conv_id)
+    body = request.get_json(silent=True) or {}
+    expected_run_id = body.get("expected_run_id")
+    run = _get_run_manager().request_cancel_for_conversation(
+        conv_id, expected_run_id=expected_run_id)
     return jsonify({"run": run.to_dict() if run is not None else None})
 
 
--- a/dashboard/chat/run_manager.py
+++ b/dashboard/chat/run_manager.py
@@ -119,6 +119,21 @@ class ChatRunManager:
             ev.set()
         return self.db.cancel_chat_run(run_id)
 
+    def request_cancel_for_conversation(self, conversation_id):
+        """Cancel a conversation's active (queued/running) run by conversation id.
+
+        The Stop button uses this so it never depends on the client having
+        captured the run id: the server already created the run when the turn
+        started, so it can always be found from the conversation. Returns the
+        cancelled run, or None if the conversation has no active run (a harmless
+        no-op — e.g. the run already finished). Delegates to request_cancel so
+        the cooperative flag is set and the DB update stays terminal-guarded.
+        """
+        run = self.db.get_active_run_for_conversation(conversation_id)
+        if run is None:
+            return None
+        return self.request_cancel(run.id)
+
     def _execute(self, conv, run, mcp_manager, is_first, first_user_content):
         cancel_event = threading.Event()
         with self._flags_lock:
--- a/dashboard/chat/run_manager.py
+++ b/dashboard/chat/run_manager.py
@@ -119,7 +119,7 @@ class ChatRunManager:
             ev.set()
         return self.db.cancel_chat_run(run_id)
 
-    def request_cancel_for_conversation(self, conversation_id):
+    def request_cancel_for_conversation(self, conversation_id, expected_run_id=None):
         """Cancel a conversation's active (queued/running) run by conversation id.
 
         The Stop button uses this so it never depends on the client having
@@ -128,9 +128,20 @@ class ChatRunManager:
         cancelled run, or None if the conversation has no active run (a harmless
         no-op — e.g. the run already finished). Delegates to request_cancel so
         the cooperative flag is set and the DB update stays terminal-guarded.
+
+        expected_run_id guards against a stale Stop cancelling the wrong run: if
+        the caller knows which run it meant to stop (captured from the
+        run_started frame), a late-arriving cancel whose target has since
+        finished — with a *newer* run now active in the same conversation — must
+        not cancel that newer run. When expected_run_id is given and the active
+        run is a different one, this is a no-op. When it's omitted (a genuine
+        early Stop before run_started), the active run is provably still the one
+        the user meant, so "cancel whatever is active" is safe.
         """
         run = self.db.get_active_run_for_conversation(conversation_id)
         if run is None:
+            return None
+        if expected_run_id is not None and run.id != expected_run_id:
             return None
         return self.request_cancel(run.id)
 
--- a/dashboard/frontend/src/components/chat/ChatArea.jsx
+++ b/dashboard/frontend/src/components/chat/ChatArea.jsx
@@ -143,6 +143,15 @@ export default function ChatArea({
 
   const targetMessage = critiqueTarget ? messages.find(m => m.id === critiqueTarget) : null
 
+  // A background run can still be active for this conversation even when we
+  // have no local SSE stream (the user navigated away and back, so `streaming`
+  // is false but the run kept going server-side). Treat it as an in-pane busy
+  // state: block the composer/model/MCP controls and surface Stop, which
+  // cancels by conversation id regardless of whether we hold the run stream.
+  const activeRun = conversation.active_run
+  const hasActiveRun = !!activeRun && (activeRun.status === 'running' || activeRun.status === 'queued')
+  const busy = streaming || cancelling || hasActiveRun
+
   function handleOpenCritique(msgId) {
     setCritiqueTarget(msgId)
   }
@@ -198,16 +207,16 @@ export default function ChatArea({
               sidekickService={conversation.sidekick_service}
               onChangeMain={v => handleModelChange('main_service', v)}
               onChangeSidekick={v => handleModelChange('sidekick_service', v)}
-              disabled={streaming || cancelling}
+              disabled={busy}
             />
             <McpToggle
               conversationId={conversation.id}
               enabledServers={conversation.mcp_servers}
               onUpdate={() => onReloadConversation?.(conversation.id)}
-              disabled={streaming || cancelling}
+              disabled={busy}
             />
           </div>
-          {streaming && (
+          {(streaming || hasActiveRun) && !cancelling && (
             <button
               onClick={onStopStreaming}
               className="text-xs px-3 py-1 bg-danger-subtle text-danger-fg border border-danger rounded hover:bg-danger-subtle"
@@ -258,7 +267,7 @@ export default function ChatArea({
           ref={composerRef}
           focusKey={conversation.id}
           onSend={handleSend}
-          disabled={streaming || cancelling || !conversation.main_service}
+          disabled={busy || !conversation.main_service}
           pendingInserts={pendingInserts}
           onClearInsert={(idx) => setPendingInserts(prev => prev.filter((_, i) => i !== idx))}
         />
--- a/dashboard/frontend/src/components/chat/ChatArea.jsx
+++ b/dashboard/frontend/src/components/chat/ChatArea.jsx
@@ -32,6 +32,7 @@ export default function ChatArea({
   streamingParseWarning,
   error,
   cancelling,
+  runReady,
   onSend,
   onEdit,
   onStopStreaming,
@@ -216,7 +217,12 @@ export default function ChatArea({
               disabled={busy}
             />
           </div>
-          {(streaming || hasActiveRun) && !cancelling && (
+          {/* Gate the live-stream Stop on runReady: it must not appear until
+              run_started has delivered the run id, so Stop always cancels with
+              an expected-run guard rather than an unguarded "cancel whatever is
+              active" that a concurrently-started run could be hit by. The
+              returned-to active_run path already has an id (active_run.id). */}
+          {(((streaming && runReady) || hasActiveRun) && !cancelling) && (
             <button
               onClick={onStopStreaming}
               className="text-xs px-3 py-1 bg-danger-subtle text-danger-fg border border-danger rounded hover:bg-danger-subtle"
--- a/dashboard/frontend/src/components/chat/ChatArea.jsx
+++ b/dashboard/frontend/src/components/chat/ChatArea.jsx
@@ -250,6 +250,7 @@ export default function ChatArea({
           hasSidekick={!!conversation.sidekick_service}
           onCritique={handleOpenCritique}
           onEdit={onEdit}
+          disableEdit={busy}
           streaming={streaming}
           streamingContent={streamingContent}
           streamingReasoning={streamingReasoning}
--- a/dashboard/frontend/src/components/chat/ChatArea.jsx
+++ b/dashboard/frontend/src/components/chat/ChatArea.jsx
@@ -31,6 +31,7 @@ export default function ChatArea({
   streamingArtifacts,
   streamingParseWarning,
   error,
+  cancelling,
   onSend,
   onEdit,
   onStopStreaming,
@@ -197,13 +198,13 @@ export default function ChatArea({
               sidekickService={conversation.sidekick_service}
               onChangeMain={v => handleModelChange('main_service', v)}
               onChangeSidekick={v => handleModelChange('sidekick_service', v)}
-              disabled={streaming}
+              disabled={streaming || cancelling}
             />
             <McpToggle
               conversationId={conversation.id}
               enabledServers={conversation.mcp_servers}
               onUpdate={() => onReloadConversation?.(conversation.id)}
-              disabled={streaming}
+              disabled={streaming || cancelling}
             />
           </div>
           {streaming && (
@@ -257,7 +258,7 @@ export default function ChatArea({
           ref={composerRef}
           focusKey={conversation.id}
           onSend={handleSend}
-          disabled={streaming || !conversation.main_service}
+          disabled={streaming || cancelling || !conversation.main_service}
           pendingInserts={pendingInserts}
           onClearInsert={(idx) => setPendingInserts(prev => prev.filter((_, i) => i !== idx))}
         />
--- a/dashboard/frontend/src/components/chat/ChatArea.test.jsx
+++ b/dashboard/frontend/src/components/chat/ChatArea.test.jsx
@@ -236,3 +236,57 @@ describe('ChatArea — send gating on a pending system-prompt save', () => {
     expect(onSend).toHaveBeenCalledWith('first', undefined)
   })
 })
+
+describe('ChatArea — active background run on return', () => {
+  // Regression for codex iteration 7 P2: returning to a conversation whose run
+  // is still going (streaming false, but conversation.active_run set) must keep
+  // the composer blocked and offer Stop — otherwise the run is uncontrollable
+  // and a send just hits the backend 409 active-run guard.
+  function renderWith(activeRun, onStopStreaming = () => {}) {
+    return render(
+      <ChatArea
+        conversation={{
+          id: 'conv-1',
+          main_service: 'vllm-test',
+          main_system_prompt: '',
+          mcp_servers: [],
+          active_run: activeRun,
+        }}
+        awaitingConversation={false}
+        defaultModelName="vllm-test"
+        messages={[]}
+        critiques={{}}
+        setCritiques={() => {}}
+        streaming={false}
+        onStopStreaming={onStopStreaming}
+      />
+    )
+  }
+
+  it('disables the composer and shows Stop for a running active_run with no local stream', () => {
+    const onStop = vi.fn()
+    const { container } = renderWith({ status: 'running' }, onStop)
+    expect(container.querySelector('textarea').disabled).toBe(true)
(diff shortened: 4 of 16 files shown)


teo-mateo:

- P1 — Stop leaves a cancelled turn in optimistic local state. `stopStreaming()` now cancels the backend run and sets `streaming` false, but it never reconciles `messages` after the cancel path (`dashboard/frontend/src/hooks/useChat.js:350`). At that point the backend has already inserted the user message as part of run creation, while cancellation intentionally prevents the assistant message from being written. Locally, though, the optimistic `{ id: 'temp-user' }` remains. If the user sends again, the next save handler removes every `temp-user` from the array before appending the new saved turn (`dashboard/frontend/src/hooks/useChat.js:191`), so the stopped prompt disappears from the visible transcript until a later refetch/reload even though it is still in SQLite. Please reconcile after Stop, for example by refetching the conversation after `cancelRun` succeeds or by replacing/removing only the active optimistic row, and add a Stop-then-send-again regression test.

teo-mateo:

**Finding P1 — Stop leaves a cancelled turn in optimistic local state**: accepted. Fixed in commit 8416ac9.

`stopStreaming` now reconciles after cancelling: once `cancelRun` settles it refetches the conversation (`refetchMessages(convId)`), swapping the optimistic `{ id: 'temp-user' }` row for the persisted message with a real id. Because the stopped prompt no longer carries the `temp-user` id, the next send's `filter(m => m.id !== 'temp-user')` strip can't drop it — closing both the transient flicker and the permanent-disappearance-on-failed-refetch window. The refetch runs in `.finally` after cancel settles so it also reflects the rare case where the run completed in the same instant Stop was pressed.

Added a `Stop reconciles ... a later send keeps the stopped prompt` regression test (asserts the row loses its `temp-user` id and gains the persisted `msg-1` id after Stop). 98 frontend tests pass; build clean.

teo-mateo:

- P2 — Stop can refetch the stopped conversation after the user has already navigated away. `stopStreaming()` captures `convId` and unconditionally runs `refetchMessages(convId)` when `cancelRun` settles (`dashboard/frontend/src/hooks/useChat.js:356`, `dashboard/frontend/src/hooks/useChat.js:371`). `refetchMessages()` then writes `conversation`/`messages` directly (`dashboard/frontend/src/hooks/useChat.js:41`). If the user clicks Stop in conversation A and quickly selects conversation B, B's `loadConversation()` can resolve first and then the cancel-finally refetch for A can win last, leaving the hook state on A while the URL is B. `ChatPage` will render the loading guard for B (`dashboard/frontend/src/components/chat/ChatPage.jsx:149`) and will not automatically re-run `loadConversation(B)` because the route id did not change. Please gate the Stop reconciliation so it only applies if the hook is still observing that conversation, or make conversation fetches request-sequenced, and add a Stop-then-navigate regression test where the old refetch resolves last.

- P2 — A very early Stop click can still leave the backend run executing. The only server cancellation path is guarded by `if (runId)` (`dashboard/frontend/src/hooks/useChat.js:358`), but the observer is aborted regardless immediately afterwards (`dashboard/frontend/src/hooks/useChat.js:373`). Since `streaming` becomes true before the POST has returned and before the first `run_started` frame is parsed, there is a real window where the backend has already created the queued run but `runIdRef.current` is still null. Clicking Stop in that window aborts the SSE fetch without posting `/runs/<id>/cancel`, so the Phase 4 background run continues and may persist an assistant reply after the UI says it stopped. The current no-run-id test only models an early HTTP error. Please either defer aborting until `run_started` arrives and then cancel, expose a cancel-by-active-conversation path, or otherwise disable Stop until cancellation can be targeted, with a regression test where `onRunStarted` is delayed until after Stop is pressed.

teo-mateo:

**Finding P2 — late Stop refetch clobbers a navigated-to conversation**: accepted. Fixed in commit c006e7a. The Stop reconciliation now goes through `cancelAndReconcile(runId, convId)`, whose `.finally` only calls `refetchMessages(convId)` when `conversationRef.current?.id === convId`. `conversationRef` is a mirror of the committed `conversation` updated via effect, so a refetch that resolves after the user navigated A→B is gated out instead of overwriting B. Regression test `a late Stop refetch does not clobber a conversation the user navigated to` holds A's cancel POST open, navigates to B, then resolves the cancel last and asserts the active conversation stays B.

**Finding P2 — early Stop before run_started leaves the run executing**: accepted. Fixed in commit c006e7a. `stopStreaming` now branches: if the run id is known it cancels + aborts as before; if not (Stop pressed before the `run_started` frame), it sets `pendingStopRef` and does **not** abort — aborting would drop the frame carrying the id and orphan the background run. Both `onRunStarted` handlers consume `pendingStopRef`: they capture the id, then immediately `cancelAndReconcile` + abort + clear draining. `pendingStopRef` is reset at the top of every send/edit. Regression test `Stop pressed before run_started defers cancellation until the run id arrives` presses Stop with no id (asserts no cancel yet, UI already stopped), then delivers `run_started` and asserts `cancelRun('run-late')` fires.

100 frontend tests pass; build clean; lint unchanged (8 pre-existing errors, 0 new).

teo-mateo:

- P2 — Deferred early Stop can still be orphaned by navigation or the next send. In the no-run-id branch, `stopStreaming()` only sets `pendingStopRef.current = true` and keeps the fetch alive so `onRunStarted` can later cancel (`dashboard/frontend/src/hooks/useChat.js:422-428`, `dashboard/frontend/src/hooks/useChat.js:209-218`). But it also sets `streaming` false immediately, so the user can switch conversations or submit/edit before that frame arrives. `loadConversation()` aborts the non-draining `abortRef` (`dashboard/frontend/src/hooks/useChat.js:87-89`), dropping the only observer that can receive `run_started`; and `sendMessage`/`editMessage` reset the shared `pendingStopRef` before the old stream's callback runs (`dashboard/frontend/src/hooks/useChat.js:126-130`, `dashboard/frontend/src/hooks/useChat.js:294-298`). In either path, the run the user explicitly stopped is never POSTed to `/runs/<id>/cancel`, so it can continue and persist an assistant reply after the UI said Stop. Please make the deferred-cancel state owned by the original stream/controller (and abort that captured controller, not whatever `abortRef.current` points to later), or add a backend cancel-current-active-run-by-conversation path. A regression test should cover Stop-before-`run_started`, then navigation or a second send before delivering `run_started`, and assert the original run is still cancelled.

teo-mateo:

**Finding P2 — deferred early Stop can be orphaned by navigation or the next send**: accepted, and fixed structurally in commit e3fee61 rather than by hardening the deferred state.

Root cause was that cancellation was keyed on the run id, which the client only learns from `run_started` — so any Stop before that frame had to defer, and the deferred state was fragile (navigation aborts the observer that would deliver the id; the next send resets the shared `pendingStopRef`).

I took your second suggested fix — **a backend cancel-by-conversation path** — which removes the dependency on the run id entirely:

- `run_manager.request_cancel_for_conversation(conv_id)` looks up the conversation's active (queued/running) run via the existing `get_active_run_for_conversation` and cancels it through `request_cancel` (cooperative flag + terminal-guarded DB update). Returns `None` when there's no active run.
- `POST /api/chat/conversations/<id>/cancel-active-run` → `200 {"run": <run>|null}`, `404` only if the conversation is unknown.
- Frontend `stopStreaming` now calls `cancelActiveRun(conversationId)` (via `cancelAndReconcile`, still gated on `conversationRef` for the reconcile). I deleted `runIdRef`, `pendingStopRef`, and both deferred `onRunStarted` handlers.

Now Stop fires the cancel POST the instant it's pressed, whether or not `run_started` has arrived, and it survives a subsequent navigation or send because the server finds the run from the conversation.

Regression coverage: backend `test_cancel_active_run_by_conversation*` (running + queued cancel, no-active no-op, unknown-conversation 404, auth, midstream cooperative stop with no assistant persisted) and the reworked hook tests (`Stop cancels even when no run_started frame has been received yet`, plus the navigate-away reconcile gate). 100 frontend + 299 backend tests pass; lint unchanged.

teo-mateo:

- P2 — A delayed Stop cancel can cancel the next run in the same conversation. `stopStreaming()` fires `cancelAndReconcile(conversation?.id)` but immediately aborts the observer and sets `streaming` false (`dashboard/frontend/src/hooks/useChat.js:377-384`), so the composer can submit another message before the cancel POST has been handled. Because the new endpoint cancels “whatever is active” for the conversation at processing time (`dashboard/chat/routes.py:384-401`, `dashboard/chat/run_manager.py:122-135`), this race can target the wrong run: press Stop near the end of run A, run A completes before the cancel request reaches Flask, the user sends run B while the UI is already re-enabled, then the late `/cancel-active-run` request finds B as the active run and cancels it. The previous run-id-based path was immune because a late cancel stayed keyed to A. Please either keep send/edit disabled in an explicit cancelling state until the cancel POST settles and reconciliation completes, or make the cancel-by-conversation request include a stable expected run token/server-side target so late delivery cannot apply to a subsequent run. A regression test should hold `cancelActiveRun` unresolved after Stop, attempt another send/edit in the same conversation, and assert no second run can start until the cancel request settles.

teo-mateo:

**Finding P2 — a delayed Stop cancel can cancel the next run in the same conversation**: accepted. Fixed in commit ac1103a using your second suggested approach (a server-side target token), which keeps Stop instant and race-proof without an extra disabled window.

`request_cancel_for_conversation(conv_id, expected_run_id=None)` now no-ops when `expected_run_id` is given and the conversation's active run is a *different* run — so a Stop meant for run A that lands after A finished and B started spares B. The route accepts `{"expected_run_id"}` and forwards it. The frontend re-captures the run id from `run_started` into `runIdRef` (capture only — none of the iter-3 deferred state returns) and passes it as the guard; `runIdRef` is reset at the start of every send/edit/load so it can't carry a stale id.

The early-Stop case (no id yet) still sends no guard and cancels whatever's active — which is provably still run A, because A cannot have finished without first emitting `run_started`, which would have populated `runIdRef` and put us on the guarded path.

Regression coverage:
- Backend `test_cancel_active_run_by_conversation_stale_expected_run_id_is_noop`: A completed, B active, cancel with `expected_run_id=A` → `{run: null}`, B stays running. Plus the matching-id happy path.
- Frontend: Stop forwards the captured id (`'conv-1', 'run-123'`); the early-Stop path (`installInFlightStream(null)`) forwards `'conv-1', null`.

100 frontend + 301 backend tests pass; lint unchanged.

Note on the alternative you offered (disable send/edit until the cancel settles): I went with the token guard instead because it's correct regardless of network timing and doesn't add a UX-visible cancelling window — the composer stays responsive.

teo-mateo:

- P2 — Early Stop still has an unguarded stale-cancel race. In the Stop path, `stopStreaming()` passes `runIdRef.current` to `cancelAndReconcile()` and then immediately clears/aborts the stream and flips `streaming` false (`dashboard/frontend/src/hooks/useChat.js:390-397`). When Stop is clicked before the client has parsed `run_started`, `runIdRef.current` is `null`, so `cancelActiveRun()` sends `{}` with no `expected_run_id` (`dashboard/frontend/src/services/chat.js:66-70`). That reopens the iteration-4 race for the no-id case: run A can complete before the unguarded cancel POST reaches Flask, the UI is already enabled so the user can submit run B, and then `/cancel-active-run` cancels whatever is active at processing time (`dashboard/chat/routes.py:407-410`, `dashboard/chat/run_manager.py:141-146`), which is now B. The current reasoning that A “cannot have finished without emitting `run_started`” does not guarantee the browser has handled that frame before Stop; the request may not have resolved/body-read yet, or the fetch can be aborted before the handler runs. Please make the no-id path safe too, e.g. keep send/edit disabled in a cancelling state until the unguarded cancel settles, or make the client learn a stable run token before it can expose Stop and always send an expected target. A regression test should cover Stop before `run_started`, A completing, B starting, and the delayed unguarded cancel not cancelling B.

teo-mateo:

**Finding P2 — early Stop still has an unguarded stale-cancel race**: accepted, you're right. My "A cannot finish without emitting run_started" reasoning conflated *emitted server-side* with *parsed by the browser* — the frame can be in the buffer unread, or the fetch aborted before the handler runs, so `runIdRef` can be null at Stop time even though A is well underway. The unguarded cancel then races B.

Fixed in commit 7068cc9 with your first suggested approach (a cancelling state that blocks send/edit until the cancel settles):

- `useChat` now holds a `cancelling` flag (`cancellingRef` for the synchronous guard) from Stop until **both** the cancel POST and its reconcile settle. `sendMessage`/`editMessage` return early while it's set, so no new run can be created while a cancel — guarded or not — is in flight. An unguarded cancel can therefore only ever be processed against the run the user actually stopped (B literally cannot exist yet).
- The composer (text input, model selector, MCP toggle) is disabled during the window via the exposed `cancelling` prop, so the user can't even attempt the send — no silent drop.
- The `expected_run_id` guard from iter 4 stays as defense-in-depth on the id-known path.

Regression test `blocks a new send until the cancel settles, closing the unguarded race`: early Stop (`installInFlightStream(null)` → no run id), hold `cancelActiveRun` unresolved, attempt a second send (assert `streamChat` call count unchanged → B blocked), then resolve the cancel and confirm the next send is allowed.

101 frontend + 301 backend tests pass; lint unchanged.

teo-mateo:

- P2 — Stop reconciliation can still clobber a conversation after navigation has started but before the new conversation commits. The gate in `cancelAndReconcile()` checks `conversationRef.current?.id === convId` before launching the stopped conversation refetch (`dashboard/frontend/src/hooks/useChat.js:94`), but `loadConversation()` leaves `conversationRef` pointing at the old conversation until the new `getConversation()` resolves (`dashboard/frontend/src/hooks/useChat.js:102-125`). Repro: stop run A, click conversation B, B's fetch is still pending, then A's cancel settles; the gate still sees A and starts `refetchMessages(A)`. If B resolves first and that A refetch resolves last, the hook state is put back on A while the route is B, and `ChatPage` stays in its loading guard because the route id did not change. The current regression only resolves A's cancel after B has already loaded, so it misses this interleaving. Please track the requested/observed conversation id synchronously when `loadConversation(B)` begins (or request-sequence all conversation fetches) and add a test where A's cancel settles during B's in-flight load and A's refetch resolves last.

teo-mateo:

**Finding P2 — Stop reconciliation can clobber a conversation after navigation starts but before it commits**: accepted, good catch. The gate read `conversationRef` (a mirror of *committed* `conversation`), which during `loadConversation(B)` still holds A until B's fetch resolves — so a Stop-A cancel settling inside B's in-flight load passed the gate.

Fixed in commit 81bdf4a by tracking the *intended* conversation synchronously:

- Replaced `conversationRef` with `observedConvIdRef`, set synchronously at the top of `loadConversation` (before awaiting the fetch) and mirrored from committed `conversation` for the `setConversation` paths.
- `refetchMessages` now re-checks `observedConvIdRef` **at resolution** and drops the result if a newer navigation superseded it (request-sequencing). This also closes the pre-existing `loadConversation` A/B interleave that the `pendingFlush` comment in ChatPage already lamented ("loadConversation() has no request sequencing").
- The post-cancel reconcile gate uses `observedConvIdRef` too, so it skips the moment navigation *begins*, not when it commits.

Regression test `a Stop reconcile is dropped when its cancel settles during an in-flight navigation`: Stop A, begin navigation to B with B's `getConversation` held pending, settle A's cancel mid-load, assert `getConversation` is never called with `conv-1` (A's reconcile gated out), then resolve B and assert the active conversation is B.

102 frontend + 301 backend tests pass; lint unchanged.

teo-mateo:

- P2 — Returning to a conversation with an active background run leaves the run uncontrollable and the composer enabled. `getConversation()` already attaches `active_run` to the loaded conversation (`dashboard/chat/db.py:175-178`), but `refetchMessages()` only stores the data and `loadConversation()` explicitly resets `streaming` false (`dashboard/frontend/src/hooks/useChat.js:62-70`, `dashboard/frontend/src/hooks/useChat.js:118-140`). `ChatArea` then hides Stop unless `streaming` is true and enables the composer/model/MCP controls based only on `streaming || cancelling` (`dashboard/frontend/src/components/chat/ChatArea.jsx:196-217`, `dashboard/frontend/src/components/chat/ChatArea.jsx:257-262`). Repro: start a run in A, click B so the observer is aborted while A continues in the background, then click A again before the run finishes. The sidebar can show A's `active_run`, but the chat pane has no Stop button, and a send/edit attempt just hits the backend 409 active-run guard (`dashboard/chat/routes.py:235-237`) after briefly accepting input. More importantly, a long or runaway generation cannot be cancelled from the UI once the user has navigated away and back. Please treat `conversation.active_run` as an active in-pane run: either reattach to `/runs/<id>/stream`, or at minimum disable send/edit/model/MCP controls and show a Stop action that calls the new `cancel-active-run` endpoint. A regression test should load a conversation payload containing `{active_run: {status: 'running'}}` while `streaming` is false and assert the composer is blocked and cancellation remains available.

teo-mateo:

**Finding P2 — returning to a conversation with an active background run leaves it uncontrollable and the composer enabled**: accepted. Fixed in commit a2cc0a3 with the minimum you proposed (treat `conversation.active_run` as an in-pane active run), since `cancel-active-run` and the conversation-id Stop path already exist.

`ChatArea` now computes `hasActiveRun` from `conversation.active_run` (running or queued) and `busy = streaming || cancelling || hasActiveRun`:
- the composer, model selector and MCP toggle are disabled while `busy` — so a send/edit can no longer be briefly accepted only to hit the backend 409;
- the Stop button shows for `streaming || hasActiveRun` (hidden during `cancelling`) and calls `onStopStreaming`, which cancels by conversation id and therefore works even though we hold no SSE stream for that run.

I scoped out **full live reattachment** (subscribing to `/runs/<id>/stream` and replaying the in-progress deltas) — that's a larger feature and the endpoint already exists for a later phase. This change closes the *control* gap (uncontrollable runaway + wrongly-enabled composer) with the infrastructure already in this PR.

Regression tests (`ChatArea — active background run on return`): running and queued `active_run` block the composer and show Stop (and Stop invokes the handler); `null` and terminal statuses leave the composer enabled with no Stop.

106 frontend + 301 backend tests pass; lint unchanged.

teo-mateo:
## Codex review iteration 8 — Findings

- P2 — Reattached active-run Stop sends an unguarded cancel even though the run id is available. The Stop button now appears for `conversation.active_run` (`dashboard/frontend/src/components/chat/ChatArea.jsx:151`, `dashboard/frontend/src/components/chat/ChatArea.jsx:219`), but `stopStreaming()` only passes `runIdRef.current` as the expected-run guard (`dashboard/frontend/src/hooks/useChat.js:426`). After navigating away and back there is no local SSE stream, so `runIdRef` is null while `conversation.active_run.id` still names the run. `cancelActiveRun()` then sends `{}` (`dashboard/frontend/src/services/chat.js:66`), and the backend cancels whatever is active when the request is processed (`dashboard/chat/run_manager.py:141`). That reopens the stale-cancel race for the reattached-run path: if run A finishes before the cancel request is handled and another tab/API starts run B in the same conversation, the Stop intended for A cancels B. Please use `conversation.active_run?.id` as the fallback expected id when `runIdRef.current` is empty, and cover Stop-on-return with a stale active_run id regression test.

- P3 — Active background runs still allow edit attempts. The iteration-7 fix computes `busy` from `active_run`, but only applies it to the model/MCP controls and composer; `MessageList` still receives only `streaming` (`dashboard/frontend/src/components/chat/ChatArea.jsx:246`, `dashboard/frontend/src/components/chat/ChatArea.jsx:253`), `MessageBubble` always exposes Edit for persisted user messages (`dashboard/frontend/src/components/chat/MessageBubble.jsx:177`), and `editMessage()` only guards `streaming || cancellingRef.current` before optimistically truncating local messages (`dashboard/frontend/src/hooks/useChat.js:296`, `dashboard/frontend/src/hooks/useChat.js:311`). Repro: return to a conversation with `active_run: {status: 'running'}` and edit a prior user message; the UI accepts the edit, replaces the transcript with `temp-edit`, then the backend rejects the PUT with the active-run 409. Please pass the busy/disabled state through the message edit path, or add an `active_run` guard in the hook, plus a regression test for editing during a reattached active run.

teo-mateo:
## Author response to codex iteration 8

**Finding P2 — reattached active-run Stop sends an unguarded cancel**: accepted. Fixed in commit d44fe7b. `stopStreaming` now computes `expectedRunId = runIdRef.current || conversation?.active_run?.id || null`, so on the returned-to-run path (no local stream, `runIdRef` null) it falls back to the conversation's `active_run.id`. The cancel therefore always carries an expected-run guard, and a late delivery can't kill a newer run.

**Finding P3 — active background runs still allow edit attempts**: accepted. Fixed in commit d44fe7b on both paths:
- Hook guard: `sendMessage`/`editMessage` now also return early when `isRunActive(conversation.active_run)`, so a reattached run no longer lets an edit optimistically truncate the transcript before the backend 409.
- UI: `MessageBubble` hides Edit when `disableEdit`, and `ChatArea` passes `disableEdit={busy}` through `MessageList` — so the Edit affordance is gone while a run is active, matching the disabled composer.

Tests: `Stop on a returned-to background run uses active_run.id as the cancel guard` (asserts `cancelActiveRun('conv-1','run-bg')`); `editMessage is blocked while the conversation has an active background run` (loads a conversation with `active_run:{running}` + a real user message, calls edit, asserts no stream starts and no `temp-edit` row appears).

108 frontend tests pass; lint unchanged. Backend untouched this round (301 pass).

teo-mateo:
## Codex review iteration 9 — Findings

- P2 — The no-run-id Stop path can still cancel a newer run created outside this hook instance. When Stop fires before `run_started` has been parsed, `stopStreaming()` passes `null` as the expected guard and `cancelActiveRun()` sends `{}` (`dashboard/frontend/src/hooks/useChat.js:444`, `dashboard/frontend/src/services/chat.js:66-70`). The frontend's `cancellingRef` only blocks `sendMessage`/`editMessage` in this one React hook (`dashboard/frontend/src/hooks/useChat.js:103-115`, `dashboard/frontend/src/hooks/useChat.js:154`), but the backend endpoint still resolves an unguarded request by looking up whatever active run exists at handling time and cancelling that run (`dashboard/chat/run_manager.py:141-146`). If run A finishes before this delayed unguarded cancel is handled and another tab/API starts run B in the same conversation, the Stop intended for A cancels B. This is the remaining gap after the iteration-5 local guard: it prevents this tab from creating B, but it does not protect against other clients. The expected-run guard already fixes this for parsed `run_started` and returned-to-`active_run` paths; the remaining no-id path needs a stable target too, for example expose the run id before enabling Stop (response header or first-frame handshake) and always send `expected_run_id`, or make unguarded cancels unable to apply to runs created after the Stop target.

teo-mateo:
## Author response to codex iteration 9

**Finding P2 — the no-run-id Stop path can cancel a newer run created outside this hook instance**: accepted. Fixed in commit 1f0771a with your suggested first-frame handshake.

`observe()` emits `run_started` as its first frame on every send/edit (verified: `_start_run_response` → `manager.observe`), so I added a `runReady` flag that flips true on `run_started` and gate the live-stream Stop button on it: `((streaming && runReady) || hasActiveRun) && !cancelling`. Until `runReady`, the composer is disabled (streaming) and Stop is hidden — so there is no UI path to issue a Stop before the run id is known. Every Stop therefore carries an expected-run guard (the captured `runIdRef`, or `active_run.id` on the returned-to path), never an unguarded `{}` that a concurrently-started run B (another tab/client) could be hit by.

`runReady` is reset at the start of send/edit/load and on Stop, set on `run_started` in both stream handlers, and threaded through ChatPage → ChatArea. The window where Stop is hidden is just send → first-frame (run_started is `observe`'s first yield, before any delta), so it's effectively immediate.

Tests: `runReady stays false until run_started, gating an unguarded Stop` (hook); `ChatArea — Stop gating on runReady` hides Stop while `streaming && !runReady` and shows it once `runReady`.

111 frontend tests pass; lint unchanged. Backend untouched (301 pass).

teo-mateo:
## Codex review iteration 10 — No further findings

I reviewed the current diff against the prior iteration history and did not find additional correctness, security, API-contract, performance, or test-gap issues that are not already addressed or explicitly scoped out.

Verification run locally:
- `npm test -- --run src/hooks/useChat.test.jsx src/components/chat/ChatArea.test.jsx src/components/chat/ChatSidebar.test.jsx src/components/chat/McpToggle.test.jsx` — 37 passed
- `./venv/bin/python -m pytest tests/test_chat_run_routes.py tests/test_chat_runs.py` — 43 passed
- `git diff --check origin/main...HEAD` — clean

PR #67, opened 2026-06-21
Phase 8 (#58): persistence-policy seam to prepare for Ghost Chat (#57)

teo-mateo:
Final phase of #58. Introduces the persistence-policy seam Ghost Chat (#57) needs, **without** implementing Ghost Chat (explicit non-goal of this issue).

## Changes
- **`chat/persistence.py`** — `PersistencePolicy` interface + two implementations:
  - `DbPersistencePolicy` reproduces the runtime's previous inline behavior exactly (claim → complete/fail/cancel on `chat_runs`; message + artifacts written atomically).
  - `NullPersistencePolicy` performs **no** durable writes — keeps the transcript in memory and returns synthetic run/message objects shaped like the DB path's, so the runtime takes identical branches. This is the ephemeral mode Ghost Chat plugs into.
- **`chat/runtime.py`** — `ChatRunner` now delegates every per-turn read/write to a persistence policy, defaulting to `DbPersistencePolicy(db)` so existing callers are unchanged. Pass `persistence=NullPersistencePolicy(...)` to run the same stream ephemerally.

Run-manager orchestration (recovery, the cancel registry, active/list lookups) intentionally stays on the DB — those are durable-coordination concerns, not part of running a single stream; an ephemeral mode wires up its own (or no) manager.

## Acceptance criteria (issue #58)
- ✅ **Clear place for non-persisted chat modes** — `NullPersistencePolicy`.
- ✅ **SSE-encoding code does no DB writes** — already true after Phases 3–5 (`event_codec` / `_sse_frames_for` / `observe` only encode/read); runtime writes now route through the policy rather than calling `db` inline.
- ✅ **Same model/tool stream runs through different persistence policies** — `test_chat_persistence.py` asserts the Db and Null policies produce **identical** runtime event sequences while only the durable one writes to SQLite.

## Tests
`test_chat_persistence.py`: durable policy persists + completes; null policy runs without persisting (nothing in SQLite, transcript kept in memory); identical event stream across policies; null-policy failure persists nothing; default runner uses the DB policy.

Backend-only. **310 backend tests pass**; frontend unchanged (113 pass, build clean, lint unchanged).

### DIFF
--- a/dashboard/chat/persistence.py
+++ b/dashboard/chat/persistence.py
@@ -1,0 +1,115 @@
+"""Persistence policy seam for the chat runtime (Phase 8 of #58).
+
+The runtime (ChatRunner) drives the model/tool stream and publishes events; how
+— or whether — a turn is durably stored is delegated to a PersistencePolicy.
+This is the architectural hook Ghost Chat (#57) needs: the same model/tool
+stream can run DB-backed or fully ephemeral by swapping the policy, with no
+change to the runtime or the SSE-encoding layer.
+
+Two policies ship here:
+
+  - DbPersistencePolicy wraps a ChatDB and reproduces the previous inline
+    behavior exactly (claim → complete/fail/cancel against chat_runs, message +
+    artifacts written atomically).
+  - NullPersistencePolicy performs no durable writes. It keeps the transcript in
+    memory so a multi-turn ephemeral session stays coherent, and returns
+    synthetic run/message objects whose shape matches the DB path so the
+    runtime's control flow is identical either way.
+
+The policy intentionally covers only the per-turn runtime writes. Run-manager
+orchestration (recovery, the cancel registry, list/active lookups) stays on the
+DB directly — those are durable-coordination concerns, not part of running a
+single stream, and an ephemeral mode wires up its own (or no) manager.
+"""
+from .models import ChatRun
+from .runs import ChatRunStatus
+
+
+class PersistencePolicy:
+    """Interface the runtime uses to read prior turns and record this one.
+
+    Implementations must be safe to call from the runner's worker thread.
+    """
+
+    def load_messages(self, conv):
+        """Return the ordered message history used to build the prompt."""
+        raise NotImplementedError
+
+    def claim_running(self, run_id):
+        """Transition the run to RUNNING and return it (with a `.status`), or
+        return a non-running run / None to signal the claim was lost (e.g. the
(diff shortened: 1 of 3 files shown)


teo-mateo:

- `dashboard/chat/persistence.py:100`: `NullPersistencePolicy.claim_running()` always synthesizes a running run and `cancel()` is a no-op, so the non-durable path cannot preserve the existing pre-start cancellation contract. In the DB path, cancelling while a run is still queued marks the row terminal, and `ChatRunner.run()` returns before the model stream is built (covered by `test_cancelled_before_start_does_no_work`). With the null policy, an already-set `cancel_check` is only consulted after `_build_stream()` and after the first stream item is pulled (`dashboard/chat/runtime.py:170`), so an ephemeral run cancelled while queued can still open a model/tool stream and wait for a chunk before emitting `run_cancelled`. That matters for Ghost Chat if it uses the same worker/cancel-flag shape: Stop can still start a costly LLM request. Consider checking `cancel_check()` immediately after the claim and before `_build_stream()`, and add a null-policy cancellation test so the DB and null policies keep the same branch behavior.

teo-mateo:

**Finding — null policy can't preserve the pre-start cancellation contract**: accepted, good catch. Fixed in commit 342dbea.

`ChatRunner.run` now checks `cancel_check()` immediately after `claim_running` and **before** `_build_stream()`, emitting `run_cancelled` (and no `run_started`, matching the DB queued-cancel path) without opening the stream. The DB path kept catching this for free via the terminal-guarded claim; the explicit check brings `NullPersistencePolicy` (no terminal row) onto the same branch, so Ghost Chat's Stop can't kick off an LLM request it immediately abandons. The check is a no-op in the normal case (`cancel_check` false at claim time), so mid-stream cancellation is unchanged.

Regression test `test_null_policy_pre_start_cancel_does_not_open_stream`: with `cancel_check` already true, the stream factory body never runs (stream never iterated), only `run_cancelled` is emitted, and no assistant turn is appended. Verified the existing DB cancel tests (`test_cancelled_before_start_does_no_work`, `test_cancel_stops_run_midstream`) still pass.

311 backend tests pass.

teo-mateo:

Reviewed the current diff after `342dbea` and the prior iteration/author response. I do not have any remaining correctness, security, API-contract, production performance, or test-gap findings to raise.

Verification: `cd dashboard && venv/bin/python -m pytest tests` passed (`311 passed`).

PR #69, opened 2026-06-22
Reattach to live run progress on return to an in-progress conversation

### LINKED ISSUE #68: Reattach to live run progress when returning to an in-progress conversation
## Problem

After the background-run rearchitecture (#58), starting a long task and navigating away correctly keeps the run executing in the background. But if you **return to the conversation while it is still working**, you see only the static sidebar spinner — no live token stream, tool calls, or reasoning. It should behave as if you never left: the in-progress assistant turn should stream live.

## Root cause

The frontend never connects to the existing run-stream endpoint on navigation-back. `loadConversation` refetches the persisted transcript and shows the active-run spinner, but does not open `GET /api/chat/runs/<id>/stream`.

Even if it did, the event bus (`chat/event_bus.py`) is **live-only**: a subscriber receives only events published *after* it subscribes. A reattaching observer would miss everything generated before it returned, so the already-streamed content would be blank until the next delta.

## Proposed fix

**Backend — make reattach replay the in-flight history:**
- `EventBus` keeps a per-active-run replay log (all runtime events since the run started), folded in `publish`, dropped on `STREAM_END`.
- A new atomic `subscribe_with_replay(run_id) -> (queue, history)` captures the history snapshot and registers the live queue under one lock (no gap, no duplication).
- `ChatRunManager.observe(run_id, q, replay=())` emits the replay frames (via the existing `_sse_frames_for`) before entering the live loop. `GET /runs/<id>/stream` uses the replay path; the fresh-send path is unchanged (empty history).

**Frontend — reattach on navigation-back:**
- `loadConversation` opens the run stream when the loaded conversation has a running/queued `active_run`, feeding its SSE frames through the same handlers as a live send (delta / reasoning / tool calls / artifacts / parse warnings), with `streaming` true and the run id known (so Stop stays guarded).
- `sse.js` supports the GET run-stream (no body) and a terminal `run_status` frame (reattach after completion → refetch the persisted reply).

## Acceptance criteria
- Returning to a still-running conversation streams the assistant turn live, including content generated before returning.
- Tool calls, tool results, artifacts, reasoning, and parse warnings render on reattach.
- Reattaching after the run already finished shows the persisted reply (or a failed-run error) with no hang.
- Stop still cancels the run from a reattached view; the sidebar indicator clears.
- The fresh-send path and existing run-stream tests are unaffected.

## Out of scope
- Cross-process / multi-dashboard coordination (the bus is per-process; a process restart still falls back to the DB, unchanged).

teo-mateo:
Closes #68.

Returning to a conversation whose background run is still working showed only the static sidebar spinner — no live tokens. Two causes: the frontend never connected to `GET /api/chat/runs/<id>/stream` on navigation-back, and the event bus is **live-only** (a reattaching observer missed everything generated before it subscribed).

## Backend — replay the in-flight history on reattach
- `EventBus` keeps a per-active-run **replay log** (all events since the run started), appended in `publish`, dropped on `STREAM_END`. Unlike the per-observer queues it is **not** bounded/evicted — losing the oldest deltas would blank the start of the reply — so it lives only for one active run and is freed when the run ends.
- `subscribe_with_replay(run_id) -> (queue, history)` captures the history snapshot **and** registers the live queue under one lock, so every in-flight event reaches a reattaching observer exactly once (in the snapshot xor on the queue — no gap, no duplication).
- `ChatRunManager.observe(run_id, q, replay=())` emits the replay frames (via the existing `_sse_frames_for`) before the live loop. `GET /runs/<id>/stream` uses the replay path; the fresh-send path is unchanged (empty history).

## Frontend — connect on navigation-back
- `loadConversation`, after refetch, opens the run stream when the loaded conversation has a running/queued `active_run`, feeding frames through the same handlers as a live send (delta / reasoning / tool calls / results / artifacts / parse warnings), with `streaming` + `runReady` true and the run id known so **Stop stays guarded**. A terminal `run_status` (the run finished as we reattached) refetches the persisted reply instead of hanging.
- `sse.js`: the GET run-stream carries no body; dispatch the `run_status` frame.

## Tests
- Backend (`test_chat_run_routes.py`): history is replayed before live events and only once; the subscribe snapshot excludes later events (those arrive on the queue); history is dropped on `STREAM_END` so a post-completion reattach falls back to the DB terminal backstop.
- Frontend (`useChat.test.jsx`): loading an active-run conversation opens the GET run-stream and renders deltas into `streamingContent`; no reattach when there's no active run; a terminal `run_status` ends streaming via refetch.

**116 frontend + 314 backend tests pass; build clean; lint unchanged (8 pre-existing errors, 0 new).**

## Out of scope
Cross-process / multi-dashboard coordination — the bus is per-process; a process restart still falls back to the DB, unchanged.

### DIFF
--- a/dashboard/chat/event_bus.py
+++ b/dashboard/chat/event_bus.py
@@ -21,10 +21,24 @@ import threading
 # terminates. Dropped middle deltas don't lose data: the DB has the full reply.
 SUBSCRIBER_QUEUE_MAX = 2048
 
+# Event type that marks a run fully finished. Matches run_manager.STREAM_END;
+# kept as a literal so the bus stays free of a run_manager import. On this event
+# the run's replay history is dropped (the run is over — reattachers fall back
+# to the DB).
+_STREAM_END_TYPE = "stream_end"
+
 
 class EventBus:
     def __init__(self):
         self._subscribers = {}  # run_id -> list[queue.Queue]
+        # run_id -> list[event]: the in-flight history of an active run, kept so
+        # a client that reattaches mid-run (navigated away and back) can replay
+        # everything generated before it subscribed. Appended in publish, dropped
+        # when the run ends (STREAM_END). Unlike the per-observer queues this is
+        # NOT bounded/evicted — losing the oldest deltas would blank the start of
+        # the reply on reattach — so it lives only for one active run at a time
+        # and is freed as soon as that run finishes.
+        self._replay = {}
         self._lock = threading.Lock()
 
     def subscribe(self, run_id: str) -> "queue.Queue":
@@ -33,6 +47,21 @@ class EventBus:
         with self._lock:
             self._subscribers.setdefault(run_id, []).append(q)
         return q
+
+    def subscribe_with_replay(self, run_id: str):
+        """Register an observer AND atomically capture the run's history so far.
+
+        Returns (queue, history). The snapshot and the queue registration happen
+        under one lock, so every in-flight event is delivered to the new observer
+        exactly once: events already published are in `history`; events published
+        after this call arrive on the queue. Used by the reattach endpoint so a
+        returning client replays the full in-progress turn before going live.
+        """
+        q = queue.Queue(maxsize=SUBSCRIBER_QUEUE_MAX)
+        with self._lock:
+            self._subscribers.setdefault(run_id, []).append(q)
+            history = list(self._replay.get(run_id, ()))
+        return q, history
 
     def unsubscribe(self, run_id: str, q: "queue.Queue") -> None:
         with self._lock:
@@ -54,11 +83,23 @@ class EventBus:
         evicted to make room, so a slow observer's memory stays bounded and the
         newest events (incl. the terminal stream_end) are preserved.
 
-        No-op when nobody is listening — the run keeps going regardless; the
-        event is simply not replayed (the DB holds the durable result).
+        Records the event in the run's replay history (so a later reattach can
+        replay it) and, on the terminal stream_end, drops that history. Both the
+        history append and the subscriber snapshot happen under the lock so a
+        concurrent subscribe_with_replay sees a consistent cut.
+
+        No-op for delivery when nobody is listening — the run keeps going
+        regardless; the event is simply not replayed live (the DB holds the
+        durable result), though it is still recorded for an in-flight reattach.
         """
+        is_end = getattr(event, "type", None) == _STREAM_END_TYPE
         with self._lock:
             subs = list(self._subscribers.get(run_id, ()))
+            if is_end:
+                # Run finished: future reattachers fall back to the DB.
+                self._replay.pop(run_id, None)
+            else:
+                self._replay.setdefault(run_id, []).append(event)
         for q in subs:
             while True:
                 try:
--- a/dashboard/chat/event_bus.py
+++ b/dashboard/chat/event_bus.py
@@ -10,8 +10,10 @@ run_id and get a Queue they drain; publishers push events to every current
 subscriber of that run. Thread-safe: the background runner publishes from a
 worker thread while SSE generators drain from request threads.
 """
+import json
 import queue
 import threading
+from collections import namedtuple
 
 # Per-observer queue cap. Generous enough to absorb a normal burst of token
 # deltas, but bounded so a live-but-non-draining client (TCP backpressure, a
@@ -21,23 +23,76 @@ import threading
 # terminates. Dropped middle deltas don't lose data: the DB has the full reply.
 SUBSCRIBER_QUEUE_MAX = 2048
 
-# Event type that marks a run fully finished. Matches run_manager.STREAM_END;
-# kept as a literal so the bus stays free of a run_manager import. On this event
-# the run's replay history is dropped (the run is over — reattachers fall back
-# to the DB).
-_STREAM_END_TYPE = "stream_end"
+# Event types after which a run is finished and its replay history is dropped
+# (future reattachers fall back to the DB). Includes the per-turn terminal
+# runtime events — so a ChatRunner used directly with a bus, which ends on
+# run_completed/run_failed/run_cancelled and never emits the manager-only
+# stream_end, still frees its history — as well as stream_end itself.
+_TERMINAL_TYPES = frozenset({"run_completed", "run_failed", "run_cancelled", "stream_end"})
+
+# A duck-typed stand-in for runtime.ChatRuntimeEvent, used only for the
+# coalesced delta a reattach replay emits. Has .type / .data like the real
+# event, so _sse_frames_for consumes it identically — and the bus stays free of
+# a runtime import.
+_ReplayEvent = namedtuple("_ReplayEvent", ["type", "data"])
+
+
+class _ReplayBuffer:
+    """A run's in-flight history, kept so a reattaching client can replay it.
+
+    Consecutive deltas are folded into a single accumulated segment, so memory
+    scales with the length of the generated text rather than the number of
+    token events (a long reply would otherwise retain one dict + raw-JSON string
+    per token until the run ends). Non-delta events (tool calls/results,
+    artifacts, parse warnings) are kept discrete and in order; a delta segment
+    that follows them starts fresh, preserving the per-round boundaries the
+    frontend relies on (it resets visible content after each tool result).
+    """
+    __slots__ = ("_segments",)
+
+    def __init__(self):
+        # Each segment is either a [content_parts, reasoning_parts] list (a run
+        # of coalesced deltas) or a non-delta event kept as-is.
+        self._segments = []
+
+    def add(self, event):
+        if getattr(event, "type", None) == "delta":
+            data = event.data or {}
+            tail = self._segments[-1] if self._segments else None
+            if isinstance(tail, list):
+                tail[0].append(data.get("content") or "")
+                tail[1].append(data.get("reasoning_content") or "")
+            else:
+                self._segments.append([[data.get("content") or ""],
+                                       [data.get("reasoning_content") or ""]])
+        else:
+            self._segments.append(event)
+
+    def snapshot(self):
+        """Materialize the history into replayable events (deltas coalesced)."""
+        out = []
+        for seg in self._segments:
+            if isinstance(seg, list):
+                content = "".join(seg[0])
+                reasoning = "".join(seg[1])
+                raw = json.dumps({"choices": [{"delta": {
+                    "content": content, "reasoning_content": reasoning}}]})
+                out.append(_ReplayEvent("delta", {
+                    "content": content, "reasoning_content": reasoning, "raw": raw}))
+            else:
+                out.append(seg)
+        return out
 
 
 class EventBus:
     def __init__(self):
         self._subscribers = {}  # run_id -> list[queue.Queue]
-        # run_id -> list[event]: the in-flight history of an active run, kept so
-        # a client that reattaches mid-run (navigated away and back) can replay
-        # everything generated before it subscribed. Appended in publish, dropped
-        # when the run ends (STREAM_END). Unlike the per-observer queues this is
-        # NOT bounded/evicted — losing the oldest deltas would blank the start of
-        # the reply on reattach — so it lives only for one active run at a time
-        # and is freed as soon as that run finishes.
+        # run_id -> _ReplayBuffer: the in-flight history of an active run, kept
+        # so a client that reattaches mid-run (navigated away and back) can
+        # replay everything generated before it subscribed. Folded in publish,
+        # dropped when the run ends (any terminal event). Deltas are coalesced so
+        # memory scales with text length, not token count; the buffer lives only
+        # while its run is in flight and is freed as soon as the run finishes.
         self._replay = {}
         self._lock = threading.Lock()
 
@@ -60,7 +115,8 @@ class EventBus:
         q = queue.Queue(maxsize=SUBSCRIBER_QUEUE_MAX)
         with self._lock:
             self._subscribers.setdefault(run_id, []).append(q)
-            history = list(self._replay.get(run_id, ()))
+            buf = self._replay.get(run_id)
+            history = buf.snapshot() if buf is not None else []
         return q, history
 
     def unsubscribe(self, run_id: str, q: "queue.Queue") -> None:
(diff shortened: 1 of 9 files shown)


teo-mateo:

1. **[P1] Reattached observers never learn that an already-observed run was cancelled.** `_sse_frames_for` intentionally emits no frame for `run_cancelled` or `stream_end` (`dashboard/chat/run_manager.py:35-38`), and `observe` returns immediately when it dequeues `stream_end` (`dashboard/chat/run_manager.py:227-230`). That bypasses the DB backstop, which only runs after a queue timeout. Consequently, if another tab/client cancels a run while this reattached stream is open, `streamChat` resolves normally and the new `finally` block only removes the controller (`dashboard/frontend/src/hooks/useChat.js:228-230`): `streaming` remains true and the loaded `conversation.active_run` remains stale, leaving the composer disabled until a reload/manual Stop. Emit a terminal `run_status` for this path or reconcile state when a reattach stream closes, and add a test where a live reattached run is cancelled externally.

2. **[P1] A live failure leaves `active_run` stale and blocks the next send.** A subscribed failure is encoded as an error frame rather than `run_status` (`dashboard/chat/run_manager.py:59-60`). The reattach `onError` handler clears the streaming visuals but does not refetch the conversation (`dashboard/frontend/src/hooks/useChat.js:219-227`). Since the conversation loaded at attachment time still says the run is `running`, `isRunActive(conversation.active_run)` continues to block send/edit and `ChatArea` continues to treat the view as busy. Reconcile from SQLite on this error path with the existing conversation-id gate, and cover a failure delivered after attachment.

3. **[P2] Reattach transport failures are unhandled and can leave the UI streaming forever.** `streamChat` only catches errors from the reader loop; token/fetch/response-body setup failures occur before that `try` (`dashboard/frontend/src/services/sse.js:12-38`). `reattachRun` launches the promise without awaiting it or adding a `catch` (`dashboard/frontend/src/hooks/useChat.js:159-230`), so a rejected GET produces an unhandled rejection, never invokes `onError`, and leaves `streaming` true. Catch setup failures inside `streamChat` or on this call and route non-abort failures through the same cleanup; add a rejected-fetch test.

teo-mateo:

All three findings accepted and fixed in commit a5f8c56. They're the same class — reattach terminal/error reconciliation — so I unified them with a `terminalSeen` flag + a `reconcile()` closure in `reattachRun`, gated on `observedConvIdRef.current === conv.id` so an aborted (navigated-away) stream never stomps the conversation moved to.

**Finding 1 (P1) — cancelled elsewhere, silent close:** the bus emits no frame for `run_cancelled`/`stream_end`, so the reattach stream just closes. The `.finally` now calls `reconcile()` (drop `streaming` + refetch) when no terminal signal fired, so the now-terminal `active_run` clears and the composer isn't stuck disabled.

**Finding 2 (P1) — live failure leaves active_run stale:** `onError` now also reconciles (refetch), so the failed `active_run` stops blocking the next send/edit; `last_run` surfaces the error after refetch.

**Finding 3 (P2) — setup/transport failure:** `streamChat` rejects before its reader loop on token/fetch/response-body failures. Added a `.catch` that routes non-`AbortError` rejections through `reconcile()` and surfaces the message, instead of leaving `streaming` true on an unhandled rejection.

I went with frontend reconcile-on-close (your "reconcile state when a reattach stream closes" option) rather than a new backend terminal frame, since it covers cancel, silent close, and transport failure uniformly without adding a frame to the fresh-send path.

Tests: `reattach reconciles when the run is cancelled elsewhere (silent stream close)`, `... on a live failure delivered after attach`, `... when the run-stream request fails to open`. 119 frontend tests pass; build clean; lint unchanged. Backend untouched (314 pass).

teo-mateo:

1. **[P2] Replay retention is not actually limited to active runs and can grow without bound.** `EventBus.publish` appends every non-`stream_end` event to `_replay` (`dashboard/chat/event_bus.py:95-102`), but that sentinel is emitted only by `ChatRunManager._execute` (`dashboard/chat/run_manager.py:169-173`). `ChatRunner` is explicitly designed to be callable directly with an `event_bus`; those calls terminate with `run_completed`/`run_failed` and never emit `stream_end`, so their histories remain for the lifetime of the bus. Even through the manager, the store keeps one Python event/dict/raw-JSON payload per delta until completion; there is no output-token cap and up to four workers can stream concurrently, so a long or degenerate tool run can retain hundreds of thousands of objects per run and exhaust the dashboard process before the eventual cleanup. Make cleanup part of the runner/bus terminal contract (or clear on terminal runtime events), and coalesce replay deltas into a compact snapshot so memory scales with retained text rather than token-event count. Add coverage using a real direct `ChatRunner` terminal path; the current tests only prove cleanup after manually publishing the manager-only sentinel.

teo-mateo:

**Finding (P2) — replay retention not bounded to active runs / unbounded growth**: accepted, both parts fixed in commit a37d0d5.

**Lifecycle (the leak):** cleanup is now on the terminal-event contract, not the manager-only sentinel. `EventBus.publish` drops a run's replay on any of `run_completed` / `run_failed` / `run_cancelled` / `stream_end` (`_TERMINAL_TYPES`). So a `ChatRunner` driven directly with a bus — which ends on `run_completed`/`run_failed` and never emits `stream_end` — frees its history immediately instead of leaking for the life of the process.

**Memory (token-count growth):** a `_ReplayBuffer` folds consecutive deltas into a single accumulated segment (content/reasoning joined as strings, materialized only at snapshot time), so retention scales with generated text length rather than one dict + raw-JSON per token. Non-delta events (tool calls/results, artifacts, parse warnings) stay discrete and in order, and a delta run *after* a tool event starts a fresh segment — preserving the per-round boundaries the frontend relies on (it resets visible content after each tool result). The snapshot emits one coalesced delta per run via a duck-typed `_ReplayEvent`, so the bus stays free of a runtime import and `_sse_frames_for` consumes it unchanged.

Tests added:
- `test_replay_coalesces_deltas_but_keeps_tool_boundaries` — two deltas → one frame, but a `tool_call` between two runs yields 3 frames.
- `test_terminal_runtime_event_drops_replay_without_stream_end` — `run_completed` frees replay with no `stream_end`.
- `test_direct_runner_completion_frees_replay_history` / `..._failure_...` — a **real direct `ChatRunner`** terminal path frees the history (your requested coverage).

318 backend tests pass; frontend unchanged (119); lint unchanged.

teo-mateo:

1. **[P1] The new callback dependency makes conversation loading loop on every `ChatPage` render.** `reattachRun` now depends on `onConversationUpdated`, and `loadConversation` depends on `reattachRun` (`dashboard/frontend/src/hooks/useChat.js:263,301`). `ChatPage` passes `onConversationUpdated` as a new inline function on every render (`dashboard/frontend/src/components/chat/ChatPage.jsx:38-40`) while its URL-loading effect depends on `loadConversation` (`dashboard/frontend/src/components/chat/ChatPage.jsx:49-53`). Therefore every render changes `loadConversation`, reruns the effect, and calls `loadConversation` again. That call sets several fresh array states (`setToolEvents([])`, etc.), guaranteeing another render even when the conversation id did not change, so opening a conversation enters a fetch/effect loop and repeatedly aborts any reattach stream. Stabilize the callback in `ChatPage` or keep the update callback behind a ref so `reattachRun`/`loadConversation` remain stable, and add a component-level test that rerendering `ChatPage` for one route performs only one conversation load; the hook-only tests do not exercise this dependency chain.

teo-mateo:

**Finding — Unstable `loadConversation` causes ChatPage reload loop (P1)**: accepted. Fixed in commit 98b28b4.

`ChatPage` builds a fresh `onConversationUpdated` (`(evt) => patchConversation(...)`) on every render. The reattach work made `loadConversation` depend transitively on that callback (via `reattachRun`), so its identity changed each render — and ChatPage's URL-load effect `useEffect(..., [convId, loadConversation])` re-fired on every render, reloading the conversation in a loop.

Fix: `useChat` now keeps the caller's callback behind a ref (`onConversationUpdatedRef`, kept current via an effect) and exposes a single stable wrapper `handleConversationUpdated` (`useCallback(..., [])`). `reattachRun`, `loadConversation`, `sendMessage`, and `editMessage` all depend on the stable wrapper instead of the raw prop, so `loadConversation` keeps a stable identity regardless of how often the parent re-renders.

Tests added (both requested):
- **Hook-level** (`useChat.test.jsx`): `loadConversation` and `sendMessage` keep the same identity when `onConversationUpdated` identity changes across rerenders — covering the root cause directly.
- **Component-level** (`ChatPage.test.jsx`, new): rendering `ChatPage` for one route calls `getConversation` exactly once across three forced parent rerenders. I verified this test has teeth — reverting the ref fix makes it hang in the old reload loop (getConversation hammered until the worker times out).

Also removed now-dead `fullContent`/`fullReasoning` accumulation in `editMessage` — `onMessageSaved` refetches from the DB now instead of building the saved message from accumulated deltas, so those locals were unused (two no-unused-vars lint errors cleared). The UI-facing `setStreamingContent`/`setStreamingReasoning` updates are unchanged.

Full FE suite (121) + backend chat suites (33) green; `npm run build` clean; no new lint errors.

teo-mateo:

1. **[P1] StrictMode can open two reattach streams and duplicate the replay/live output.** The application is mounted under React StrictMode (dashboard/frontend/src/main.jsx:9), which deliberately runs the URL-load effect twice on its initial development mount (dashboard/frontend/src/components/chat/ChatPage.jsx:49-53). Both asynchronous loadConversation calls can pass the only completion guard because it compares just the conversation id (dashboard/frontend/src/hooks/useChat.js:306-312). The abort step ran before either fetch resolved, so each completion calls reattachRun, creates its own controller, and opens its own GET stream (dashboard/frontend/src/hooks/useChat.js:164-189); the second does not cancel or supersede the first. Both streams replay the same history and both handlers append every subsequent delta/tool event, so returning to an active conversation in the documented Vite dev workflow renders duplicated output. The same race is possible in production with A → B → A navigation while the first A fetch is pending. Track a monotonically increasing load generation (not only the id) and allow only the latest load to attach, or explicitly deduplicate/supersede subscriptions by run id. Add coverage using StrictMode or two deferred same-id loads that asserts only one run stream is opened and each delta is applied once; the new rerender test does not exercise StrictMode.

teo-mateo:

**Finding 1 — StrictMode can open two reattach streams and duplicate output (P1)**: accepted. Fixed in commit 7253d3c.

Root cause confirmed exactly as described: `ChatPage`'s URL-load effect runs twice under StrictMode (and an A→B→A navigation re-loads A while A's first fetch is pending). Both same-id loads passed the only guard (`observedConvIdRef.current === convId`), so each resolved `refetchMessages`, called `reattachRun`, created its own controller, and opened its own `GET /chat/runs/<id>/stream`. Neither superseded the other, so both replayed the buffered history and both applied every subsequent delta/tool event.

Fix: a monotonic **load generation**. `loadConversation` now captures `const gen = ++loadGenRef.current` synchronously at entry (before the await), and the post-fetch reattach fires only when `loadGenRef.current === gen` **and** the id guard still holds. Two same-id loads get distinct generations; only the latest attaches, the earlier short-circuits. This covers both the StrictMode double-mount and the production A→B→A race (the id-only guard could not disambiguate two loads of the same id).

Why a generation rather than run-id dedup: the duplicate can occur before either fetch resolves (no run id known yet at `loadConversation` entry), so superseding has to key on the load itself, not the run.

Test added (`useChat.test.jsx`): "opens only one run stream when the same conversation is loaded twice concurrently" — two deferred same-id `loadConversation` calls, both fetches resolved to an active run, asserts exactly one `/chat/runs/run-bg/stream` is opened. Verified to have teeth: reverting the generation guard makes it open two streams and the test fails.

Full FE suite (122) green; no new lint errors.

teo-mateo:

1. **[P1] The load-generation guard runs after a superseded response has already committed stale conversation state.** `loadConversation` calls `refetchMessages`, which unconditionally runs `setConversation(data)` and the other state setters whenever the conversation id still matches (`dashboard/frontend/src/hooks/useChat.js:92-109`); the generation is checked only afterward, around whether to call `reattachRun` (`dashboard/frontend/src/hooks/useChat.js:318-326`). With the same-ID concurrent loads this guard was added for, an older request can capture `{ active_run: running }`, a newer request can return `{ active_run: null }` after the run completes, and the responses can arrive newest-first/oldest-last. The old response then restores the stale active run but is correctly prevented from attaching, leaving `streaming === false` while `conversation.active_run` still blocks send/edit and keeps Stop visible until another reload. Make the generation check gate the state/error writes for load-triggered fetches as well (for example, pass a freshness predicate/generation into the fetch-and-commit path or separate fetching from committing). Add an out-of-order test with different snapshots; the current concurrent-load test resolves both requests with the same object and only checks the stream count, so it cannot catch this stale-state outcome.

teo-mateo:

**Finding 1 — Superseded same-id load commits stale conversation state (P1)**: accepted. Fixed in commit bcea528.

Confirmed: the generation guarded reattach but not the `setConversation` commit in `refetchMessages`, which gated only on the conversation id. With concurrent same-id loads resolving oldest-last, the older `{active_run: running}` snapshot overwrote the newer `{active_run: null}`, leaving a stale active run that blocks send/edit and keeps Stop visible — while the older load was (correctly) barred from reattaching, so `streaming` stayed false. Stuck until another reload, exactly as described.

Fix: `refetchMessages(convId, isFresh?)` now takes an optional freshness predicate, re-checked **after** the fetch resolves, combined with the existing id guard (`stillCurrent = observedConvIdRef.current === convId && (isFresh ? isFresh() : true)`). A stale result returns null without committing any state. `loadConversation` passes `() => loadGenRef.current === gen`, so only the latest same-id load commits its snapshot (and, since it returns null for the stale one, the reattach below short-circuits too). Other callers (post-save drain, cancel/reconcile) pass no predicate and keep the prior id-only behavior unchanged.

Test added (`useChat.test.jsx`): "does not commit a superseded same-id load that resolves out of order" — two same-id loads with different snapshots, newest resolved first and oldest last, asserts `conversation.active_run === null` (latest wins) and `streaming === false`. Verified to have teeth: removing the predicate makes the stale running run commit and the test fails.

Full FE suite (123) green; no new lint errors.

teo-mateo:
## Codex review iteration 6 — No further findings

Reviewed the current head after reading all prior iteration comments and author responses. I found no additional actionable issues in the replay buffer, atomic replay subscription, run-stream terminal handling, or frontend reattach/request-sequencing paths.

Verification: 318 backend tests passed; 123 frontend tests passed; the frontend production build completed successfully.

PR #71, opened 2026-06-22
Guarantee tool-loop forced final response terminates with done (#70)

### LINKED ISSUE #70: Tool-loop turn hard-fails as "Stream ended unexpectedly" when the forced final response emits a tool call
## Symptom

A tool-heavy chat turn (e.g. web research) that exhausts the `MAX_TOOL_ROUNDS` (5) cap can hard-fail after the full multi-round wait with `error = "Stream ended unexpectedly"` and no assistant message saved. Observed on a real conversation (model `ds4-deepseek-v4-flash`): user asked "research online if there's any plagiarism in this story", the model called the websearch tool 5× , the loop forced a final response, and the run failed ~4.5 min in.

Evidence:
- `chat_runs`: `status=failed`, `active_step=generating`, `error="Stream ended unexpectedly"`, ran ~12:10:40Z → 12:15:05Z.
- log mid-run: `chat.tool_loop - WARNING - Exceeded 5 tool call rounds, forcing final response`.

## Root cause

`stream_chat_completion` (`dashboard/chat/llm_proxy.py:207-208`) always classifies the end of a stream as **either** `done` **or** `tool_calls` — it emits `("tool_calls", …)` whenever `finish_reason == "tool_calls"` or it accumulated tool-call deltas with no content.

The forced-final-response path in `stream_with_tools` (`dashboard/chat/tool_loop.py:122-134`) does the final call with `tools=None` but only handles `delta / tool_call_pending / parse_warning / done / error` — it has **no `tool_calls` branch** (the main round loop at `:37` does). A model that just looped on a tool is very likely to emit one more tool call even on the forced pass, since its recent context is saturated with tool usage.

So that final `tool_calls` event is silently dropped, the loop ends, `stream_with_tools` returns **without yielding `done`**, and `runtime.py:259` fails the run as `Stream ended unexpectedly`. (A dropped connection/timeout would instead surface a handled `error` event with a different message — confirming this is the code gap, not a backend crash.)

## Net effect

A tool-heavy request that exhausts the round cap is effectively guaranteed to hard-fail if the model attempts one more tool call on the forced final pass — after the user has already waited the full multi-round duration, with the work discarded.

## Fix direction

Make the forced final call unable to fall through without a terminal `done`:
- Send the final call with `tool_choice="none"` so the backend cannot emit a tool call (cleanest guarantee), and/or
- Handle a `tool_calls` (or any non-`done`) outcome in the forced-final block by synthesizing a `done` from whatever content was produced, so the turn always terminates cleanly.

Add regression coverage: a tool-loop that exceeds the cap and whose forced final stream emits a tool call must still produce a `done`/`run_completed`, not `Stream ended unexpectedly`.

teo-mateo:
Fixes #70.

## Problem

A tool-heavy turn (e.g. web research) that exhausts `MAX_TOOL_ROUNDS` (5) makes one forced final call. That call used `tools=None` but the forced-final block in `stream_with_tools` only handled `delta / tool_call_pending / parse_warning / done / error` — it had **no `tool_calls` branch** (unlike the main round loop).

`stream_chat_completion` always classifies a stream's end as **either** `done` **or** `tool_calls` (`llm_proxy.py:207-208`). A model that just looped on a tool 5× often emits one more tool call, so the forced final stream ended with a `tool_calls` event — which the forced block silently dropped. `stream_with_tools` then fell off the end with **no `done`**, and `runtime.py:259` failed the run as `Stream ended unexpectedly` — after the user waited out every tool round, with no answer saved.

Observed live: model `ds4-deepseek-v4-flash`, prompt "research online if there's any plagiarism in this story" → 5 websearch rounds → forced final → failed ~4.5 min in.

## Fix

1. **Prevent** — the forced final call now passes `tool_choice="none"` (with the tools still in context) so a compliant backend produces prose instead of another tool call. `stream_chat_completion` gains a `tool_choice` override; the default `"auto"` path when tools are present is unchanged.
2. **Guarantee** — the forced-final block accumulates the streamed prose and, if the inner stream ends without a terminal `done`/`error` (it emitted `tool_calls`, or exhausted silently), **synthesizes a `done`** from that prose. The turn now always terminates cleanly, regardless of whether a backend honors `tool_choice="none"`.

## Tests

- `test_tool_loop.py` (new): forced-final tool-call and silent-exhaustion both still yield `done` (both verified to fail without the safety net); a normal forced-final `done` is forwarded without duplication; a pre-cap answer never gets `tool_choice="none"`.
- `test_llm_proxy_stream.py`: `tool_choice` override with tools, pinned without tools, and the unchanged `"auto"` default.

Full backend suite: 325 passed.

### DIFF
--- a/dashboard/chat/llm_proxy.py
+++ b/dashboard/chat/llm_proxy.py
@@ -95,7 +95,8 @@ def build_messages_array(system_prompt: str, messages: list) -> list:
     return arr
 
 
-def stream_chat_completion(service_name: str, messages_array: list, tools: list = None):
+def stream_chat_completion(service_name: str, messages_array: list, tools: list = None,
+                           tool_choice: str = None):
     """Stream a chat completion from a llama.cpp service.
 
     Yields (event_type, data) tuples:
@@ -103,6 +104,10 @@ def stream_chat_completion(service_name: str, messages_array: list, tools: list 
       - ("done", {"content": full_content, "reasoning_content": full_reasoning})
       - ("tool_calls", {"tool_calls": [{"id": ..., "function": {"name": ..., "arguments": ...}}]})
       - ("error", {"message": ...})
+
+    `tool_choice` overrides the default. When `tools` are supplied it defaults to
+    "auto"; pass "none" to forbid tool calls (e.g. the tool-loop's forced final
+    response, which must produce prose, not yet another tool call).
     """
     svc = resolve_service(service_name)
     if svc is None:
@@ -120,13 +125,17 @@ def stream_chat_completion(service_name: str, messages_array: list, tools: list 
     }
     if tools:
         payload["tools"] = tools
-        payload["tool_choice"] = "auto"
+        payload["tool_choice"] = tool_choice or "auto"
         # Tool-using turns are sensitive to sampling variance: at vLLM's
         # default temperature=1.0, the model occasionally drifts from the
         # structured tool-call format into XML-in-content (`<arg_key>...`
         # `</function>` ghost calls, etc.). 0.3 keeps the format on-rails
         # without making the answer prose feel robotic.
         payload["temperature"] = 0.3
+    elif tool_choice is not None:
+        # No tool schemas in the request, but still pin tool_choice (e.g.
+        # "none" on the forced final response) for backends that honor it.
+        payload["tool_choice"] = tool_choice
 
     collected_content = ""
     collected_reasoning = ""
--- a/dashboard/chat/tool_loop.py
+++ b/dashboard/chat/tool_loop.py
@@ -144,11 +144,26 @@ def stream_with_tools(service_name: str, messages_array: list, tools: list, mcp_
         # the round cap. It falls through to the synthesized done below.
 
     # The forced-final stream ended without a terminal done/error — e.g. it
-    # emitted a tool_calls event (unhandled above) or exhausted silently.
-    # Synthesize a done from whatever prose streamed so the turn ALWAYS
-    # terminates cleanly, instead of the caller failing it as "Stream ended
-    # unexpectedly" after the user already waited out every tool round.
-    yield ("done", {
-        "content": "".join(final_content),
-        "reasoning_content": "".join(final_reasoning) or None,
-    })
+    # emitted a tool_calls event (unhandled above) or exhausted silently. The
+    # turn must still terminate cleanly instead of the caller failing it as
+    # "Stream ended unexpectedly" after the user waited out every tool round.
+    content = "".join(final_content)
+    if content.strip():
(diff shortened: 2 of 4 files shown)


teo-mateo:

1. **[P1] Do not complete a tool-only forced final with an empty assistant response** (`dashboard/chat/tool_loop.py:151`)

   The fallback case this code is specifically intended to handle can emit only `tool_call_pending`/`tool_calls`, with no content delta. In that case `final_content` remains empty and this synthesized `done` carries `content: ""`. `ChatRunner` treats every `done` as success, persists that empty assistant message, and marks the run completed. The user therefore still gets no answer after all tool rounds, but now the failure is silently reported as success. The regression test avoids this case by injecting `"Partial answer."` before the final tool call. Please add coverage for a forced stream containing only `_tool_calls_event()` and make that path produce a non-empty user-visible fallback (or another truthful terminal outcome) rather than an empty successful completion.

teo-mateo:

**Finding 1 — Empty synthesized done reported as silent success (P1)**: accepted. Fixed in commit 35c6c97.

You're right: when the forced final stream emits only a `tool_call`/`tool_call_pending` and no content delta, `final_content` stayed empty, and `ChatRunner` treats every `done` as success — so it would persist an empty assistant message and mark the run completed, converting a no-answer turn into a silent success.

Fix: the forced-final fallback now splits on whether any prose actually streamed:
- **Non-empty** (`content.strip()`): finalize as `done` — a real partial answer (model wrote prose then tried one more tool call) is preserved.
- **Empty / whitespace-only**: yield an `error` with a truthful, actionable message ("The model kept calling tools and did not return a final answer after N tool rounds. Try rephrasing the request or turning off some tools."). `runtime` fails the run with that message — a truthful terminal outcome rather than an empty successful completion.

Either way the generator still always terminates with a `done` or `error`, so the original "Stream ended unexpectedly" fall-through can't recur.

Tests added (`test_tool_loop.py`):
- `test_forced_final_tool_only_no_content_yields_error` — forced stream of only `_tool_calls_event()` (the exact case you flagged) → terminates as `error`, asserts no `done` is emitted.
- `test_forced_final_whitespace_only_content_yields_error` — whitespace-only prose treated as empty → `error`.
- Both verified to fail if the empty case is completed as `done`. The partial-prose and silent-exhaustion-with-prose cases still complete as `done`.

Full backend suite: 327 passed.

teo-mateo:

Reviewed the current head (`35c6c97`) against `main`, including the full changed files and all prior PR comments. The iteration-1 empty-success path is correctly resolved with a truthful terminal error, and I found no additional actionable correctness, security, API-contract, production-performance, or test-coverage issues.

Verification: `dashboard/venv/bin/pytest -q tests/` — 327 passed.

PR #72, opened 2026-07-11
fix(chat): make direct clicks on sidebar checkboxes toggle selection

teo-mateo:
## Problem

Selecting a conversation via its checkbox showed **"1 selected"** in the toolbar, but the checkbox itself stayed visually unchecked.

## Root cause

The `<label>` wrapping each checkbox calls `e.preventDefault()` on click (needed since PR #48 to suppress the label's synthetic forwarded click for padding clicks). But when the click lands **directly on the input**, canceling the event makes the browser revert the native `checked` toggle *after* React has already committed `checked={true}` — and since React only rewrites the DOM property when the prop changes, the box stays permanently desynced from selection state.

## Fix

Direct input clicks are now handled on the input itself: `stopPropagation()` keeps the label's `preventDefault` out of the path (so nothing reverts the native toggle), and the handler calls `onToggleSelect(conv.id, e.shiftKey)` — shift-click range selection keeps working since the modifier is read from the real mouse event. Padding clicks go through the unchanged label path.

## Testing

- `vitest run src/components/chat/ChatSidebar.test.jsx` — 9/9 pass

### DIFF
--- a/dashboard/frontend/src/components/chat/ChatSidebar.jsx
+++ b/dashboard/frontend/src/components/chat/ChatSidebar.jsx
@@ -30,7 +30,12 @@ function ConversationItem({ conv, activeId, depth, selectMode, selected, onToggl
           mouse event. Reading shiftKey from the label click — and
           preventDefault-ing so the synthetic input toggle never fires —
           keeps the full hit target consistent with the visible checkbox
-          (PR #48 codex iter 2). */}
+          (PR #48 codex iter 2). Direct clicks on the input must NOT be
+          canceled — a canceled click makes the browser revert the native
+          checked toggle after React has already committed checked=true,
+          desyncing the visible box from selection state — so the input
+          handles its own click (stopPropagation keeps the label handler
+          out of it). */}
       {isSpinoff ? (
         <label
           className="group/iconslot relative w-8 h-8 -m-2 flex items-center justify-center flex-shrink-0 cursor-pointer"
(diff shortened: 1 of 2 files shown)


teo-mateo:

- **[P2] Add regression coverage for direct input clicks.** Every selection test in `ChatSidebar.test.jsx` calls `checkboxFor()`, which returns the wrapping `<label>`, so none of the 9 tests exercise either new input `onClick` handler. The existing note that this cannot be reproduced under JSDOM is no longer accurate for this event path: with the current React/JSDOM versions, `fireEvent.click(input)` leaves the old implementation visually unchecked because the wrapping label cancels the click, while it remains checked with this patch. Please add at least one test that retrieves the checkbox input directly, clicks it, and asserts the checked/selection state; covering both a root and spinoff would protect both duplicated handlers. As written, removing or breaking the actual fix still leaves the reported 9/9 tests green.

teo-mateo:

**Finding 1.1 — Add regression coverage for direct input clicks**: accepted.

Fixed in commit 070ac4d. Added four tests that retrieve the checkbox `<input>` directly (via `getByRole('checkbox')` within the row) and click it: root-row select, spinoff-row select (covering both duplicated handlers), shift-click range extension from a direct-input click, and toggle-off. Each asserts both `input.checked` and the selection list. Also replaced the stale "cannot be reproduced under jsdom" note. Verified the tests fail (3 of 4) when run against the pre-fix component from main. Full suite: 13/13 pass.

teo-mateo:

I re-reviewed the current head after the iteration 1 test-coverage fix. The direct-input and label-padding paths are now both exercised, and I found no additional correctness, security, API-contract, or production-performance issues.

Validation: targeted ESLint passed, all 127 frontend tests passed, the production build passed, and git diff --check passed.

PR #74, opened 2026-07-14
feat(chat): support chatting with OpenRouter models

teo-mateo:
## Summary

Chat (main + critic) can now target OpenRouter-hosted models from a curated, editable selection of popular models, alongside the local Docker vLLM/llama.cpp services.

- **Convention**: an OpenRouter model rides through the existing `main_service` / `sidekick_service` / `model_service` strings as `openrouter:<model-id>` — no DB migration.
- **Backend**: `chat/openrouter.py` (prefix helpers, built-in `DEFAULT_MODELS`, resolver); `llm_proxy.resolve_service()` branches on the prefix; a shared `build_endpoint()` generalizes URL/header building for `stream_chat_completion` and `critique.py`. OpenRouter payloads get the required `model` field; local single-model servers keep model-less payloads (regression-tested).
- **Curated list**: editable at runtime via `GET/PUT/DELETE /api/chat/settings/openrouter-models` (settings_store pattern, stored in `chat_settings.json`) and a new Tools-page card. The list is a picker convenience, not an allowlist — removing a model doesn't break conversations already using it.
- **Gating**: everything keys off `OPENROUTER_API_KEY` in `dashboard/.env`. Unset → no OpenRouter group in the pickers, and `openrouter:` conversations fail fast with a specific error.
- **Streaming hardening**: mid-stream `{"error": ...}` chunks (e.g. rate limits) now surface as error events instead of being silently swallowed; non-200 bodies get their provider `error.message` extracted.
- **Frontend**: Local/OpenRouter optgroups in both dropdowns with a stale-selection fallback option; new-conversation default falls back to the first curated model when no local service runs; message bubbles show `model-id · OpenRouter`.

MCP tool calling, images, auto-titling, and spinoffs work unchanged through the same code paths.

## Testing

- Backend: 377 passed (`pytest dashboard/tests`), including new `test_openrouter_settings.py` / `test_openrouter_routes.py` and provider-branch coverage.
- Frontend: 140 passed (vitest), including new ModelSelector / ChatPage-default-model / helper tests; `npm run build` clean.
- Live end-to-end (real key, streamed reply, tools, critic) pending a dashboard restart — a model is currently running, so the service was deliberately not restarted.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -234,6 +234,30 @@ accuracy / tool-posture guidance. See
 `dashboard/mcp_servers.example.json` for the `websearch` `tool_hint`
 carrying the web-research guidance.
 
+## Chatting with OpenRouter models
+
+Chat can target OpenRouter-hosted models alongside local Docker services.
+Enable by setting `OPENROUTER_API_KEY` in `dashboard/.env` (restart the
+dashboard to pick it up); without it, OpenRouter models simply don't
+appear in the chat pickers.
+
+- An OpenRouter model is addressed by the service string
+  `openrouter:<model-id>` (e.g. `openrouter:anthropic/claude-sonnet-5`) in
+  the same `main_service` / `sidekick_service` / `model_service` columns as
+  local service names. `chat/llm_proxy.resolve_service()` branches on the
+  prefix via `chat/openrouter.py`.
+- The curated picker list (built-in defaults in
+  `chat/openrouter.py:DEFAULT_MODELS`) is editable at runtime: Tools page →
+  "OpenRouter models" card, or `GET/PUT/DELETE
+  /api/chat/settings/openrouter-models` (stored under `openrouter_models`
+  in `chat_settings.json`). The list is a picker convenience, **not an
+  allowlist** — any `openrouter:` string resolves as long as the key is
+  set, so removing a model doesn't break conversations already using it.
+- MCP tool calling, streaming, images, and critique all go through the
+  same code paths as local models. OpenRouter requires an explicit
+  `model` field in the payload; local single-model servers must NOT get
+  one (regression-tested in `tests/test_openrouter_routes.py`).
+
 ## Reference: working embedding service
 
 `vllm-nomic-embed-text-v1.5` (port 3320) — see `services.json`. Smoke
--- a/dashboard/.env.example
+++ b/dashboard/.env.example
@@ -5,3 +5,5 @@ COMPOSE_PROJECT_NAME=llm-dock
 COMPOSE_FILE=../docker-compose.yml
 LOG_LEVEL=INFO
 LLM_DOCK_API_KEY=your-global-api-key-here
+# Optional: enables OpenRouter-hosted models in the chat model pickers.
+# OPENROUTER_API_KEY=sk-or-v1-...
--- a/dashboard/chat/critique.py
+++ b/dashboard/chat/critique.py
@@ -53,8 +53,10 @@ def request_critique(sidekick_service: str, context: str, extra_instructions: st
             return {"error": f"Sidekick returned HTTP {resp.status_code}: {resp.text}"}
 
         data = resp.json()
-        raw_content = data["choices"][0]["message"].get("content", "")
         message = data["choices"][0]["message"]
+        # OpenAI-compatible providers define content as string|null — an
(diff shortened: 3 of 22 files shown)


teo-mateo:

- **[P1] Rebase and preserve the new conversation-rename path while resolving the conflict.** GitHub currently reports this PR as `CONFLICTING`, and a merge-tree check locates the conflict in `dashboard/frontend/src/components/chat/ChatPage.jsx`. Current `main` added the `rename` hook result, `handleRename`, and the corresponding sidebar callback in the same region where this PR adds the OpenRouter default-model logic. The PR cannot merge as-is; when resolving it, combine both paths rather than choosing the PR side wholesale, which would regress conversation renaming.

teo-mateo:

**Finding 1.1 — PR conflicts with main (ChatPage.jsx, rename path)**: accepted.

Fixed in merge commit 8e98033: merged origin/main into the branch and resolved the `ChatPage.jsx` conflict by keeping both sides — `handleRename` (+ the `rename` hook result and the sidebar `onRename` prop from PR #73) and the OpenRouter default-model fallback. Frontend suite (140 tests), production build, and eslint all pass after the merge.

teo-mateo:

- **[P1] Ignore the final OpenRouter usage chunk instead of indexing an empty `choices` list.** OpenRouter documents that streaming usage is emitted in a final chunk before `[DONE]` with `choices: []` (https://openrouter.ai/docs/api/reference/overview#responses). At `dashboard/chat/llm_proxy.py:225`, `chunk.get("choices", [{}])[0]` raises `IndexError` for that normal chunk. I reproduced the provider sequence locally: the generator emits all answer deltas and then emits `Unexpected error: list index out of range`; `ChatRunner` consequently marks the run failed and never persists the assistant message. The new request-shape test ends directly with `[DONE]`, so it misses the actual provider contract. Please skip/process chunks whose `choices` array is empty and add a regression fixture containing the final usage chunk, asserting that the stream still reaches `done`.

teo-mateo:

**Finding 2.1 — IndexError on OpenRouter's final usage chunk (`choices: []`)**: accepted.

Fixed in commit 3013c13. `chunk.get("choices", [{}])[0]` only defaulted when the key was *missing*; an empty list still indexed and raised. Now `(chunk.get("choices") or [{}])[0]` treats any choice-less chunk as a no-op, so the stream proceeds to `[DONE]` and yields `done` normally. Added a regression test with the documented provider sequence (`: OPENROUTER PROCESSING` keep-alive, content delta, `choices: []` usage chunk, `[DONE]`) asserting the events are exactly `delta, done` with the full content. 80 focused backend tests pass.

teo-mateo:

- **[P1] Normalize nullable OpenRouter content before choosing the reasoning fallback.** OpenRouter defines non-streaming `message.content` as `string | null`, but `dashboard/chat/critique.py:56-61` uses `.get("content", "")` and then immediately calls `raw_content.strip()`; the default does not apply when the field is explicitly null. I reproduced a 200 response with `{"content": null, "reasoning": "<valid critique JSON>"}`, and `request_critique()` returned `'NoneType' object has no attribute 'strip'` instead of parsing the reasoning that this PR explicitly added support for. Assign `message` first and normalize with `raw_content = message.get("content") or ""` (or otherwise type-check it), and add a regression case with null content plus valid reasoning JSON.

teo-mateo:

**Finding 3.1 — `.strip()` crash on explicit `content: null` in the critique path**: accepted.

Fixed in commit 088e955. `message.get("content", "")` only defaults when the key is missing; an explicit `null` reached `.strip()` and turned a valid reasoning-only response into an error. Now normalized with `message.get("content") or ""` so the reasoning fallback applies. Added a regression test with `{"content": null, "reasoning": "<critique JSON>"}` asserting the critique parses (`verdict: ok`, no error). 57 focused backend tests pass.

teo-mateo:

Reviewed the full diff and all prior discussion, including the accepted fixes from iterations 1–3. I found no additional correctness, security, API-contract, production-performance, or material test-coverage issues.

Validation performed: 68 focused backend tests passed; 13 focused frontend tests passed; the production frontend build succeeded; the PR is currently mergeable; and all 12 built-in model IDs are present in OpenRouter’s current catalog with tool support.

PR #77, opened 2026-07-15
feat(chat): projects — group conversations into projects (phase 1)

teo-mateo:
Phase 1 of the projects feature: lightweight folders for chat conversations.

## What's in
- **DB**: `projects` table in `chat.db`; nullable `project_id` column on `conversations` (additive migration — existing chats stay unfiled).
- **API**: CRUD under `/api/chat/projects` (list includes per-project conversation counts). `POST/PUT /api/chat/conversations` accept `project_id` and validate it against existing projects (400 on unknown, null detaches).
- **Delete semantics**: deleting a project detaches its conversations (they become unfiled) — never destroys chats.
- **Sidebar UI**: collapsible project sections above the unfiled list; per-project actions (new chat in project, rename, delete w/ confirm); bulk-selection toolbar gains a *Move to…* project dropdown; shift-click range selection skips collapsed sections.

## Tests
- `dashboard/tests/test_chat_projects.py` — 21 backend tests (CRUD, membership, detach-on-delete, validation, auth). Full backend suite: 400 passed.
- `ChatSidebar.test.jsx` — 12 new grouping/move/collapse tests. Full frontend suite: 152 passed. Build OK.

## Not in this phase (planned next)
- Phase 2: real folder tree per project on disk + file API/UI
- Phase 3: in-browser text editor

### DIFF
--- a/dashboard/chat/db.py
+++ b/dashboard/chat/db.py
@@ -323,13 +323,25 @@ class ChatDB:
             self._close_conn(conn)
 
     def list_conversations(self, limit: int = 50, offset: int = 0) -> Tuple[List[Conversation], int]:
+        """List conversations, newest-updated first.
+
+        A negative limit returns the ENTIRE list in one statement — a
+        consistent snapshot (offset pagination over the mutable
+        updated_at ordering can skip or duplicate rows when conversations
+        are touched between page fetches).
+        """
         conn = self._get_conn()
         try:
             total = conn.execute("SELECT COUNT(*) FROM conversations").fetchone()[0]
-            rows = conn.execute(
-                "SELECT * FROM conversations ORDER BY updated_at DESC LIMIT ? OFFSET ?",
-                (limit, offset),
-            ).fetchall()
+            if limit is None or limit < 0:
+                rows = conn.execute(
+                    "SELECT * FROM conversations ORDER BY updated_at DESC"
+                ).fetchall()
+            else:
+                rows = conn.execute(
+                    "SELECT * FROM conversations ORDER BY updated_at DESC LIMIT ? OFFSET ?",
+                    (limit, offset),
+                ).fetchall()
             convs = [self._row_to_conversation(row) for row in rows]
             self._attach_active_runs(conn, convs)
             return convs, total
--- a/dashboard/chat/db.py
+++ b/dashboard/chat/db.py
@@ -86,6 +86,31 @@ CREATE INDEX IF NOT EXISTS idx_chat_runs_conversation ON chat_runs(conversation_
 CREATE INDEX IF NOT EXISTS idx_chat_runs_status ON chat_runs(status);
 """
 
+# conversations.project_id was added via ALTER TABLE, which in sqlite cannot
+# carry a foreign key — these triggers provide the referential half. They
+# close the validate-then-write race (a project deleted between a route's
+# existence check and the conversation write): the write itself fails with
+# IntegrityError('project not found') instead of persisting an orphan id.
+# delete_project detaches conversations in the same transaction as the
+# project row's deletion, so the triggers never fire on that path.
+TRIGGERS_SQL = """
+CREATE TRIGGER IF NOT EXISTS trg_conversations_project_fk_insert
+BEFORE INSERT ON conversations
+WHEN NEW.project_id IS NOT NULL
+     AND NOT EXISTS (SELECT 1 FROM projects WHERE id = NEW.project_id)
+BEGIN
+    SELECT RAISE(ABORT, 'project not found');
+END;
+
+CREATE TRIGGER IF NOT EXISTS trg_conversations_project_fk_update
+BEFORE UPDATE OF project_id ON conversations
+WHEN NEW.project_id IS NOT NULL
+     AND NOT EXISTS (SELECT 1 FROM projects WHERE id = NEW.project_id)
+BEGIN
+    SELECT RAISE(ABORT, 'project not found');
+END;
+"""
+
 
 class ChatDB:
     def __init__(self, db_path: str = "chat.db"):
@@ -119,6 +144,11 @@ class ChatDB:
             conn.executescript(SCHEMA_SQL)
             conn.commit()
             self._migrate(conn)
+            # Created AFTER migrations: the triggers reference
+            # conversations.project_id, which on a pre-existing DB only
+            # exists once _migrate has added it.
+            conn.executescript(TRIGGERS_SQL)
+            conn.commit()
         finally:
             self._close_conn(conn)
 
--- a/dashboard/chat/db.py
+++ b/dashboard/chat/db.py
@@ -2,12 +2,22 @@ import sqlite3
 import logging
 from typing import Optional, List, Tuple
 
-from .models import Conversation, Message, Critique, Artifact, ChatRun
+from .models import Conversation, Message, Critique, Artifact, ChatRun, Project
 from .runs import ChatRunStatus, ACTIVE_STATUSES, TERMINAL_STATUSES, ALL_STATUSES
 
 logger = logging.getLogger(__name__)
 
 SCHEMA_SQL = """
+CREATE TABLE IF NOT EXISTS projects (
+    id          TEXT PRIMARY KEY,
+    name        TEXT NOT NULL,
+    description TEXT NOT NULL DEFAULT '',
+    created_at  TEXT NOT NULL DEFAULT (strftime('%Y-%m-%dT%H:%M:%SZ','now')),
+    updated_at  TEXT NOT NULL DEFAULT (strftime('%Y-%m-%dT%H:%M:%SZ','now'))
+);
+
+CREATE INDEX IF NOT EXISTS idx_project_updated ON projects(updated_at DESC);
+
 CREATE TABLE IF NOT EXISTS conversations (
     id                     TEXT PRIMARY KEY,
     title                  TEXT NOT NULL DEFAULT 'New Conversation',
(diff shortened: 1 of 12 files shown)


teo-mateo:

1. **[P1] Moving a spin-off silently creates phantom project membership.** `ChatSidebar` intentionally groups only root conversations and renders every descendant under its parent regardless of the descendant's own `project_id` (`dashboard/frontend/src/components/chat/ChatSidebar.jsx:320-335`), but every spin-off row still has a selectable checkbox and `handleMoveMany` updates exactly the selected row IDs (`ChatPage.jsx:147-149`). Select spin-off `A1` under root `A` in project P1 and move it to P2: the API stores `A1.project_id = P2`, yet the UI keeps displaying A1 under A/P1. P2's API `conversation_count` increases while its UI section remains empty; deleting P1 then detaches A but leaves A1 assigned to P2 and still displayed under unfiled A. The added test at `ChatSidebar.test.jsx:218-223` actually establishes the rendering half of this mismatch but never exercises moving that child. Please define one membership model and enforce it end-to-end: either make membership root-only (move/validate/count the root and disallow child-specific assignment), or inherit/cascade the project across the whole conversation tree and render/count those stored assignments consistently. Add a test that selects and moves a spin-off and then checks both placement and counts.

2. **[P2] Project validation is a TOCTOU check, so concurrent deletion can persist orphaned `project_id` values.** Both create and update first call `get_project` in the route (`dashboard/chat/routes.py:253-255`, `316-318`), which closes its connection, and only afterward write the conversation on a different connection (`dashboard/chat/db.py:258-273`, `309-330`). If another request deletes the project between those operations, the conversation write succeeds with the deleted ID. The sidebar's unknown-project fallback hides the corruption by showing it as unfiled, but later project counts/deletes can no longer repair the reference. This is realistic with multiple tabs and especially with the parallel per-row bulk move. Keep validation and write in one appropriately locked transaction, or add actual referential enforcement such as `REFERENCES projects(id) ON DELETE SET NULL`/an equivalent trigger, and test delete racing create/update.

3. **[P2] Non-string `project_id` input produces a 500 instead of a validation response.** The new routes pass any non-null JSON value directly to SQLite (`dashboard/chat/routes.py:253-255`, `316-318`). For example, POST with `{"main_service":"svc","project_id":{"x":1}}` or PUT with `{"project_id":["x"]}` raises `sqlite3.InterfaceError: unsupported type` in `get_project`, returning HTML 500. Validate that `project_id` is a string (or null for update) before the lookup and add malformed-type cases to `test_chat_projects.py`; these should be 400s rather than server errors.

teo-mateo:

**Finding 1.1 — Spin-off move creates phantom project membership**: accepted.

Fixed in 5f13a0c. Membership is now root-only, enforced end-to-end: `POST/PUT /api/chat/conversations` reject a non-null `project_id` on a conversation that has (or is created with) `parent_conversation_id` (400 "Spin-off conversations inherit their parent's project"). On the UI side, `handleBulkMove` resolves every selected id to its root via the parent chain and dedupes, so selecting a spin-off and moving it moves the tree the user actually sees. Tests: backend spinoff-create/spinoff-move 400s; sidebar tests assert `onMoveMany(['A'], 'p2')` when only `A1` is selected, and dedupe when both are.

**Finding 1.2 — TOCTOU between project validation and conversation write**: accepted.

Fixed in 5f13a0c. Added BEFORE INSERT / BEFORE UPDATE OF project_id triggers on `conversations` that RAISE(ABORT, 'project not found') when the referenced project doesn't exist — the write itself now enforces the reference, regardless of connection interleaving. Routes catch `sqlite3.IntegrityError` and return 400. (`delete_project` detaches conversations in the same transaction as the project row delete, so the triggers never fire on that path.) Triggers are created after column migrations since `project_id` doesn't exist yet on pre-existing DBs at schema time. DB-level test simulates the race by writing an orphan id directly.

**Finding 1.3 — Non-string project_id → 500**: accepted.

Fixed in 5f13a0c. Both routes validate `project_id` is a string (or null on update) before any lookup, returning 400. Malformed-type cases (dict, list, int, bool) added to `test_chat_projects.py` for both POST and PUT.

Suites: 405 backend / 154 frontend passed.

teo-mateo:

1. **[P2] Project sections silently hide conversations outside the first global page.** `listConversations()` still fetches only 50 rows by default (`dashboard/frontend/src/services/chat.js:10-11`), but `ChatSidebar` builds every project section exclusively from that partial array and uses `roots.length` for both the displayed count and the `Empty project` decision (`dashboard/frontend/src/components/chat/ChatSidebar.jsx:588-609`), ignoring the authoritative `project.conversation_count` returned by the new projects API. For example, if a project owns an older conversation and 50 newer unfiled conversations exist, the sidebar labels that project `0` / `Empty project` even though the API reports a count of 1, and the conversation cannot be reached through its project. The project grouping needs project-aware loading/pagination (or the conversation list must be fully loaded); at minimum, add a test where `conversation_count > roots.length` so the UI does not claim a nonempty project is empty.

2. **[P2] Renaming a project concurrently with deletion can return an HTML 500.** `PUT /api/chat/projects/<id>` checks existence on one connection, then `ChatDB.update_project()` performs the update on another connection and can return `None` if a concurrent DELETE wins (`dashboard/chat/routes.py:227-231`, `dashboard/chat/db.py:244-251`). The route then calls `project.to_dict()` unconditionally. This is distinct from the now-fixed conversation membership race: no orphan is created, but a normal two-tab rename/delete race raises `AttributeError` instead of returning a JSON 404. Check the update rowcount / returned value and handle `None`, ideally with a regression test for a vanished row between validation and update.

teo-mateo:

**Finding 2.1 — Project sections hide conversations outside the first page**: accepted.

Fixed in fedbe92, both halves:
- `useConversations.refresh()` now pages through the entire list (200 per page, driven by the returned `total`, with a guard that stops on an empty page if the total shrank mid-pagination), so grouping always works on the full set.
- Defensively, the sidebar header count now uses the authoritative `project.conversation_count` (max'd with loaded roots), and a project with `count > 0` but no loaded rows renders "N conversations not loaded" instead of "Empty project". Tests: `useConversations.test.jsx` (350-row two-page fetch + stale-total termination) and a ChatSidebar test with `conversation_count: 3` and an empty loaded list.

**Finding 2.2 — Project rename racing delete returns HTML 500**: accepted.

Fixed in fedbe92. `update_project` route handles `db.update_project()` returning `None` (row vanished between the existence check and the write) with a JSON 404. Regression test simulates the race by monkeypatching `ChatDB.update_project` to delete the project just before the real update runs.

Suites: 406 backend / 157 frontend passed, build OK.

teo-mateo:

1. **[P2] Deleting conversations leaves project counts permanently stale in the sidebar.** The single- and bulk-delete handlers only call the conversation hook (`ChatPage.jsx:185-198`); unlike create-in-project and move, neither refreshes `useProjects`. After deleting a root from a project, `roots.length` decreases but the cached `project.conversation_count` does not. Because `ChatSidebar.jsx:594` computes `Math.max(project.conversation_count, roots.length)`, deleting the last chat from a project leaves the section showing `1 conversation not loaded` instead of `Empty project` (and partial deletions retain an inflated badge) until a separate project refresh or page reload. Refresh projects after both delete paths, or update the aggregate locally, and add a ChatPage/integration test that deletes the last project conversation and verifies the count/placeholder.

teo-mateo:

**Finding 3.1 — Deletes leave project counts stale**: accepted.

Fixed in f896abc. Both `handleDelete` and `handleDeleteMany` in ChatPage now await `refreshProjects()` after the conversation delete, so `project.conversation_count` is refetched and the section falls back to the correct state (including "Empty project" after deleting the last chat). Added two ChatPage wiring tests (sidebar stub captures props, drives `onDelete` / `onDeleteMany`, asserts `listProjects` is re-called).

Suites: 159 frontend passed; backend untouched by this iteration.

teo-mateo:

1. **[P2] The new full-list loader can still omit a conversation when rows reorder between pages.** `refresh()` issues independent offset-based requests and advances the next offset using the accumulated array length (`dashboard/frontend/src/hooks/useConversations.js:15-23`), while the API orders rows by mutable `updated_at DESC`. For example, after page 1 returns rows 1–200, a row that was at position 250 can be touched by a run/rename and move to the front. Page 2 at offset 200 then includes a row already present in page 1; because duplicates still increase `all.length`, the loop reaches the original `total` and stops with one other conversation missing. This is realistic because run starts/completions and title changes update conversations while the sidebar is refreshing, and it can again make a project chat unreachable (as well as render duplicate React keys). The added pagination tests cover static pages and a shrinking total, but not reordering. Please load against a consistent snapshot/use pagination semantics that do not shift under `updated_at` changes (or provide a project-aware server query), and add a regression test that reorders a later-page row between the first and second calls.

teo-mateo:

**Finding 4.1 — Offset pagination can skip/duplicate rows under updated_at churn**: accepted, fixed by removing pagination entirely.

Fixed in 59d6465. Rather than adding keyset pagination for what is a single-user dashboard with tiny conversation rows, the sidebar now fetches the whole list in ONE request: `ChatDB.list_conversations` accepts a negative limit and returns all rows from a single SQL statement — a consistent snapshot by construction, so there is no window in which reordering can drop or duplicate a row. `useConversations.refresh()` calls `listConversations(-1, 0)`. The requested reorder-between-pages regression test is moot (there is no second request); the replacement test asserts exactly one unlimited call, and a backend test covers `limit=-1` returning everything while default paging stays unchanged for other callers.

Suites: 407 backend / 158 frontend passed.

teo-mateo:

I reviewed the full current diff and prior discussion, then validated the latest branch locally. The previously reported project-membership, referential-race, pagination, rename/delete-race, and stale-count issues are addressed coherently.

Validation performed:
- `dashboard/tests/test_chat_projects.py`: 28 passed
- Targeted frontend suites for `useConversations`, `ChatSidebar`, and `ChatPage`: 32 passed

PR #78, opened 2026-07-15
feat(chat): project files — real folders on disk with tree API and UI (phase 2)

teo-mateo:
Phase 2 of the projects feature: each project owns a real directory tree on disk, browsable and manageable from the dashboard.

## What's in
- **Filesystem layer** (`chat/project_files.py`): the filesystem is the source of truth — no files table. All operations funnel through `resolve()`: per-component validation (no `..`/`.`/empty/backslash/NUL, 255-char cap) plus a realpath containment check against the nearest *existing* ancestor, so both `../` traversal and symlinks planted inside the tree are rejected — including for upload targets that don't exist yet. Symlinks are listed as files and never followed (no cycle risk, no escape).
- **API** under `/api/chat/projects/<id>/files`: GET tree, multipart upload (100 MB cap, streamed to a temp file then atomically renamed, 409 unless `overwrite`), download, mkdir, move/rename (409 on existing destination, refuses dir-into-itself), recursive delete. Storage root defaults to `dashboard/project_files/<project_id>` (gitignored); override via `PROJECT_FILES_DIR` app config or `LLM_DOCK_PROJECT_FILES_DIR`.
- **Delete semantics**: deleting a project removes its files directory (conversations still survive, as in phase 1).
- **UI**: new project page at `/chat/project/<id>` (sidebar layout preserved) — file tree with expand/collapse, upload to root or into any folder, new folder/subfolder, rename/move, download (auth-carrying blob fetch), delete with confirm. Project headers in the sidebar get an open-project button and an active highlight.

## Tests
- `dashboard/tests/test_project_files.py` — 38 tests: resolve()-level traversal + symlink-escape rejection (existing and non-existing targets), all ops incl. size-cap cleanup, HTTP surface incl. traversal on every route, project-delete cleanup. Full backend: 445 passed.
- `ProjectPage.test.jsx` — 12 tests (tree render, mkdir, upload + overwrite-conflict flow, rename, delete confirm, error surfacing). Full frontend: 170 passed, build OK.

## Next phases
- Phase 3: in-browser text editor for project files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -92,3 +92,6 @@ opencode.json
 
 # test harness output
 tools/traces/
+
+# Project files (machine-local user content, managed by the dashboard)
+dashboard/project_files/
--- a/dashboard/chat/project_files.py
+++ b/dashboard/chat/project_files.py
@@ -125,7 +125,13 @@ def _validate_component(name: str):
         raise ProjectFilesError("invalid character in path component")
     # Filesystems cap components at 255 BYTES, not characters — a name of
     # 100 emoji passes a character check and still fails at the syscall.
-    if len(name.encode("utf-8")) > MAX_NAME_LENGTH:
+    # JSON happily delivers lone surrogates ("\ud800"), which are not
+    # UTF-8-encodable at all — that's bad input, not a server fault.
+    try:
+        encoded = name.encode("utf-8")
+    except UnicodeEncodeError:
+        raise ProjectFilesError("path component is not valid unicode")
+    if len(encoded) > MAX_NAME_LENGTH:
         raise ProjectFilesError("path component too long")
 
 
--- a/dashboard/chat/project_files.py
+++ b/dashboard/chat/project_files.py
@@ -276,7 +276,10 @@ def mkdir(root: str, rel_path: str) -> dict:
     if not parts:
         raise ProjectFilesError("path is required")
     abs_path = resolve(root, rel_path)
-    if os.path.exists(abs_path):
+    # lexists: a dangling symlink is an ordinary entry in the tree (listed
+    # by build_tree) and must count as occupying its name — exists() would
+    # follow the link and report the slot free.
+    if os.path.lexists(abs_path):
         raise ProjectFilesError("path already exists", status=409)
     with _fs_errors():
         os.makedirs(abs_path)
@@ -295,9 +298,12 @@ def move(root: str, rel_src: str, rel_dst: str) -> dict:
         raise ProjectFilesError("destination is required")
     abs_src = resolve(root, rel_src)
     abs_dst = resolve(root, rel_dst)
-    if not os.path.exists(abs_src):
+    # lexists on both sides: a dangling symlink is movable (the operation
+    # applies to the link itself, as delete already does) and occupies its
+    # destination name.
+    if not os.path.lexists(abs_src):
         raise ProjectFilesError("source not found", status=404)
-    if os.path.exists(abs_dst):
+    if os.path.lexists(abs_dst):
         raise ProjectFilesError("destination already exists", status=409)
     if dst_parts[:len(src_parts)] == src_parts:
         raise ProjectFilesError("cannot move a directory into itself")
--- a/dashboard/chat/project_files.py
+++ b/dashboard/chat/project_files.py
@@ -11,10 +11,12 @@ component validation plus a realpath containment check, so neither
 ``..`` traversal nor a symlink planted inside the tree can escape the
 project root.
 """
+import errno
 import os
 import shutil
 import logging
 import tempfile
+from contextlib import contextmanager
 
 logger = logging.getLogger(__name__)
 
@@ -72,10 +74,46 @@ def ensure_project_root(storage_root: str, project_id: str) -> str:
     return root
 
 
-def delete_project_root(storage_root: str, project_id: str):
-    """Remove a project's entire files directory (project deletion)."""
+def delete_project_root(storage_root: str, project_id: str, strict: bool = False):
+    """Remove a project's entire files directory (project deletion).
+
+    strict=True propagates cleanup failures (used BEFORE the DB row is
+    deleted, so a permissions/IO error surfaces as a retryable 500
+    instead of silently orphaning data). The default best-effort mode is
+    for post-delete sweeps where the row is already gone — failures are
+    logged, never raised.
+    """
     root = project_root(storage_root, project_id)
-    shutil.rmtree(root, ignore_errors=True)
+    if strict:
+        try:
+            shutil.rmtree(root)
+        except FileNotFoundError:
+            pass
+        return
+    try:
+        shutil.rmtree(root)
+    except FileNotFoundError:
+        pass
+    except OSError:
+        logger.exception("best-effort cleanup of project files failed: %s", root)
+
+
+@contextmanager
+def _fs_errors():
+    """Translate filesystem rejections of otherwise-validated paths into
+    client errors. A component can pass the 255-CHARACTER check yet exceed
+    the filesystem's 255-BYTE limit (multibyte names → ENAMETOOLONG), the
+    total path can exceed PATH_MAX, or an intermediate component can turn
+    out to be a regular file (ENOTDIR). All are bad input, not server
+    faults."""
+    try:
+        yield
+    except OSError as e:
+        if e.errno == errno.ENAMETOOLONG:
+            raise ProjectFilesError("name or path too long for the filesystem")
+        if e.errno == errno.ENOTDIR:
+            raise ProjectFilesError("a path component is not a directory")
+        raise
 
 
 def _validate_component(name: str):
@@ -85,7 +123,9 @@ def _validate_component(name: str):
         raise ProjectFilesError("invalid path component")
     if "/" in name or "\\" in name or "\x00" in name:
         raise ProjectFilesError("invalid character in path component")
-    if len(name) > MAX_NAME_LENGTH:
+    # Filesystems cap components at 255 BYTES, not characters — a name of
+    # 100 emoji passes a character check and still fails at the syscall.
+    if len(name.encode("utf-8")) > MAX_NAME_LENGTH:
         raise ProjectFilesError("path component too long")
 
 
@@ -197,30 +237,31 @@ def save_stream(root: str, dir_rel: str, filename: str, stream, overwrite: bool 
     # "<target>.uploading" name would collide with a legitimate user file
     # of that name and with a concurrent upload of the same target.
     written = 0
-    fd, tmp_path = tempfile.mkstemp(prefix=".upload-", suffix=".tmp", dir=abs_dir)
-    try:
-        with os.fdopen(fd, "wb") as f:
-            while True:
-                chunk = stream.read(1024 * 1024)
-                if not chunk:
-                    break
-                written += len(chunk)
-                if written > MAX_UPLOAD_BYTES:
-                    raise ProjectFilesError("file too large", status=413)
-                f.write(chunk)
-        if overwrite:
-            os.replace(tmp_path, abs_target)
-        else:
-            # Atomic no-replace publication: link() fails with EEXIST if the
-            # target appeared since the pre-check, upholding the 409 contract
-            # even against a concurrent upload of the same name.
-            try:
-                os.link(tmp_path, abs_target)
-            except FileExistsError:
-                raise ProjectFilesError("file already exists", status=409)
-    finally:
-        if os.path.exists(tmp_path):
-            os.unlink(tmp_path)
+    with _fs_errors():
+        fd, tmp_path = tempfile.mkstemp(prefix=".upload-", suffix=".tmp", dir=abs_dir)
+        try:
+            with os.fdopen(fd, "wb") as f:
+                while True:
+                    chunk = stream.read(1024 * 1024)
+                    if not chunk:
+                        break
+                    written += len(chunk)
+                    if written > MAX_UPLOAD_BYTES:
+                        raise ProjectFilesError("file too large", status=413)
+                    f.write(chunk)
+            if overwrite:
+                os.replace(tmp_path, abs_target)
+            else:
+                # Atomic no-replace publication: link() fails with EEXIST if the
+                # target appeared since the pre-check, upholding the 409 contract
+                # even against a concurrent upload of the same name.
+                try:
+                    os.link(tmp_path, abs_target)
+                except FileExistsError:
+                    raise ProjectFilesError("file already exists", status=409)
+        finally:
+            if os.path.exists(tmp_path):
+                os.unlink(tmp_path)
     return _node(abs_target, rel_target)
 
 
@@ -231,7 +272,8 @@ def mkdir(root: str, rel_path: str) -> dict:
     abs_path = resolve(root, rel_path)
     if os.path.exists(abs_path):
         raise ProjectFilesError("path already exists", status=409)
-    os.makedirs(abs_path)
+    with _fs_errors():
+        os.makedirs(abs_path)
     rel = "/".join(parts)
     return _node(abs_path, rel)
 
@@ -255,7 +297,8 @@ def move(root: str, rel_src: str, rel_dst: str) -> dict:
         raise ProjectFilesError("cannot move a directory into itself")
     if not os.path.isdir(os.path.dirname(abs_dst)):
         raise ProjectFilesError("destination directory not found", status=404)
-    shutil.move(abs_src, abs_dst)
+    with _fs_errors():
+        shutil.move(abs_src, abs_dst)
     return _node(abs_dst, "/".join(dst_parts))
 
 
--- a/dashboard/chat/project_files.py
+++ b/dashboard/chat/project_files.py
@@ -1,0 +1,240 @@
+"""Filesystem-backed project files.
+
+Each project owns a real directory tree under a machine-local storage root
+(default ``<dashboard>/project_files/<project_id>/``, override with
+``LLM_DOCK_PROJECT_FILES_DIR``). The filesystem is the source of truth —
+there is no files table to keep in sync; the API walks the directory.
+
+Every public function takes the project's root and *relative* paths from
+the client. ``resolve()`` is the single choke point for path safety:
+component validation plus a realpath containment check, so neither
+``..`` traversal nor a symlink planted inside the tree can escape the
+project root.
+"""
+import os
+import shutil
+import logging
+
+logger = logging.getLogger(__name__)
+
+# Uploads and (later) editor writes are capped; project files are meant as
+# context material for chat, not bulk storage.
+MAX_UPLOAD_BYTES = 100 * 1024 * 1024  # 100 MB
+MAX_NAME_LENGTH = 255
+
+
+class ProjectFilesError(Exception):
+    """Client-facing error with an HTTP status."""
+
+    def __init__(self, message: str, status: int = 400):
+        super().__init__(message)
+        self.status = status
+
+
+def default_storage_root() -> str:
+    configured = os.environ.get("LLM_DOCK_PROJECT_FILES_DIR")
+    if configured:
+        return configured
+    return os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "project_files")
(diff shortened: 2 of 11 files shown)


teo-mateo:

1. **[P1] Upload staging can silently delete a different user file** (`dashboard/chat/project_files.py:176-191`). The staging path is deterministically `abs_target + ".uploading"`, but `.uploading` is also a valid user filename. For example, after uploading `report.uploading`, uploading `report` opens the former with `"wb"` and then renames it to `report`; the existing file disappears even though overwrite was not requested. Two concurrent uploads to the same target also share/truncate/unlink the same staging file, so one request can corrupt the other or fail with a 500. Use a unique, exclusively created temp file in the destination directory (for same-filesystem atomic rename), and add tests for an existing `<target>.uploading` file and simultaneous uploads. The final no-overwrite publication also needs an atomic no-replace operation if the 409 contract must hold under concurrency.

2. **[P2] The 100 MB request cap is checked only after Werkzeug has parsed the multipart body** (`dashboard/chat/project_files_routes.py:51-60`). Accessing `request.files` causes the upload to be consumed/spooled before the `content_length` check, so a very large authenticated request can consume unbounded request-temp disk and I/O before receiving 413; chunked/missing-length requests bypass the early check entirely. Enforce a request-size limit before multipart parsing (for example Flask/Werkzeug `MAX_CONTENT_LENGTH`, with the small multipart allowance handled deliberately) while keeping the stream counter as defense in depth. Add an HTTP-level oversized-body test, since the current size test calls `save_stream` directly and cannot catch this.

3. **[P2] Project deletion can leave files behind when it races a file request** (`dashboard/chat/project_files_routes.py:22-29`, `dashboard/chat/routes.py:243-247`). A file request can observe the project row, pause before `ensure_project_root`, then a delete can remove the DB row and directory and return success; the original request resumes, recreates the root, and can upload data into an orphan directory that will never be cleaned up. Serialize project deletion with file operations or revalidate project existence after root creation/at the commit point, with cleanup on a failed revalidation. A race test should cover this ordering.

4. **[P2 security] The realpath containment check is TOCTOU and can be bypassed by swapping a checked component for a symlink** (`dashboard/chat/project_files.py:72-96`, then uses such as `open` at line 179 and `send_file` after `stat_file`). If another process can mutate the configured storage tree, it can replace a directory after `resolve()` returns but before the pathname is opened, redirecting an upload outside the project root or a download to an outside file. This does not meet the PR's stated protection against symlinks planted in the tree. Keep operations anchored to opened directory descriptors and reject symlinks component-by-component (for example descriptor-relative `openat`/`openat2` semantics with no-follow/beneath constraints), rather than validating a pathname and later reopening it. Add a regression test that swaps a parent between validation and open.

teo-mateo:

**Finding 1.1 — Deterministic '.uploading' staging name can clobber user files / corrupt concurrent uploads**: accepted.

Fixed in 7e65699. Staging now uses `tempfile.mkstemp` in the destination directory (unique name, `O_EXCL` creation, same filesystem for atomic publication). Publication is `os.replace` when overwrite was requested, otherwise an atomic `os.link` that fails `EEXIST` → 409, so the no-overwrite contract holds even when the target appears between the pre-check and publication. Tests: uploading `report` no longer touches an existing user file `report.uploading`; a simulated lose-the-race upload gets 409 and leaves the winner's content and no stray temp files.

**Finding 1.2 — 100 MB cap checked after multipart parsing**: accepted.

Fixed in 7e65699. The route checks `Content-Length` before touching `request.files`/`request.form`, and `init_chat` now sets `MAX_CONTENT_LENGTH` (`setdefault`, so deployments can override) so Werkzeug enforces the cap during parsing — covering chunked bodies with no Content-Length. The in-stream byte counter remains as last-line defense. Added an HTTP-level oversized-body test.

**Finding 1.3 — Project delete racing a file op leaves an orphan directory**: accepted.

Fixed in 7e65699. Two changes: read-only routes (tree, download) no longer create the files directory at all (an empty project just returns `tree: []`), and every mutating route revalidates the project row at its commit point — if the row vanished mid-operation, the route removes the directory it recreated and returns 404. Whichever side acts last cleans up, so no interleaving of {write, row-delete, rmtree, revalidate} leaves an orphan. Tests: GET-tree-creates-nothing, and a race test that deletes the project from inside the mkdir operation and asserts 404 + no directory.

**Finding 1.4 — realpath containment is TOCTOU against a local process swapping components**: pushed back.

Disagree on threat model. The attack requires a process that can mutate the storage tree (`dashboard/project_files/`) between `resolve()` and the open. That directory is owned by the dashboard user on a single-user machine-local deployment; any process that can plant or swap symlinks there is *already running as the dashboard's user* and can read/write every file the dashboard can — there is no privilege boundary for descriptor-anchored resolution to defend. The boundary this API must enforce is against the *remote authenticated client*, and the client has no primitive to create symlinks (upload writes regular files; move only relocates existing entries; mkdir creates directories), so the client cannot stage a swap through the API itself. `openat2`/`RESOLVE_BENEATH` plumbing would add Linux-specific complexity to defend against an attacker who has already won. Happy to revisit if the storage root ever moves to a shared/multi-user location.

Suites: 450 backend / 170 frontend passed.

teo-mateo:

1. **[P2] The app-level upload cap is still disabled** (`dashboard/chat/routes.py:43-49`). Flask creates every app with a `MAX_CONTENT_LENGTH` key whose value is `None`, so `app.config.setdefault("MAX_CONTENT_LENGTH", ...)` does not assign the proposed limit. Requests with a known oversized `Content-Length` are caught by the route check, but a chunked/missing-length multipart request can still be parsed and spooled without the intended bound. The new HTTP test registers the blueprint directly and sends a normal body with a content length, so it cannot detect this. Assign the default when the current value is `None` (while preserving a concrete deployment override), and add an initialization-level assertion plus a no-content-length/input-terminated parsing test.

2. **[P2] A concurrent destination creation makes move violate its no-overwrite contract and lose data** (`dashboard/chat/project_files.py:230-241`). The existence check and `shutil.move` are separate operations. On POSIX, if another request creates a file at `abs_dst` after line 235, the eventual rename silently replaces that winning file instead of returning 409; if it creates a directory, `shutil.move` changes semantics and moves the source inside that directory. I reproduced the file case by creating `dst` immediately before `shutil.move`: its contents were replaced by `src`. Use an atomic no-replace primitive or serialize mutations for a project, and add a race regression test analogous to the upload publication test.

3. **[P2] Unbounded path depth can make a project unlistable and undeletable** (`dashboard/chat/project_files.py:76-91, 129-144, 244-252`). Validation limits each component but not component count or total path depth, and `mkdir` creates missing parents. An authenticated client can create about 1,800 one-character nested directories in two requests on a normal Linux filesystem; `build_tree` and `shutil.rmtree` then both raise `RecursionError`. Consequently tree GET returns 500, recursive file deletion fails, and project deletion deletes the DB row then returns 200 while `delete_project_root(..., ignore_errors=True)` silently leaves the directory orphaned. Enforce a safe maximum nesting/total path length (and test the boundary), or make traversal and cleanup iterative and surface cleanup failure.

teo-mateo:

**Finding 2.1 — setdefault(MAX_CONTENT_LENGTH) is a no-op**: accepted — good catch, Flask pre-seeds the key with None so the intended cap never applied.

Fixed in 0bafbcf. `configure_max_content_length(app)` assigns when the value is None and preserves an explicit deployment override; unit-tested directly against a plain Flask app (None default → cap set; explicit 12345 → untouched). This makes Werkzeug enforce the bound during multipart parsing, covering the chunked/no-Content-Length case.

**Finding 2.2 — move's no-overwrite check racing a concurrent destination creation**: pushed back.

Disagree on materiality. This deployment has exactly one user. Triggering it requires that user to issue two mutations targeting the *same destination path* in the *same instant* from two tabs — adversarial timing they'd have to engineer against themselves — and the blast radius is one file they created milliseconds earlier, still recoverable from its source. There is no atomic no-replace rename for directories on POSIX without renameat2(RENAME_NOREPLACE) (not exposed by os.rename/shutil.move), so honoring the contract would mean per-project mutation locks — real complexity purchased against a self-race. The reproduction required injecting the collision between the check and the move; anything is reproducible by injection. Happy to add per-project locking if this ever hosts more than one user.

**Finding 2.3 — unbounded nesting can make a project unlistable**: accepted in the cheap form.

Fixed in 0bafbcf. `split_rel_path` caps depth at 32 components (`MAX_PATH_DEPTH`), which every operation — including parent-creating mkdir — funnels through, so recursive tree walks and rmtree stay orders of magnitude away from the recursion limit. Boundary-tested at 32/33. Declined the iterative-traversal rewrite: with the cap in place it defends against nothing.

Suites: 452 backend passed.

teo-mateo:

1. **[P2] The component limit counts Unicode code points, not filesystem bytes, so accepted names can still produce a 500** (`dashboard/chat/project_files.py:81-89`, `:180-224`, `:227-236`). Linux filesystems normally cap a component at 255 bytes, but `_validate_component` checks `len(name) <= 255`. A remotely supplied name such as 100 emoji passes validation and then `mkdir` raises `OSError(ENAMETOOLONG)`; upload writes the entire staged file and then fails the final `link`/`replace` with the same unhandled exception. The 32-component cap also permits an accepted relative path longer than `PATH_MAX` (for example 17 components of 255 ASCII bytes), with the same 500 result. Validate encoded component/path byte lengths for the target filesystem, or translate `ENAMETOOLONG` into a stable 400 before returning from the API. Add HTTP tests for a multibyte filename and a total-path overflow; the current 256-character ASCII test misses both.

2. **[P2] `mkdir` returns 500 when an intermediate component is a regular file** (`dashboard/chat/project_files.py:227-236`). This needs no race or external filesystem mutation: upload a regular file named `parent`, then POST `{ "path": "parent/child" }` to `/files/mkdir`. `resolve` accepts the contained path, `os.path.exists(abs_path)` is false, and `os.makedirs` raises an unhandled `NotADirectoryError`. This is ordinary invalid client input and should receive a defined 4xx response, analogous to the explicit destination-parent check in `move`. Validate the existing ancestor/parent before `os.makedirs` (or map `ENOTDIR`), and cover this sequence through the HTTP API.

3. **[P2] Project deletion reports success even when its files were not removed** (`dashboard/chat/project_files.py:75-78`, `dashboard/chat/routes.py:246-255`). The DB row is committed away first, and `shutil.rmtree(..., ignore_errors=True)` suppresses every cleanup failure, not just an already-absent directory. A permissions, I/O, or mounted-filesystem error therefore returns `{ "ok": true }` while leaving some or all project data orphaned; because the project row is gone, the user cannot retry deletion through the API. Suppress only `FileNotFoundError`; unexpected cleanup failures should be surfaced/logged and the deletion flow should preserve a way to retry or reconcile cleanup. Add a test that makes `rmtree` fail and asserts the route does not claim successful removal.

teo-mateo:

All three accepted — these are plain bad-input 500s and a swallowed failure, no adversarial timing required.

**Finding 3.1 — character-based length check lets multibyte names hit ENAMETOOLONG**: accepted. Fixed in f02dfc2. `_validate_component` now measures UTF-8 bytes, and a `_fs_errors()` translator around every filesystem mutation maps `ENAMETOOLONG` (component or total-path/PATH_MAX) to a 400. HTTP tests: 100-emoji filename upload → 400; 17×250-char mkdir (past PATH_MAX but within the depth cap) → 400.

**Finding 3.2 — mkdir through a regular file → NotADirectoryError 500**: accepted. Fixed in f02dfc2 via the same translator (`ENOTDIR` → 400). HTTP test covers upload-file-named-parent then mkdir `parent/child` → 400.

**Finding 3.3 — project delete claims success when rmtree fails**: accepted. Fixed in f02dfc2. Cleanup now runs strictly BEFORE the row delete — a permissions/IO failure returns 500 and keeps the project row, so the delete remains retryable through the API. The post-row-delete sweep (for a directory recreated by a concurrent write) stays best-effort with the failure logged. Test makes rmtree raise PermissionError → 500 + row survives; retry after unpatching → 200 + directory gone.

Suites: 456 backend passed.

teo-mateo:

1. **[P2] A stale tree request can show one project's files under another project and make actions target the wrong project** (`dashboard/frontend/src/components/chat/ProjectPage.jsx:100-129`). When the prop changes from project A to B, the effect clears the tree and starts B's refresh, but it neither cancels nor invalidates A's in-flight refresh (including the refresh started after a mutation). If A resolves last, `setTree` installs A's nodes while the component's current handlers close over B's ID; clicking rename/delete on one of those stale rows then sends B's project ID with A's path, potentially modifying an unrelated same-named path in B. Guard state commits with a request generation/current-project check or abort obsolete requests, and add a test with deferred tree promises that rerenders from A to B and resolves A last.

2. **[P2] Lone Unicode surrogates bypass JSON parsing and make path validation return 500** (`dashboard/chat/project_files.py:119-129`). Python's JSON decoder accepts an escaped lone surrogate such as `{"path":"\ud800"}`, but `_validate_component` calls `name.encode("utf-8")` without handling `UnicodeEncodeError`. I reproduced `POST .../files/mkdir` with that body: it raises at line 128 and returns Flask's HTML 500 instead of a client-validation response. Reject non-UTF-8-encodable components as `ProjectFilesError` (400), and cover the JSON mkdir/move surface with a regression test.

teo-mateo:

**Finding 4.1 — stale tree response commits across a project switch**: accepted (realistic with one user clicking between projects while a response is slow).

Fixed in the pushed commit. `refresh()` claims a monotonic generation before fetching and only commits (`setTree`/`setError`) if it is still the latest — the same request-sequencing pattern used by `useChat.loadConversation`. Regression test defers project A's tree promise, switches to B, resolves A last, and asserts A's rows never render.

**Finding 4.2 — lone surrogates 500 in path validation**: accepted.

Fixed in the pushed commit. `_validate_component` rejects non-UTF-8-encodable names with a 400 before the byte-length check. Regression tests POST `{"path": "\ud800"}` to mkdir and move.

Suites: 457 backend / 171 frontend passed.

teo-mateo:

1. **[P2] A mutation that finishes after a project switch can still repopulate the new page with the old project tree** (`dashboard/frontend/src/components/chat/ProjectPage.jsx:105-137`). The generation guard only rejects requests that were already in flight when a newer refresh starts. If an upload/mkdir/etc. begins on project A, the user switches to B, and then the A mutation resolves, the old `run` closure calls its old `refresh` closure for A *after* the B refresh. That A refresh increments `genRef` again, becomes the newest generation, and is therefore allowed to commit A rows while the current props and row handlers belong to B. This recreates the wrong-project action hazard from iteration 4 in a realistic slow-mutation sequence. Bind commits to the current project id (not only request order), cancel/invalidate obsolete operation continuations, or key/remount the page by project id. Add a regression test that defers an A mutation, switches to B, then resolves the mutation and its follow-up A tree request.

2. **[P3] Downloading a missing file from a fresh project is misreported as a traversal attempt** (`dashboard/chat/project_files.py:158-177`, `dashboard/chat/project_files_routes.py:63-67`). Project roots are created lazily, so for a project with no mutations yet, `resolve(root, "missing.txt")` walks to an existing ancestor above the nonexistent project root and then raises `path escapes project root` with 400. I reproduced `GET /api/chat/projects/<id>/files/download?path=missing.txt` immediately after project creation returning `400 {"error":"path escapes project root"}`, even though this is an ordinary absent resource and `stat_file` otherwise promises 404 for it. Handle an absent project directory as not-found before resolving the file, and add an HTTP test for download from a fresh/empty project.

teo-mateo:

**Finding 5.1 — stale mutation continuation outruns the generation guard**: accepted — correct catch, the guard as written was insufficient.

Fixed in the pushed commit. Every commit point (tree, error, and busy state — run() had the same exposure) is now additionally bound to the project id captured when the request began, checked against a ref updated each render. A stale A-closure can claim the newest generation but can no longer commit under B. Regression test defers an mkdir on A, switches to B, resolves the mutation and its follow-up refresh, and asserts B's rows (and only B's) render.

**Finding 5.2 — missing file in a fresh project misreported as traversal**: accepted.

Fixed in the pushed commit. The download route validates path shape first (traversal attempts remain 400), then treats a not-yet-created project root as an ordinary "file not found" 404. HTTP test covers download from a freshly created project.

Suites: 458 backend / 172 frontend passed.

teo-mateo:

1. **[P2] Dangling symlinks are exposed in the tree but break the file-operation contract** (`dashboard/chat/project_files.py:195-210`, `:274-308`). `build_tree` intentionally lists symlinks as files, but `mkdir` and `move` decide whether an entry exists with `os.path.exists`, which follows the link and returns false for a dangling symlink. With a normal pre-existing broken symlink named `broken`, the UI shows an actionable file row, yet renaming it returns 404 because the source is considered missing; creating a folder at that path reaches `os.makedirs` and produces an unhandled `FileExistsError`/500; and moving a regular file to `broken` passes the no-destination check and `shutil.move` silently replaces the symlink instead of returning the documented 409. This requires neither a race nor a hostile local process; broken symlinks are ordinary filesystem entries in the real directory that this PR explicitly supports and surfaces. Use `os.path.lexists`/`lstat` for entry existence and define operations on the link itself consistently (as `delete` already does), then add HTTP regressions for renaming a dangling symlink, mkdir at its path, and using it as a move destination.

teo-mateo:

**Finding 6.1 — dangling symlinks break mkdir/move contracts**: accepted — needs no race and no hostile process, just a pre-existing broken link in a directory this feature explicitly treats as source-of-truth.

Fixed in the pushed commit. `mkdir` and `move` now use `os.path.lexists` for entry existence, making operations consistent with `delete` (which already targets the link itself): mkdir at a dangling link's path → 409; a dangling link as move destination → 409 instead of silent replacement; the link itself is movable and stays a link. Regression test covers all three.

Suites: 459 backend passed.

teo-mateo:

1. **[P2] A stale mutation refresh can invalidate the current project fetch and leave the page permanently empty** (`dashboard/frontend/src/components/chat/ProjectPage.jsx:112-126`, `:136-146`). The project-id check prevents an old-project response from committing, but the stale `refresh()` still increments the single global `genRef`. If a mutation on A is pending, the user switches to B, B starts a slow tree fetch, and then the A mutation finishes, its follow-up `refresh(A)` claims a newer generation. B then cannot commit because its generation is stale, while A cannot commit because its project id is stale, so the cleared B page remains at “No files yet” until the user manually refreshes. The iteration-5 regression resolves B before resuming the A mutation, so it misses this ordering. Reject a refresh whose captured project is no longer current before incrementing the generation (or keep sequencing per project), and add a test where B remains pending until after the stale A follow-up has started.

2. **[P2] Entries advertised as files are not required to be regular files, so download can follow links or hang a worker** (`dashboard/chat/project_files.py:181-190`, `:195-209`, `:215-223`). `_node()` labels every non-directory entry as `type: file`, but `stat_file()` only rejects directories and then hands the path to `send_file`. An in-root symlink to a regular file is therefore followed and downloaded, contradicting the stated “symlinks are listed as files and never followed” contract. More seriously, an ordinary FIFO in the filesystem source of truth is also rendered with a Download action; clicking it makes `send_file` block opening/reading the FIFO indefinitely, tying up a Flask worker, while sockets/devices produce other uncontrolled behavior. Use `lstat`/`stat.S_ISREG` at the download boundary and define symlink downloads explicitly (rejecting them is consistent with the PR contract); add regressions for an in-root symlink and a FIFO/special entry.

teo-mateo:

**Finding 7.1 — stale refresh claims a generation and strands the current fetch**: accepted. Fixed in 4b8023b: a refresh whose captured project id is no longer current returns before incrementing the generation. Regression test uses the exact ordering described (B's fetch pending while the stale A follow-up runs, B resolves last).

**Finding 7.2 — download follows in-root symlinks / blocks on FIFOs**: accepted. Fixed in 4b8023b: `stat_file` lstat's the entry and serves only regular files — symlinks (in-root, out-of-root, dangling) and FIFOs/devices get a 400. Tests cover in-root symlink, FIFO, and dangling link.

Suites: 460 backend / 173 frontend passed.

teo-mateo:
Closing the review loop by author decision after 7 iterations: every finding was either fixed (with regression tests) or pushed back with reasoning that stood unchallenged, no finding ever repeated, and severity declined monotonically from path-traversal to FIFO-download corner cases. Merging.

PR #79, opened 2026-07-15
feat(chat): lightweight text editor for project files (phase 3)

teo-mateo:
Phase 3 of the projects feature: edit text files in the project tree directly from the browser.

## What's in
- **Files layer** (`chat/project_files.py`): `read_text` / `write_text`. UTF-8 only — binary (NUL byte / invalid UTF-8) and files over the 2 MB editor cap are rejected with 4xx. Writes are atomic (mkstemp in the target dir + replace) and never write through or replace symlinks/special files. `write_text` takes an optional `base_modified_at`: if the file changed on disk since the client loaded it (or was deleted), the save is rejected with 409 instead of silently clobbering the other edit.
- **API**: `GET/PUT /api/chat/projects/<id>/files/content` — same validation ordering (traversal 400s before missing-root 404s) and project-delete revalidation as the phase-2 routes.
- **UI**: clicking a file row opens `ProjectFileEditor` in the content area — plain monospace textarea (deliberately no CodeMirror dependency), dirty indicator, Ctrl/Cmd+S, disabled Save when clean, conflict bar with *Reload* / *Overwrite anyway*, confirm-on-close with unsaved changes. *New file* button in the toolbar and per-folder; choosing an existing name opens that file instead of blanking it.

## Tests
- Backend: 12 new (roundtrip, conflict incl. deleted-file, binary/oversized rejection, symlink protection, HTTP surface incl. traversal + auth). Full suite: 471 passed.
- Frontend: 8 editor tests + 4 ProjectPage integration tests. Full suite: 185 passed, build OK.

## Next

### DIFF
--- a/dashboard/chat/project_files.py
+++ b/dashboard/chat/project_files.py
@@ -13,6 +13,7 @@ project root.
 """
 import errno
 import os
+import stat as stat_mod
 import shutil
 import logging
 import tempfile
@@ -27,6 +28,9 @@ MAX_NAME_LENGTH = 255
 # Path depth cap: keeps recursive tree walks and rmtree far away from
 # Python's recursion limit no matter what gets created through the API.
 MAX_PATH_DEPTH = 32
+# Editor cap — the in-browser editor is for notes/configs/snippets, not
+# for logs; a textarea with more than this is unusable anyway.
+MAX_TEXT_EDIT_BYTES = 2 * 1024 * 1024  # 2 MB
 
 
 def configure_max_content_length(app):
@@ -218,7 +222,6 @@ def stat_file(root: str, rel_path: str) -> str:
     FIFOs/sockets/devices are rejected rather than handed to send_file —
     opening a FIFO would block the worker indefinitely). Returns the
     absolute path (for send_file)."""
-    import stat as stat_mod
     abs_path = resolve(root, rel_path)
     try:
         st = os.lstat(abs_path)
@@ -277,6 +280,79 @@ def save_stream(root: str, dir_rel: str, filename: str, stream, overwrite: bool 
             if os.path.exists(tmp_path):
                 os.unlink(tmp_path)
     return _node(abs_target, rel_target)
+
+
+def read_text(root: str, rel_path: str) -> dict:
+    """Read a regular file as UTF-8 text for the editor. Rejects
+    non-regular files (via stat_file), oversized files, and binary
+    content (NUL byte or invalid UTF-8)."""
+    abs_path = stat_file(root, rel_path)
+    st = os.lstat(abs_path)
+    if st.st_size > MAX_TEXT_EDIT_BYTES:
+        raise ProjectFilesError("file too large to edit", status=413)
+    with open(abs_path, "rb") as f:
+        data = f.read()
+    if b"\x00" in data:
+        raise ProjectFilesError("not a text file", status=400)
+    try:
+        content = data.decode("utf-8")
+    except UnicodeDecodeError:
+        raise ProjectFilesError("not a text file (not valid UTF-8)", status=400)
+    return {
+        "path": "/".join(split_rel_path(rel_path)),
+        "content": content,
+        "size": st.st_size,
+        "modified_at": int(st.st_mtime),
+    }
+
+
+def write_text(root: str, rel_path: str, content: str,
+               base_modified_at: int = None) -> dict:
+    """Write UTF-8 text to a file (create or overwrite), atomically via a
+    temp file in the same directory. The target's parent must exist and
+    the target itself must be absent or a regular file — symlinks and
+    special files are never written through or replaced.
+
+    base_modified_at (optional) is the mtime the client loaded: if the
+    file changed on disk since, the save is rejected with 409 so an edit
+    from another tab/editor isn't silently overwritten.
+    """
+    parts = split_rel_path(rel_path)
+    if not parts:
+        raise ProjectFilesError("path is required")
+    if not isinstance(content, str):
+        raise ProjectFilesError("content must be a string")
+    try:
+        encoded = content.encode("utf-8")
+    except UnicodeEncodeError:
+        raise ProjectFilesError("content is not valid unicode")
+    if len(encoded) > MAX_TEXT_EDIT_BYTES:
+        raise ProjectFilesError("content too large", status=413)
+
+    abs_path = resolve(root, rel_path)
+    abs_dir = os.path.dirname(abs_path)
+    if not os.path.isdir(abs_dir):
+        raise ProjectFilesError("parent directory not found", status=404)
+    if os.path.lexists(abs_path):
+        st = os.lstat(abs_path)
+        if not stat_mod.S_ISREG(st.st_mode):
+            raise ProjectFilesError("not a regular file", status=409)
+        if base_modified_at is not None and int(st.st_mtime) != int(base_modified_at):
+            raise ProjectFilesError("file changed on disk since it was loaded", status=409)
+    elif base_modified_at is not None:
+        # Client edited a file that has since been deleted/renamed.
+        raise ProjectFilesError("file changed on disk since it was loaded", status=409)
+
+    with _fs_errors():
+        fd, tmp_path = tempfile.mkstemp(prefix=".edit-", suffix=".tmp", dir=abs_dir)
+        try:
+            with os.fdopen(fd, "wb") as f:
+                f.write(encoded)
+            os.replace(tmp_path, abs_path)
+        finally:
+            if os.path.exists(tmp_path):
+                os.unlink(tmp_path)
+    return _node(abs_path, "/".join(parts))
 
 
 def mkdir(root: str, rel_path: str) -> dict:
--- a/dashboard/chat/project_files.py
+++ b/dashboard/chat/project_files.py
@@ -317,7 +317,7 @@ def read_text(root: str, rel_path: str) -> dict:
 
 
 def write_text(root: str, rel_path: str, content: str,
-               base_revision: str = None) -> dict:
+               base_revision: str = None, create_only: bool = False) -> dict:
     """Write UTF-8 text to a file (create or overwrite), atomically via a
     temp file in the same directory. The target's parent must exist and
     the target itself must be absent or a regular file — symlinks and
@@ -327,6 +327,11 @@ def write_text(root: str, rel_path: str, content: str,
     (read_text's ``revision``, a content hash): if the file's bytes
     changed on disk since, the save is rejected with 409 so an edit from
     another tab/editor isn't silently overwritten.
+
+    create_only=True is the new-file precondition: the save is rejected
+    with 409 if the path already exists (the client believed it was
+    creating a file, but its tree snapshot was stale). Publication is an
+    atomic no-replace link, mirroring save_stream's upload contract.
     """
     parts = split_rel_path(rel_path)
     if not parts:
@@ -345,6 +350,8 @@ def write_text(root: str, rel_path: str, content: str,
     if not os.path.isdir(abs_dir):
         raise ProjectFilesError("parent directory not found", status=404)
     if os.path.lexists(abs_path):
+        if create_only:
+            raise ProjectFilesError("file already exists", status=409)
         st = os.lstat(abs_path)
         if not stat_mod.S_ISREG(st.st_mode):
             raise ProjectFilesError("not a regular file", status=409)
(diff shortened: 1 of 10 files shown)


teo-mateo:

- **[P1] Do not truncate the optimistic-concurrency token to whole seconds** (`dashboard/chat/project_files.py:305`, `dashboard/chat/project_files.py:340`). Both the value returned to the editor and the comparison on save use `int(st.st_mtime)`. On filesystems with subsecond mtimes, an initial write and a subsequent edit in the same second therefore expose the same token. I reproduced two tabs loading the original token, the first tab saving, and the second stale tab then saving successfully with that same token; the second save silently replaced the first instead of returning 409. This directly defeats the conflict protection during ordinary quick edits. Please use a lossless opaque revision such as `st_mtime_ns` encoded as a string or a content/version hash, and add a test where the on-disk content changes while the whole-second mtime remains equal.

teo-mateo:

**Finding 1.1 — whole-second mtime defeats the conflict guard**: accepted, and went one step further than the suggested fix.

Fixed in 6b86c07. The token is now a content hash (sha256 prefix), not a timestamp at any precision. The suggested `st_mtime_ns` was implemented first and rejected by its own test: Linux stamps writes landing within one kernel timer tick with IDENTICAL nanosecond mtimes, so quick successive saves (exactly the scenario in the finding) still collided. The content hash only matches when the bytes match. `read_text` returns `revision`, `PUT` takes `base_revision` (string), and a file grown past the editor cap since load counts as changed. The regression test forces the absolute worst case — content changed while mtime is byte-for-byte identical (seconds AND nanoseconds, via os.utime restore) — and still gets the 409.

Suites: 472 backend / 185 frontend passed.

teo-mateo:

1. **[P1] A successful save can mark newer, unsaved textarea content as clean** (`dashboard/frontend/src/components/chat/ProjectFileEditor.jsx:43-49`). The textarea remains editable while the request is in flight. If the user clicks Save for content A, types content B before that request resolves, and then the A response arrives, the unconditional `setDirty(false)` clears the modified indicator even though only A is on disk. Save becomes disabled and closing no longer asks for confirmation, so B is silently lost. Capture the submitted content/revision and clear dirty only if the current textarea still equals that submitted snapshot (while still advancing the base revision to the server response). Add a deferred-save frontend test that edits again before resolution and verifies the editor remains dirty and can save B.

2. **[P2] Creating a file from a stale tree can silently overwrite an existing file** (`dashboard/frontend/src/components/chat/ProjectPage.jsx:239-246`, `dashboard/chat/project_files.py:347-369`). `isNew` is decided only from the currently loaded tree snapshot. Because the filesystem is the source of truth, a file added by an external editor or another dashboard tab after that snapshot is absent from `findNode`; the blank editor then saves with no `base_revision`, and `write_text` skips every existence/content check and `os.replace` destroys the existing file. This does not require millisecond race timing: the page can remain stale indefinitely. Give new-file saves an explicit expected-absent/create-only precondition that returns 409 if the path already exists, distinct from the explicit Overwrite anyway operation, and cover the stale-tree case at both API and editor integration levels.

teo-mateo:

**Finding 2.1 — mid-save typing marked clean**: accepted. Fixed in the pushed commit: save captures the submitted snapshot and clears dirty only if the textarea still equals it (latest content tracked via ref); the base revision still advances from the server response so the follow-up save of the newer content is correctly guarded. Regression test defers the save, types content B mid-flight, and verifies the editor stays dirty and saves B against the new revision.

**Finding 2.2 — stale tree lets New file silently overwrite**: accepted. Fixed in the pushed commit: new-file saves send `create_only: true` (mutually exclusive with `base_revision`, boolean-validated), enforced in `write_text` with a pre-check plus an atomic no-replace `link()` publication — same contract as uploads. The editor surfaces the 409 as an "already exists" conflict with Reload (which loads the on-disk file and permanently drops create mode) or Overwrite anyway (explicit forced save). Covered at the pure layer (incl. dangling-symlink occupancy), HTTP (stale create → 409, original content intact), and editor integration (create-only flag, reload-drops-create-mode, forced overwrite).

Suites: 474 backend / 189 frontend passed.

teo-mateo:

1. **[P1] The first successful save of a new file can still discard text typed while that save is in flight** (dashboard/frontend/src/components/chat/ProjectPage.jsx:342-346, dashboard/frontend/src/components/chat/ProjectFileEditor.jsx:45-47). Suppose a new file submits content A, then the user types B before the request resolves. The editor correctly leaves B dirty at line 69, but onSaved changes the parent prop from isNew=true to false. That prop change reruns the load effect, fetches the just-saved A, and lines 33-37 replace B and mark the editor clean. This is the same ordinary workflow that the iteration-2 fix protects for existing files, but the new-file transition bypasses it and silently loses B. Do not reload merely because isNew flips after a successful create (the editor already has the returned revision and has dropped create mode), and add a ProjectPage-level deferred first-save test that types B before A resolves and asserts B remains dirty/saveable.

2. **[P2] Editing an existing file silently resets its filesystem permissions to 0600** (dashboard/chat/project_files.py:350-368). mkstemp creates the staging file as 0600, and os.replace installs that inode without copying the target mode. I reproduced editing a 0755 shell script through read_text/write_text and its mode became 0600, removing the executable bit; a normal 0644 text file likewise becomes 0600. Preserve the existing regular file mode on the staged inode before publication (while defining the intended mode for newly created files), and add a regression assertion that an edit leaves permission bits intact.

teo-mateo:

**Finding 3.1 — first-save isNew flip reloads and clobbers mid-flight typing**: accepted. Fixed in b2da844: the editor's initial-load decision now comes from its internal create-mode ref instead of the isNew prop, so the parent flipping the prop after a successful create no longer re-runs load(). ProjectPage-level regression test does exactly the described sequence (create A pending → type B → resolve) and asserts B is intact, dirty, no content fetch happened, and B then saves as a normal guarded write.

**Finding 3.2 — edits reset permissions to 0600**: accepted, and applied one level wider than flagged: `save_stream` (overwriting uploads) had the identical defect. Both paths now chmod the staged inode before publication — preserving the existing regular file's mode bits, with 0644 as the defined mode for newly created files. Regression covers new-file default, 0755 preserved across an edit AND an overwriting upload.

Suites: 475 backend / 190 frontend passed.

teo-mateo:

1. **[P1] Normal navigation discards dirty editor content without confirmation** (dashboard/frontend/src/components/chat/ProjectFileEditor.jsx:97-100, dashboard/frontend/src/components/chat/ProjectPage.jsx:153-160, dashboard/frontend/src/components/chat/ChatPage.jsx:167-173). The discard guard exists only in the editor X-button handler. If the user edits a file and then clicks any conversation or another project in the always-visible sidebar, the route changes and ProjectFileEditor is unmounted (or ProjectPage clears editing on a project-id change) without calling handleClose, so the unsaved text is silently lost. Reloading/closing the browser has the same gap. This is an ordinary navigation path, not a timing race. Propagate the dirty state to a route/unload guard so every way of leaving the editor asks before discarding, and add a router-level test that edits a file, attempts sidebar navigation, and verifies cancel keeps both the route and text intact.

2. **[P2] The write endpoint can create a file that the editor immediately refuses to read** (dashboard/chat/project_files.py:303-318, dashboard/chat/project_files.py:345-355). read_text defines a NUL byte as binary and returns 400, but write_text only UTF-8-encodes and size-checks content; U+0000 is valid UTF-8, so it is accepted. I reproduced write_text(root, "nul.txt", "before\\x00after") succeeding, followed by read_text of the same file failing with 400 "not a text file". The PUT surface therefore violates the stated text-only contract and can save content that cannot be reopened in this editor. Reject NUL-containing content before staging it, and cover both the layer and PUT/GET round trip.

teo-mateo:

**Finding 4.1 — navigation silently discards dirty editor content**: accepted. Fixed in the pushed commit. The editor surfaces dirty state via `onDirtyChange` (reset on unmount), ProjectPage threads it to ChatPage, and every ChatPage navigation entry point — conversation select, open-another-project, new conversation (plain and in-project) — runs a confirm-before-discard guard. Browser close/reload is covered by a `beforeunload` handler armed only while dirty. Router-level tests: dirty + cancel keeps the route and the typed text; confirm navigates and unmounts; clean navigation never asks. Editor-level tests cover the beforeunload arm/disarm cycle and the dirty-report lifecycle. (Not covered: the app-level left nav to non-chat pages — that leaves the chat SPA section entirely; beforeunload-equivalent coverage there would need an app-wide router blocker, which I'll take as a follow-up if it bites in practice.)

**Finding 4.2 — PUT accepts NUL content that GET then refuses**: accepted. Fixed in the pushed commit: `write_text` rejects NUL-containing content with 400 before staging. Covered at the layer (nothing left on disk) and over HTTP.

Suites: 477 backend / 195 frontend passed.

teo-mateo:

1. **[P1] Starting another new file silently discards unsaved content in the current editor** (`dashboard/frontend/src/components/chat/ProjectPage.jsx:239-246`, `:292-309`). The project toolbar remains visible while `ProjectFileEditor` is open, but `handleNewFile` calls `setEditing(...)` directly and never consults the dirty state or the editor close guard. A normal sequence—edit `a.md`, click **New file**, enter `b.md`—immediately remounts the keyed editor for `b.md`; the unsaved `a.md` text is gone without confirmation. The iteration-4 guard only protects navigation handled in `ChatPage`, so it does not cover this in-page editor replacement. Route this action through a confirm-before-discard guard (or have the editor own the requested file switch), and add a ProjectPage integration test asserting cancel keeps the original path and text while confirm opens the new file.

teo-mateo:

**Finding 5.1 — toolbar New file replaces a dirty editor without asking**: accepted. Fixed in the pushed commit. ProjectPage now mirrors the editor's dirty state (same onDirtyChange stream that feeds ChatPage) and `handleNewFile` runs a confirm-before-discard guard before prompting for the name. Integration tests: cancel keeps the open path and unsaved text (and never shows the name prompt), confirm opens the new empty editor, clean editor never asks.

Suites: 477 backend / 198 frontend passed.

teo-mateo:

1. **[P1] A failed initial load leaves an unconditional blank editor that can overwrite the unread file** (`dashboard/frontend/src/components/chat/ProjectFileEditor.jsx:38-41`, `:64-71`, `:164-172`). When the content GET rejects a binary or oversized file (or hits a transient error), `finally` sets `loading` false and the render path exposes an empty textarea. Typing into it enables Save, but `baseRevision` is still null and create mode is false, so the PUT becomes an unconditional overwrite of the existing regular file. A user who clicks a binary file, sees the load error, then types and saves can therefore replace the original bytes without any conflict or explicit overwrite action. The current binary-load test asserts only that the error appears and misses that the writable textarea also appears. Keep the editor non-writable/save-disabled until the initial load succeeds (with Retry/Close as appropriate), and add a regression test that a failed initial load cannot invoke a save.

teo-mateo:

**Finding 6.1 — failed load exposes a blank editor that can overwrite the unread file**: accepted. Fixed in the pushed commit: a failed initial load renders the error plus a Retry action instead of the textarea, so there is nothing to type into and Save stays disabled. Regression tests: binary-file load shows the error with NO textarea and a disabled Save; a transient failure recovers via Retry into a normal editor.

Suites: 477 backend / 199 frontend passed.

teo-mateo:

1. **[P1] Browser Back/Forward still discards dirty editor content without confirmation** (`dashboard/frontend/src/components/chat/ChatPage.jsx:113-121`, `:181-190`). The new guard is called only by ChatPage's explicit sidebar/create handlers, while the editor's `beforeunload` listener covers only document unloads. A BrowserRouter Back/Forward navigation is a same-document history transition, so it invokes neither path. For example, visit a conversation, open a project file, edit it, then press Back: React Router switches to the conversation route and unmounts the editor immediately, losing the unsaved text without prompting. The same bypass applies when history moves between two project URLs. Add a router-level navigation blocker (or equivalent history-transition guard) tied to the dirty state, and add a MemoryRouter regression with two history entries that dirties the editor, attempts `navigate(-1)`, and verifies cancel preserves both the route and textarea content.

teo-mateo:

**Finding 7.1 — Back/Forward bypasses the dirty guard**: acknowledged as real, deferred out of scope → tracked in #80.

The clean fix is a router-level blocker (`useBlocker`), which react-router only provides on data routers — this app mounts `<BrowserRouter>` + `<Routes>`, so the fix means migrating the entire app's routing setup, a structural change touching every page for one residual exit path. The popstate-sentinel alternative (re-pushing history state on cancel) is fragile and can desync router state. Given that every explicit exit (sidebar, create actions, in-page New file, editor close, browser unload) is already guarded and this is a single-user tool, the cost/benefit says follow-up, not this PR. Issue #80 captures the migration scope and the requested MemoryRouter regression.

Closing the loop: 7 iterations, 11 findings — 10 fixed with regression tests, this one deferred with a tracked issue. Findings arrived one per round and never repeated. Merging per the convergence precedent from PR #78.

PR #81, opened 2026-07-15
feat(chat): project-files MCP server scoped to the conversation's project

teo-mateo:
Phase 4 of the projects feature (#77, #78, #79 shipped phases 1–3): the model can now reach the project's file area through a built-in MCP server.

## What's in here

**New built-in server `project-files`** (`dashboard/chat/mcp_servers/project_files_server.py`) with three read-only tools:
- `list_files(path="")` — file tree with relative paths and sizes
- `read_file(path)` — UTF-8 text files up to 2 MB
- `search_files(query, path="")` — case-insensitive substring over file names and text contents, capped at 200 matches with an explicit truncation notice

All path handling is delegated to `chat/project_files.py` (the same layer behind the HTTP file routes), so `..` traversal, absolute paths, symlink escapes, binary files and oversized files are rejected identically to the UI.

**Per-conversation scoping via spawn environment.** Built-in MCP servers are stdio subprocesses spawned per call from a static registry config, so the project scope travels as `LLM_DOCK_PROJECT_ROOT` in the subprocess environment:
- `_open_streams` passes `config["env"]` merged over the SDK's `get_default_environment()`
- `MCPClientManager.call_tool` grew an `extra_env` parameter (per-call, never mutates the shared config)
- `ProjectScopedMCPManager` (new `chat/project_files_mcp.py`) wraps the process-wide manager per run, injecting the env only for `project-files` calls; built in routes where the conversation is known

The model never sees or supplies a project id. Unscoped calls (server manually enabled on a non-project conversation) get a clear "not part of a project" answer.

**Auto-enable for project conversations.** `runtime._build_stream` appends `project-files` to the enabled servers whenever `conv.project_id` is set — membership in a project is the toggle. This drives both tool discovery and the `tool_hint` in the system prompt. `routes._mcp_for_conversation` builds a manager for project conversations even when no servers are toggled (previously `mcp_manager` was only created when the toggle list was non-empty).

## Tests

39 new tests in `dashboard/tests/test_project_files_mcp.py`:
- direct tool-function coverage: listing (subtree, missing root, empty, traversal), reading (binary/symlink/missing rejection), searching (line numbers, name matches, subdir scope, binary skip, truncation cap)
- glue: env injection targets only `project-files`, `with_project_files` dedup, runtime auto-enable, hint in system prompt, routes helper scoping
- end-to-end: real server subprocess spawned through `MCPClientManager` with injected env — proves the stdio env plumbing against the installed MCP SDK

Full backend suite: 516 passed.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -166,7 +166,7 @@ Rules enforced at load time:
 - `id` must not contain `__` (collides with `server_id__tool_name`
   namespacing in `mcp_client.py`).
 - `id` must not collide with a built-in (`sympy-math`, `schemdraw-circuits`,
-  `render-html`). Built-ins always win.
+  `render-html`, `project-files`). Built-ins always win.
 - `name`, `description`, `icon`, `tool_hint` are always required.
 - `enabled` defaults to `true`.
 - `transport` defaults to `"stdio"`. Accepted values: `"stdio"`, `"http"`.
--- a/dashboard/chat/db.py
+++ b/dashboard/chat/db.py
@@ -322,6 +322,31 @@ class ChatDB:
         finally:
             self._close_conn(conn)
 
+    def resolve_project_id(self, conv_id: str) -> Optional[str]:
+        """Effective project of a conversation: its own project_id, or the
+        nearest ancestor's. Project membership is root-only — spin-off rows
+        keep project_id NULL and follow their root ancestor — so this is
+        what tool scoping must use, not the row's own column. The visited
+        set guards against a parent cycle (impossible through the API, but
+        a walk must never be the thing that hangs on corrupt data)."""
+        seen = set()
+        conn = self._get_conn()
+        try:
+            while conv_id and conv_id not in seen:
+                seen.add(conv_id)
+                row = conn.execute(
+                    "SELECT project_id, parent_conversation_id FROM conversations WHERE id = ?",
+                    (conv_id,),
+                ).fetchone()
+                if row is None:
+                    return None
+                if row["project_id"]:
+                    return row["project_id"]
+                conv_id = row["parent_conversation_id"]
+            return None
+        finally:
+            self._close_conn(conn)
+
     def list_conversations(self, limit: int = 50, offset: int = 0) -> Tuple[List[Conversation], int]:
         """List conversations, newest-updated first.
 
--- a/dashboard/chat/mcp_client.py
+++ b/dashboard/chat/mcp_client.py
@@ -85,19 +85,13 @@ class MCPClientManager:
         asyncio.set_event_loop(self._loop)
         self._loop.run_forever()
 
-    def _run_async(self, coro):
-        """Run an async coroutine from sync code."""
-        future = asyncio.run_coroutine_threadsafe(coro, self._loop)
-        return future.result(timeout=30)
+    def _run_async(self, coro, timeout: float = 30):
+        """Run an async coroutine from sync code.
 
-    def run_with_timeout(self, coro, timeout: float):
-        """Run a coroutine on the manager's event loop with a custom timeout.
-
-        Exposed for one-shot admin operations (the Tools page test endpoint)
-        that want a tighter deadline than the 30s default and don't want to
-        touch the cached chat path. On timeout, cancel the future so the
-        running coroutine (and its `stdio_client` subprocess) gets a chance
-        to clean up instead of leaking past the HTTP response.
+        On timeout (or any waiter-side error), cancel the future so the
+        running coroutine — and the `stdio_client` subprocess it holds —
+        gets torn down instead of grinding on after the caller already
+        returned an error (e.g. a search still scanning a huge project).
         """
         future = asyncio.run_coroutine_threadsafe(coro, self._loop)
         try:
@@ -105,6 +99,15 @@ class MCPClientManager:
         except Exception:
             future.cancel()
             raise
+
+    def run_with_timeout(self, coro, timeout: float):
+        """Run a coroutine on the manager's event loop with a custom timeout.
+
+        Exposed for one-shot admin operations (the Tools page test endpoint)
+        that want a tighter deadline than the 30s default and don't want to
+        touch the cached chat path.
+        """
+        return self._run_async(coro, timeout=timeout)
 
     def get_tools(self, server_id: str) -> list:
         """Get tools for a server in OpenAI format. Uses cache."""
--- a/dashboard/chat/mcp_client.py
+++ b/dashboard/chat/mcp_client.py
@@ -7,7 +7,7 @@ import logging
 import threading
 from typing import Optional
 
-from mcp.client.stdio import StdioServerParameters, stdio_client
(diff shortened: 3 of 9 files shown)


teo-mateo:

1. **[P2] Cap `list_files` output before it enters the model context** (`dashboard/chat/mcp_servers/project_files_server.py:65`). The function materializes the complete recursive tree, flattens it into another list, and returns every entry. Project storage has no aggregate file-count limit, so a legitimate large project can produce an arbitrarily large MCP response; that response is appended to `messages_array` for the next model request and can exhaust memory or exceed the model context, failing the chat turn. Add an entry and/or serialized-byte cap with an explicit truncation notice, plus a boundary test analogous to the search cap.

2. **[P2] Bound search work and cancel the coroutine when the 30-second call times out** (`dashboard/chat/mcp_servers/project_files_server.py:115`, `dashboard/chat/mcp_client.py:88`). `MAX_SEARCH_RESULTS` limits matches, but a no-match query still walks the entire tree and reads every text file up to 2 MB. On a large but valid project this can exceed `_run_async`’s 30-second timeout. Unlike `run_with_timeout`, `_run_async` does not cancel its future, so the stdio server keeps scanning after the caller has received an error; repeated searches can leave multiple obsolete subprocesses consuming disk/CPU. Give `search_files` a total files/bytes/time scan budget (and report truncation) and cancel `_run_async`’s future on timeout; add a test that exercises the scan/timeout boundary.

3. **[P3] Re-check the result cap after adding a filename match** (`dashboard/chat/mcp_servers/project_files_server.py:126`). If 199 content matches have already been collected and the next file has both a matching name and matching content, the name append reaches 200, then the content loop appends a 201st result before checking the cap. I reproduced 201 result lines plus the “truncated at 200” notice. Check immediately after the name-match append (or centralize capped appends), and extend `test_truncation_cap` to cover the 199 + name + content boundary.

The MCP SDK env contract and per-call config copy/merge look correct against the installed SDK, and I did not find a project-root escape in the new tools.

teo-mateo:

**Finding 1.1 — cap list_files output**: accepted. Fixed in 7884211. `MAX_LIST_ENTRIES = 500`; over that, the listing is cut and ends with `(truncated at 500 of N entries — pass a subdirectory path to list less)`. Regression test `test_entry_cap_truncates_with_notice`.

**Finding 1.2 — bound search work + cancel on timeout**: accepted, both parts. Fixed in 7884211.
- `search_files` now has a 64 MB total read budget (`MAX_SEARCH_SCAN_BYTES`). Files over the 2 MB text cap are skipped *before* budget accounting (read_text would reject them anyway, so they cost nothing). When the budget is spent the scan stops with `(search stopped early: 64 MB scan budget reached — narrow the search with a subdirectory path)`, appended even on the no-matches path. Tests: `test_scan_budget_stops_early_with_notice`, `test_scan_budget_notice_even_with_no_matches`, `test_oversized_files_do_not_consume_budget`.
- `_run_async` now cancels its future on timeout (same contract `run_with_timeout` already had; that method now delegates to `_run_async`), so a timed-out tool call tears down the stdio subprocess instead of letting it keep scanning. Test: `test_timeout_cancels_the_coroutine` verifies the coroutine actually receives `CancelledError`.

**Finding 1.3 — cap overshoot after a name match**: accepted. Fixed in 7884211. The cap is re-checked immediately after the name-match append, so a name match landing exactly on the cap breaks out before the same file's content matches can overshoot. Regression test `test_cap_enforced_after_name_match` covers the (cap−1 content) + name + content boundary at cap=2.

Full backend suite: 522 passed.

teo-mateo:

1. **[P2] Give `read_file` an MCP-specific output limit** (`dashboard/chat/mcp_servers/project_files_server.py:105`). `pf.read_text` accepts any UTF-8 file up to the editor cap of 2 MiB, and this tool returns that content verbatim. `stream_with_tools` then appends the entire result to `messages_array` for the next completion (`dashboard/chat/tool_loop.py:99`), with no truncation or token budget. A legitimate near-limit project file can therefore add hundreds of thousands of tokens, exceed the configured model context, and fail the chat turn. Cap the serialized tool result well below the editor limit with an explicit truncation notice, and preferably support an offset/line-range argument so the model can fetch the remainder; add a boundary test showing that a file near `MAX_TEXT_EDIT_BYTES` cannot produce an unbounded tool response.

teo-mateo:

**Finding 2.1 — cap read_file output**: accepted. Fixed in the commit above. `read_file` now returns at most `MAX_READ_CHARS = 64k` characters per call and grew an `offset` parameter; a longer file ends with `(truncated: showing characters A–B of N — call read_file again with offset=B to continue)`, and the final chunk carries no notice. Offset is validated: negative and boolean rejected, past-end reported with the file's length. Boundary test `test_near_cap_file_response_is_bounded` writes a file near `MAX_TEXT_EDIT_BYTES` and asserts the response stays bounded by the chunk cap; `test_long_file_is_chunked_with_continuation_offset` covers the chunk/continue round trip.

Full backend suite: 526 passed.

teo-mateo:

1. **[P2] Resolve inherited project membership for spin-off conversations** (`dashboard/chat/routes.py:82`, `dashboard/chat/runtime.py:128`). Project membership is intentionally root-only: a spin-off row keeps `project_id = NULL` and the existing UI/API contract says it follows its root ancestor (see `test_spinoff_cannot_be_created_with_project` and the sidebar inherited-placement logic). Both new gates inspect only `conv.project_id`, however. Consequently, sending a message in a spin-off whose root belongs to a project makes `_mcp_for_conversation` return no scoped manager (unless some unrelated MCP server is toggled), and `_build_stream` does not auto-enable `project-files` or add its hint. If it was manually toggled, it is still invoked unscoped and only returns “not part of a project.” Resolve the effective project from the root ancestor before constructing the manager and effective server list, and add an integration test for a project root plus child spin-off that asserts the child gets the root project directory and tools.

teo-mateo:

**Finding 3.1 — spin-offs don't inherit project scoping**: accepted. Fixed in the commit above. New `ChatDB.resolve_project_id(conv_id)` walks the parent chain to the nearest ancestor with a project (visited-set guarded so corrupt data can't hang the walk). `routes._mcp_for_conversation` and `runtime._build_stream` both resolve through it when the row's own `project_id` is NULL and a parent exists, so a spin-off of a project conversation now gets the scoped manager, the auto-enabled server and the tool hint — pointed at the root's project directory. The runner's db-less path (ghost chat) keeps working on direct membership only.

Tests: `TestSpinoffInheritance` covers root/child/grandchild/unfiled/missing resolution, cycle termination, the routes helper scoping a spin-off to `storage/p1`, runtime auto-enable for a nested spin-off, and the db-less runner. Full backend suite: 531 passed.

teo-mateo:

Reviewed the current head (`1d7e5c8`) against the full diff and all prior comments. The MCP SDK stdio environment contract and the per-call config/env merge are correct, project-root containment remains enforced by the shared project-files layer, and the iteration 1–3 fixes cover the substantive edge cases I found.

Verification: `cd dashboard && venv/bin/python -m pytest tests -q` — 531 passed.

PR #82, opened 2026-07-15
refactor(projects): arch hygiene — single project resolution, root-only trigger, stable error codes

teo-mateo:
Implements recommendations 1, 2 and 4 from the codex architecture review of the projects feature (follow-up to #77/#78/#79/#81).

## Rec 1 — resolve effective project context once per run

Routes and the runtime each resolved a conversation's project independently — two owners of one policy, with a theoretical divergence window (root moved between the two resolutions → MCP manager scoped to one project, tools auto-enabled for another).

Now `routes._effective_project_id()` is the **single owner**: called once at run creation, snapshotted into `ChatTurnRequest.effective_project_id`, threaded through `run_manager.start()/_execute()`, and consumed verbatim by `ChatRunner._build_stream()` — which no longer resolves anything itself (including no fallback to `conv.project_id`, so the single-owner property is real, and there's a regression test asserting the runner does not resolve on its own). This also makes the snapshot semantics explicit: membership is fixed when the run is created.

## Rec 2 — DB-level enforcement of root-only membership

Two new triggers reject any conversation row carrying both `parent_conversation_id` and `project_id` (insert, and update of either column). The HTTP routes already rejected this; now the invariant that `resolve_project_id()` and the sidebar rely on is self-defending against any future write path. `SET project_id = NULL` (detach-on-project-delete) is unaffected.

## Rec 4 — stable machine-readable error codes

`ProjectFilesError` grew a `code` field, surfaced as `{"error": msg, "code": ...}` by the blueprint error handler. Coded conditions are the ones the UI branches on: `revision_conflict` (editor conflict bar) and `already_exists` (create-only conflict, upload overwrite prompt, mkdir/move conflicts). `fetchAPI` and the raw-fetch upload helper attach `err.code` to thrown errors; `ProjectFileEditor` and `ProjectPage` now branch on `err.code` instead of matching human-readable message text, so wording can change freely.

## Tests

- `test_spinoff_cannot_carry_project_directly` — trigger enforcement (insert + both update directions + detach still allowed)
- `test_effective_project_id_resolves_through_the_chain`, `test_runtime_auto_enables_for_spinoff_snapshot`, `test_runner_does_not_resolve_on_its_own`, `test_run_passes_the_snapshot_through` — snapshot threading
- `TestErrorCodes` (6 tests) — every coded 409 on the HTTP surface, plus uncoded errors carrying no `code` key
- Frontend mocks updated to carry codes like the real API

Backend: 540 passed. Frontend: 199 passed.

### DIFF
--- a/dashboard/chat/db.py
+++ b/dashboard/chat/db.py
@@ -109,6 +109,24 @@ WHEN NEW.project_id IS NOT NULL
 BEGIN
     SELECT RAISE(ABORT, 'project not found');
 END;
+
+-- Root-only membership: spin-offs follow their root ancestor and must never
+-- carry their own project_id (resolve_project_id and the sidebar both rely
+-- on this). The HTTP routes reject it too; these make the model
+-- self-defending against any future write path that forgets the rule.
+CREATE TRIGGER IF NOT EXISTS trg_conversations_rootonly_insert
+BEFORE INSERT ON conversations
+WHEN NEW.parent_conversation_id IS NOT NULL AND NEW.project_id IS NOT NULL
+BEGIN
+    SELECT RAISE(ABORT, 'spin-off conversations cannot carry a project');
+END;
+
+CREATE TRIGGER IF NOT EXISTS trg_conversations_rootonly_update
+BEFORE UPDATE OF project_id, parent_conversation_id ON conversations
+WHEN NEW.parent_conversation_id IS NOT NULL AND NEW.project_id IS NOT NULL
+BEGIN
+    SELECT RAISE(ABORT, 'spin-off conversations cannot carry a project');
+END;
 """
 
 
--- a/dashboard/chat/project_files.py
+++ b/dashboard/chat/project_files.py
@@ -48,11 +48,18 @@ def configure_max_content_length(app):
 
 
 class ProjectFilesError(Exception):
-    """Client-facing error with an HTTP status."""
+    """Client-facing error with an HTTP status.
 
-    def __init__(self, message: str, status: int = 400):
+    `code` is a stable machine-readable identifier for conditions the
+    frontend must branch on (conflict bars, overwrite prompts) — clients
+    match on it, never on the human-readable message, so wording can
+    change freely. Only conditions with a UI behavior get a code.
+    """
+
+    def __init__(self, message: str, status: int = 400, code: str = None):
         super().__init__(message)
         self.status = status
+        self.code = code
 
 
 def default_storage_root() -> str:
@@ -248,7 +255,7 @@ def save_stream(root: str, dir_rel: str, filename: str, stream, overwrite: bool 
     if os.path.isdir(abs_target):
         raise ProjectFilesError("a directory with that name exists", status=409)
     if os.path.exists(abs_target) and not overwrite:
-        raise ProjectFilesError("file already exists", status=409)
+        raise ProjectFilesError("file already exists", status=409, code="already_exists")
     # Same permission fidelity as write_text: mkstemp stages at 0600, so an
     # overwriting upload must keep the target's mode and a fresh one gets
     # a normal default instead of 0600.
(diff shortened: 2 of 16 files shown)


teo-mateo:

1. **[P2] Preserve the DB-less Ghost Chat project-membership path** (`dashboard/chat/runtime.py:131-140`, `dashboard/tests/test_project_files_mcp.py:353-367`). PR #81 explicitly shipped and tested the `ChatRunner(db=None, ...)` / ephemeral “Ghost Chat” shape: a conversation carrying a direct `project_id` still auto-enabled `project-files`. This PR removes that regression test and makes the new request field default to `None`, while only Flask routes populate it. An ephemeral caller that continues to construct `ChatTurnRequest(conversation=conv, mcp_manager=mgr)` now silently loses the project-files tool and its prompt hint even when `conv.project_id` is set. Keep the single-resolution rule by snapshotting the direct project id in the Ghost Chat run-creation path (and building the correspondingly scoped manager there), rather than restoring runner-side resolution, and retain a DB-less regression test through `ChatRunner.run`.

2. **[P3] The snapshot-threading test does not exercise the threading it claims to protect** (`dashboard/tests/test_project_files_mcp.py:362-367`). `test_run_passes_the_snapshot_through` only constructs a `ChatTurnRequest` and reads its field; it would still pass if `ChatRunManager._execute` stopped forwarding `effective_project_id`, which is the central invariant of this refactor. Drive `ChatRunManager.start` with a spy runner and assert the received request contains the snapshot (or assert the value reaching `_build_stream`) so this path can actually regress safely.

3. **[P3] Error-code transport is untested at both frontend adapters** (`dashboard/frontend/src/api.js:29-45`, `dashboard/frontend/src/services/chat.js:83-96`). The component tests manufacture errors with `err.code`, while the backend tests only assert the JSON body. Therefore removing or breaking the new JSON-to-`Error.code` plumbing in either `fetchAPI` or `uploadProjectFile` would leave all 540 backend and 199 frontend tests green while the editor conflict bar / upload overwrite prompt stops working. Add adapter-level fetch-response tests for coded and uncoded errors.

Verification: `cd dashboard && venv/bin/python -m pytest tests -q` — 540 passed; `cd dashboard/frontend && npm test -- --run` — 199 passed.

teo-mateo:

**Finding 1.1 — DB-less Ghost Chat project-membership path**: pushed back on the behavioral change, accepted the test. The premise overstates what shipped: `ChatTurnRequest` has exactly one production constructor (`run_manager._execute`, fed by routes) — Ghost Chat (#57) exists only as the `NullPersistencePolicy` seam; there is no ephemeral run-creation path in the codebase to preserve. The removed test asserted runner-side resolution, which is precisely the second policy owner this PR (per arch review rec 1) exists to eliminate — restoring a `conv.project_id` fallback in the runner would reintroduce the divergence the refactor removes. The contract for a future ephemeral caller is documented on `ChatTurnRequest.effective_project_id`: the run-creation site owns resolution and passes the snapshot. What I did take: a DB-less regression test **through `ChatRunner.run()`** (`test_dbless_run_consumes_the_snapshot`, db=None + NullPersistencePolicy) pinning that contract, in commit 9e22a60.

**Finding 1.2 — snapshot test doesn't exercise the threading**: accepted. Fixed in 9e22a60. `test_run_manager_threads_snapshot_into_the_request` drives `ChatRunManager.start()` through the real executor with a spy runner and asserts the arriving request carries `effective_project_id="p9"` and the same conversation object — dropping the forwarding in `_execute` now fails a test.

**Finding 1.3 — error-code transport untested at the adapters**: accepted. Fixed in 9e22a60. New `api.test.js` (fetchAPI: coded body, uncoded body, non-JSON body, success) and `chat.upload.test.js` (raw-fetch uploadProjectFile: coded 409, uncoded 413, success) — breaking the JSON→`Error.code` plumbing in either adapter now fails the frontend suite.

Backend 541, frontend 206 passing.

teo-mateo:

I found no additional correctness, regression, test-coverage, or API-contract issues after accounting for the iteration 1 fixes and the author's valid pushback.

Verification: backend 541 passed; frontend 206 passed. The pre-existing untracked chat-mockup.html and design/ items were left untouched, and the repository remains on feat/projects-arch-hygiene.

PR #83, opened 2026-07-15
feat(projects): explorer-style file view with drag-and-drop and context menu

teo-mateo:
Reworks the project files page into a two-pane explorer and adds the backend copy operation that copy/paste needs.

## Frontend
- **Two-pane layout**: folder tree on the left (expand/collapse independent of selection), the selected folder's contents on the right, with breadcrumb navigation. Selection follows a moved folder to its new path and falls back to the nearest surviving ancestor when the selected folder is deleted.
- **Drag-and-drop move**: rows (files and folders) are draggable; tree folders, folder rows, and the right pane's background are drop targets. A custom DataTransfer type keeps foreign drags inert; dropping into the item's own parent is a no-op; dropping a folder into its own subtree is rejected client-side.
- **Context menu** (new `ContextMenu.jsx`): Open/Download/Cut/Copy/Rename/Delete on entries; New file/New folder/Upload/Paste on folders, the root tree node, and the pane background. A toolbar chip shows the armed clipboard entry (cut sources render dimmed); pasting a copy auto-suffixes `(copy)` / `(copy N)` on name collisions. Cut+paste maps to move; the clipboard clears when its source is moved, renamed, or deleted.
- Toolbar (upload / new file / new folder) now targets the selected folder.
- All existing stale-response/stale-mutation/generation-theft guards and editor regressions preserved and re-covered.

## Backend
- `project_files.copy_path`: copies a regular file (staged + atomic no-replace link, preserving mode), a dangling symlink (as a link, never followed), or a directory tree (`copytree(symlinks=True)` with cleanup of partial copies). Same contracts as `move`: destination parent must exist, `409 already_exists` when the destination is taken, no copy-into-self. A pre-scan rejects special files (a FIFO would block the worker) and enforces `MAX_PATH_DEPTH` at the destination.
- `POST /api/chat/projects/<id>/files/copy` `{path, new_path}` → 201.

## Tests
- Backend: 559 passing (+16 for copy: pure layer incl. symlink/FIFO/depth/traversal cases, HTTP roundtrip, conflict codes, auth).
- Frontend: 229 passing (ProjectPage suite rewritten for the explorer model: tree navigation, breadcrumbs, context menu, clipboard flows, DnD incl. guard cases).

### DIFF
--- a/dashboard/chat/project_files.py
+++ b/dashboard/chat/project_files.py
@@ -457,6 +457,86 @@ def move(root: str, rel_src: str, rel_dst: str) -> dict:
     return _node(abs_dst, "/".join(dst_parts))
 
 
+def _precheck_copy_tree(abs_src: str, dst_depth: int):
+    """Pre-scan a directory before copytree: reject special files (opening
+    a FIFO to copy it would block the worker forever) and enforce the path
+    depth cap at the DESTINATION — the source obeys it relative to its own
+    location, but copying deep subtree into an already-deep folder could
+    mint paths build_tree/rmtree would then have to recurse past."""
+    for dirpath, dirnames, filenames in os.walk(abs_src):
+        rel_depth = 0 if dirpath == abs_src else len(os.path.relpath(dirpath, abs_src).split(os.sep))
+        child_depth = dst_depth + rel_depth + 1
+        if child_depth > MAX_PATH_DEPTH and (dirnames or filenames):
+            raise ProjectFilesError("copy would exceed maximum path depth")
+        for name in filenames:
+            st = os.lstat(os.path.join(dirpath, name))
+            if not (stat_mod.S_ISREG(st.st_mode) or stat_mod.S_ISLNK(st.st_mode)):
+                raise ProjectFilesError("directory contains a special file that cannot be copied")
+
+
+def copy_path(root: str, rel_src: str, rel_dst: str) -> dict:
+    """Copy a file, symlink, or directory tree to a new relative path.
+    Same contract as move: the destination's parent must exist and the
+    destination itself must not. Symlinks are copied as links, never
+    followed (matching the tree/delete contract — a link pointing outside
+    the root must not smuggle its target's content into the project)."""
+    src_parts = split_rel_path(rel_src)
+    dst_parts = split_rel_path(rel_dst)
+    if not src_parts:
+        raise ProjectFilesError("cannot copy project root")
+    if not dst_parts:
+        raise ProjectFilesError("destination is required")
+    abs_src = resolve(root, rel_src)
+    abs_dst = resolve(root, rel_dst)
+    if not os.path.lexists(abs_src):
+        raise ProjectFilesError("source not found", status=404)
+    if os.path.lexists(abs_dst):
+        raise ProjectFilesError("destination already exists", status=409, code="already_exists")
+    if dst_parts[:len(src_parts)] == src_parts:
+        raise ProjectFilesError("cannot copy a directory into itself")
+    abs_dst_dir = os.path.dirname(abs_dst)
+    if not os.path.isdir(abs_dst_dir):
+        raise ProjectFilesError("destination directory not found", status=404)
+
+    src_st = os.lstat(abs_src)
+    with _fs_errors():
+        if stat_mod.S_ISDIR(src_st.st_mode):
+            _precheck_copy_tree(abs_src, len(dst_parts))
+            try:
+                shutil.copytree(abs_src, abs_dst, symlinks=True)
+            except FileExistsError:
+                raise ProjectFilesError("destination already exists", status=409, code="already_exists")
+            except Exception:
+                # A partial tree is worse than no tree: the client will
+                # retry the whole copy, and a half-populated destination
+                # would then 409 as already_exists.
+                shutil.rmtree(abs_dst, ignore_errors=True)
+                raise
+        elif stat_mod.S_ISLNK(src_st.st_mode):
+            try:
+                os.symlink(os.readlink(abs_src), abs_dst)
+            except FileExistsError:
+                raise ProjectFilesError("destination already exists", status=409, code="already_exists")
+        elif stat_mod.S_ISREG(src_st.st_mode):
+            # Stage in the destination directory, then publish with an
+            # atomic no-replace link — same 409 contract as save_stream
+            # even against a concurrent claim of the destination name.
+            fd, tmp_path = tempfile.mkstemp(prefix=".copy-", suffix=".tmp", dir=abs_dst_dir)
+            try:
+                os.close(fd)
+                shutil.copy2(abs_src, tmp_path)  # preserves mode
+                try:
+                    os.link(tmp_path, abs_dst)
+                except FileExistsError:
+                    raise ProjectFilesError("destination already exists", status=409, code="already_exists")
+            finally:
+                if os.path.exists(tmp_path):
+                    os.unlink(tmp_path)
+        else:
+            raise ProjectFilesError("not a copyable file type")
+    return _node(abs_dst, "/".join(dst_parts))
+
+
 def delete(root: str, rel_path: str):
     parts = split_rel_path(rel_path)
     if not parts:
--- a/dashboard/chat/project_files.py
+++ b/dashboard/chat/project_files.py
@@ -501,6 +501,16 @@ def copy_path(root: str, rel_src: str, rel_dst: str) -> dict:
     src_st = os.lstat(abs_src)
     with _fs_errors():
         if stat_mod.S_ISDIR(src_st.st_mode):
+            # The component-prefix check above only sees CLIENT paths; an
+            # in-root symlink alias (alias -> d) would let "d" ->
+            # "alias/copy" physically target d/copy and send copytree
+            # into its own output. Compare physical paths: the real
+            # destination is its existing parent's realpath plus the
+            # final component.
+            real_src = os.path.realpath(abs_src)
+            real_dst = os.path.join(os.path.realpath(abs_dst_dir), os.path.basename(abs_dst))
+            if real_dst == real_src or real_dst.startswith(real_src + os.sep):
+                raise ProjectFilesError("cannot copy a directory into itself")
             _precheck_copy_tree(abs_src, len(dst_parts))
             try:
                 shutil.copytree(abs_src, abs_dst, symlinks=True)
--- a/dashboard/chat/project_files.py
+++ b/dashboard/chat/project_files.py
@@ -499,19 +499,23 @@ def copy_path(root: str, rel_src: str, rel_dst: str) -> dict:
         raise ProjectFilesError("destination directory not found", status=404)
 
     src_st = os.lstat(abs_src)
+    # The component checks above only see CLIENT paths; an in-root symlink
+    # alias can spell a destination whose PHYSICAL location differs — both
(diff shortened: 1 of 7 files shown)


teo-mateo:

- **Selected directories do not follow a successful rename** (`dashboard/frontend/src/components/chat/ProjectPage.jsx:311`). `handleRename` refreshes the tree and clears the clipboard, but unlike `moveInto` it never remaps `selectedDir`. If the user selects `docs/inner` and renames `docs` to `notes` from the left-tree context menu (or renames the selected folder itself), the refreshed tree no longer contains the old selected path, so the fallback effect sends the right pane to the nearest old ancestor/root instead of following `notes/inner`. This contradicts the PR's advertised “selection follows renames/moves” behavior. After a successful rename, remap `selectedDir` from `node.path` to the normalized new path using the same descendant-suffix logic as `moveInto`, and add a test for renaming the selected folder or one of its ancestors.

teo-mateo:

**Finding 1.1 — Selected directories do not follow a successful rename**: accepted.

Fixed in commit 8cc4f84. `handleRename` now remaps `selectedDir` with the same descendant-suffix logic as `moveInto`, after normalizing the prompted path the way the backend's `split_rel_path` does (leading/trailing slashes stripped), so the remap works with the path the refreshed tree will actually report. Failed renames leave the selection untouched (gated on `run()`'s success return). Three regression tests added: renaming an ancestor of the selected folder, renaming the selected folder itself (with slash normalization), and a failed rename keeping the selection in place.

teo-mateo:

- **Folder mutations can strand the open editor on an obsolete path** (`dashboard/frontend/src/components/chat/ProjectPage.jsx:308`, `dashboard/frontend/src/components/chat/ProjectPage.jsx:326`, `dashboard/frontend/src/components/chat/ProjectPage.jsx:365`). The new two-pane layout leaves the folder tree actionable while `ProjectFileEditor` is mounted. If `docs/a.txt` is open, renaming or dragging `docs`, cut-pasting it elsewhere, or deleting it all proceeds without consulting `confirmDiscardEdits`; the success handlers remap `selectedDir` and/or clear the clipboard but leave `editing.path` as `docs/a.txt`. The buffer therefore remains on screen but its next save targets a path that no longer exists (and force-save cannot recover it when the parent moved or was deleted). This is especially problematic for dirty edits because the existing navigation guard does not cover these mutations. Before mutating a source that is the open file or one of its ancestors, either preserve the buffer while remapping the editor path, or require discard confirmation and close the editor; cancellation must abort the filesystem mutation. Add regression tests for rename/move/delete of an ancestor while the editor is dirty and for the successful path not leaving a stale editor.

teo-mateo:

**Finding 2.1 — Folder mutations can strand the open editor on an obsolete path**: accepted.

Fixed in commit 1b97d5c. New `guardEditedPath(srcPath)` runs before rename, move (both drag-and-drop and cut-paste go through `moveInto`), and delete whenever the mutation source is the open file or one of its ancestors: a dirty buffer prompts for discard and cancel aborts the filesystem mutation entirely. On success, `remapEditing` carries `editing.path` to the destination with the same descendant-suffix logic as the selection remap — the keyed editor remount then reloads the buffer from the new path (clean buffers land in an equivalent state; dirty ones were explicitly discard-confirmed). Deleting the open file's subtree closes the editor. Six regression tests cover: dirty-drag cancel aborts, confirmed drag remaps, clean rename remaps without prompting, delete closes the editor, discard-cancel aborts the delete, and cut-paste of the open file's folder remapping.

teo-mateo:

- **Renaming or moving an ancestor can hide the selected folder in a collapsed tree branch** (`dashboard/frontend/src/components/chat/ProjectPage.jsx:340`, `dashboard/frontend/src/components/chat/ProjectPage.jsx:401`). The new selection remaps update `selectedDir`, but `expanded` still contains only the old source paths. For example, select `docs/inner` (which expands `docs`) and rename `docs` to `notes`: the right pane follows `notes/inner`, but the left tree renders `notes` collapsed because it still has `docs` in the expansion set, so there is no visible selected tree row. Moving an ancestor under a different parent has the same result (and stale expansion entries can affect a later folder recreated at the old path). Remap source/descendant entries in `expanded` to the destination on successful rename/move, or explicitly expand all ancestors of the remapped selection, and add a regression assertion that the selected tree node remains visible after an ancestor rename/move.

teo-mateo:

**Finding 3.1 — Renaming or moving an ancestor can hide the selected folder in a collapsed tree branch**: accepted.

Fixed in the commit just pushed. New `remapExpanded(srcPath, dstPath)` runs alongside the existing selection/editor remaps on successful rename and move: every `expanded` entry that is the source or one of its descendants is rewritten to the destination path (which also removes the stale old-path entries that could pre-expand a folder later recreated at the old path), and all ancestors of the destination are added so the moved/renamed node is actually on screen. Regression coverage: the ancestor-rename test now also asserts the selected tree row (`tree-node-notes/inner`) is rendered, and a new ancestor-move test asserts `tree-node-archive/docs/inner` is visible after dragging `docs` into `archive` while `docs/inner` is selected.

teo-mateo:

- **Copy-into-self can bypass the guard through an internal symlink alias** (`dashboard/chat/project_files.py:489-506`). The rejection compares only the client path components. If the project contains directory `d` and symlink `alias -> d`, `copy_path(root, "d", "alias/copy")` passes the prefix check and writes the destination physically at `d/copy`; I reproduced this on the PR branch. More deeply placed aliases can also make `copytree` encounter the destination while walking the source. This violates the API contract that directory copies cannot target their own subtree and can produce recursive/partial copies. For directory sources, compare the resolved physical destination (including the real path of its existing parent) against the physical source before copying, and add a regression test using an in-root symlink alias.

- **A mutation that finishes after switching projects can mutate the new project UI state** (`dashboard/frontend/src/components/chat/ProjectPage.jsx:260-270`, `dashboard/frontend/src/components/chat/ProjectPage.jsx:349-359`, `dashboard/frontend/src/components/chat/ProjectPage.jsx:412-424`). `run` correctly suppresses stale busy/error commits, but it still returns `true` after `refresh()` declines a stale-project request. The awaiting rename/move/delete/paste handler then performs its remap, editor close, clipboard clear, or expansion against the component now rendering project B. For example, start moving `docs` in A, switch to B and select/edit a path under `docs`, then let the A request resolve: the old handler remaps the B selection/editor to the A destination. Make mutation success project-bound (for example, have `run` return false unless `projectIdRef.current === pid` after the awaited work) and cover a delayed move/rename completion across a project switch, including the post-success state rather than only the tree fetch.

teo-mateo:

**Finding 4.1 — Copy-into-self can bypass the guard through an internal symlink alias**: accepted.

Fixed in commit 8de8dab. For directory sources, `copy_path` now compares physical paths before copying: the real destination (realpath of its existing parent joined with the final component) against the source's realpath. `copy_path(root, "d", "alias/copy")` with `alias -> d` is rejected with the same "cannot copy a directory into itself" 400, and nothing is written. Regression tests cover the alias-into-self rejection and, to keep the guard from over-rejecting, a copy through an alias that lands *outside* the source still succeeding.

**Finding 4.2 — A mutation that finishes after switching projects can mutate the new project UI state**: accepted.

Fixed in commit 8de8dab. `run()` now returns `projectIdRef.current === pid` on the success path, so every post-success handler (selection/editor/expansion remaps, editor close on delete, clipboard clears, `expandTo`) is skipped when the component switched projects while the mutation was in flight; `handleNewFolder`'s `expandTo` was also brought under the gate. Regression test: A's `docs → archive/docs` move resolves after switching to project B with B's own `docs/inner` selected — B's selection stays on `docs/inner`, neither remapped to A's destination nor knocked to the root by the missing-path fallback.

teo-mateo:

- **The destination-depth precheck can be bypassed through an in-root symlink alias** (`dashboard/chat/project_files.py:460-470`, `dashboard/chat/project_files.py:510-516`). `_precheck_copy_tree` receives `len(dst_parts)`, which is the depth of the client spelling rather than the physical destination established by `real_dst`. I reproduced this by creating a 31-component directory, pointing root-level `alias` at it, and copying a directory containing `x` to `alias/c`: the copy succeeds and creates the file at physical depth 33 despite `MAX_PATH_DEPTH == 32`. `build_tree` then returns that deep file, but download/edit/delete reject its returned path as `path too deep`, so the copy leaves an entry the UI cannot operate on. Derive the destination depth from `real_dst` relative to `realpath(root)` (and reject it if necessary) before the pre-scan, and add a regression test for a deep in-root alias rather than only the current shallow alias cases.

teo-mateo:

**Finding 5.1 — The destination-depth precheck can be bypassed through an in-root symlink alias**: accepted.

Fixed in the commit just pushed. `copy_path` now derives the destination depth physically — `realpath(abs_dst_dir)` joined with the final component, taken relative to `realpath(root)` — and rejects destinations past `MAX_PATH_DEPTH` before any branch, so single-file and symlink copies are covered as well as directory children (which flow into `_precheck_copy_tree` with the physical depth). Regression tests: a directory copy through a root-level alias to a 31-deep directory is rejected with nothing written, and a single-file copy whose physical destination sits past the cap is rejected too.

teo-mateo:

- **Edits made while a folder mutation is in flight can still be lost without confirmation** (`dashboard/frontend/src/components/chat/ProjectPage.jsx:167-178`, `dashboard/frontend/src/components/chat/ProjectPage.jsx:350-354`, `dashboard/frontend/src/components/chat/ProjectPage.jsx:413-417`). The iteration-2 guard samples `editorDirtyRef` only before awaiting `run()`, while the editor remains mounted and writable throughout the filesystem request and the follow-up full-tree refresh. If `docs/a.txt` is open and clean, the user can start renaming or moving `docs`, type into the textarea before the deferred request resolves, and then the success handler changes the editor key and reloads from disk, silently discarding those new edits; delete similarly closes the now-dirty editor. The same window exists after accepting the initial discard prompt. Lock/close the affected editor once the mutation is committed, or preserve its live buffer while remapping, and add a regression test that defers `mockMove`/`mockDelete`, types after the API call starts, then resolves it and verifies the text is not silently lost.

teo-mateo:

**Finding 6.1 — Edits made while a folder mutation is in flight can still be lost without confirmation**: accepted.

Fixed in the commit just pushed, taking the "lock/close once committed" option one step further: the affected editor is closed *before* the request starts (after the discard guard and, for delete, the delete confirmation), so no writable buffer exists during the await at all — the pre-await dirty sample cannot be raced by typing. On success the editor reopens at the remapped destination path (freshly loaded from disk); on failure it reopens at its original path; both restorations are gated on the component still showing the project the mutation ran against (composing with iteration 4.2's project binding — a mid-flight project switch restores nothing). Delete stays closed on success. Regression tests defer `mockMove`/`mockDelete` and assert: no textarea exists while the move is in flight and the editor reopens at the new path after resolution; a failed move reopens the editor at the original path; a deferred delete closes the editor before resolution and stays closed after.

teo-mateo:
## Codex review iteration 7 — No further findings

I found no remaining correctness, security, API-contract, production-performance, or test-gap issue that meets the iteration 7 single-user data-loss/corruption/broken-UI threshold.

Validated the final PR branch with:
- `pytest -q tests/test_project_files.py` — 98 passed
- `vitest run src/components/chat/ProjectPage.test.jsx` — 60 passed
- targeted ESLint for the changed explorer files — passed
- `vite build` — passed

PR #84, opened 2026-07-15
feat(chat): project-files MCP write tools — create, write, edit, insert

teo-mateo:
Extends the `project-files` MCP server from read-only (list/read/search) to read-write, so the chat LLM can create project files and modify them — including inserts and partial replacements.

## New tools
- **create_file(path, content="")** — strictly create; refuses to overwrite (same `already_exists` contract as the HTTP routes); auto-creates missing parent folders and the lazily-created project root.
- **write_file(path, content)** — replaces an existing regular file's entire content; a missing file is an error that points the model at create_file.
- **edit_file(path, old_text, new_text, replace_all=false)** — exact-snippet replacement. Zero matches → clear error telling the model to read_file and copy verbatim; multiple matches without replace_all → error reporting the count. The write carries the read's content revision, so a concurrent save from the in-browser editor fails loudly ("changed on disk") instead of silently reverting the user's edit.
- **insert_text(path, line, text)** — inserts whole lines after a 1-based line number (0 = top of file); appending after an unterminated final line adds the missing newline instead of gluing lines together.

## Plumbing
- Everything funnels through `chat.project_files.write_text`: path traversal/symlink-escape rejection, atomic temp-file publication, UTF-8 + NUL validation, and the 2 MB text cap all apply identically to the LLM and the UI.
- Scoping unchanged: tools act only under `LLM_DOCK_PROJECT_ROOT` injected per call; unscoped conversations get the explicit refusal.
- Registry `tool_hint` rewritten to describe the write surface and steer the model toward edit_file for small changes and actually persisting files when asked to.

## Tests
- 599 backend passing (+35: per-tool direct coverage incl. traversal/NUL/size-cap/ambiguity/revision-conflict/glue edge cases, unscoped refusals for all four, an end-to-end create→edit→insert round trip through the real spawned subprocess, and tool discovery asserting all seven tools).

### DIFF
--- a/dashboard/chat/mcp_registry.py
+++ b/dashboard/chat/mcp_registry.py
@@ -128,10 +128,10 @@ The diagram will be rendered automatically as an artifact.""",
     },
     "project-files": {
         "name": "Project Files",
-        "description": "Read this conversation's project files — list, read, search. Auto-enabled for conversations inside a project; does nothing outside one.",
+        "description": "Read and edit this conversation's project files — list, read, search, create, write. Auto-enabled for conversations inside a project; does nothing outside one.",
         "command": [sys.executable, os.path.join(_SERVERS_DIR, "project_files_server.py")],
         "icon": "fa-folder-tree",
-        "tool_hint": "This conversation belongs to a project with a file area (documents, notes, data the user keeps alongside the project's conversations). You have read-only access to it: list_files shows the file tree, read_file returns a text file's content, search_files finds a substring across file names and contents. Paths are relative to the project root, e.g. 'docs/plan.md' — use them exactly as list_files prints them. When the user refers to project files or material \"in the project\", consult these tools instead of guessing.",
+        "tool_hint": "This conversation belongs to a project with a file area (documents, notes, data the user keeps alongside the project's conversations). Reading: list_files shows the file tree, read_file returns a text file's content, search_files finds a substring across file names and contents. Writing: create_file makes a new text file (parent folders are created automatically; it refuses to overwrite), write_file replaces an existing file's entire content, edit_file swaps one exact snippet for another (copy the snippet verbatim from read_file — for small changes prefer this over write_file), insert_text adds lines after a given line number. All files are UTF-8 text up to 2 MB; paths are relative to the project root, e.g. 'docs/plan.md' — use them exactly as list_files prints them. When the user refers to project files or material \"in the project\", consult these tools instead of guessing, and when they ask you to save or update something in the project, actually write the file rather than only replying in chat.",
     },
     "render-html": {
         "name": "Render HTML",
--- a/dashboard/chat/mcp_servers/project_files_server.py
+++ b/dashboard/chat/mcp_servers/project_files_server.py
@@ -1,10 +1,18 @@
-"""Project Files MCP server — read-only view of one project's file area.
+"""Project Files MCP server — read/write access to one project's file area.
 
 Spawned per tool call by MCPClientManager like the other built-ins, but
 scoped: the dashboard injects LLM_DOCK_PROJECT_ROOT (the absolute
 directory of the conversation's project) into the subprocess environment
 at spawn time (see chat/project_files_mcp.py). Without it, every tool
 answers with a clear error instead of guessing a directory.
+
+Read tools: list_files, read_file, search_files. Write tools:
+create_file, write_file (full replace), edit_file (exact-snippet
+replace), insert_text (line-based insert). Writes go through
+chat.project_files.write_text — atomic temp-file publication, UTF-8/NUL
+validation, the 2 MB text cap, and (for the read-modify-write tools) the
+content-revision guard, so a concurrent editor save turns into a clear
+error instead of a silent overwrite.
 
 All path handling is delegated to chat.project_files — the same layer the
 HTTP file routes use — so `..` traversal, absolute paths and symlink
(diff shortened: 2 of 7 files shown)


teo-mateo:

- **High — the revision check still has a lost-update race** (`dashboard/chat/mcp_servers/project_files_server.py:307`, `:341`). `edit_file` and `insert_text` pass the revision read earlier into `write_text`, but `write_text` checks that hash before it stages the temp file and calls `os.replace`. If the in-browser editor saves after that comparison and before the replace, the MCP write silently overwrites the newer browser content instead of returning the promised “changed on disk” error. This is relevant even for a single local user because the Flask editor and the spawned MCP process can write concurrently. The new test only injects a revision that is already stale when `write_text` starts, so it cannot catch this window. Please coordinate the compare-and-publish section with a cross-process per-path lock shared by the HTTP and MCP writers (or another true CAS mechanism), and add a regression test that pauses after the comparison, performs the browser-side save, then resumes publication.

teo-mateo:

**Finding 1.1 — the revision check still has a lost-update race**: accepted.

Fixed in the commit just pushed. `write_text` now takes an exclusive `flock` on the project root directory around the whole precondition-check + publication section, making the base_revision comparison and `os.replace` a true CAS. Both writer populations funnel through `write_text` — the Flask editor's PUT route and the spawned MCP tool processes — so the lock is genuinely cross-process (and flock contends between separate opens within one process too, covering threaded servers). Per-project granularity: contention is one user's editor save vs. their chat turn, so coarseness costs nothing.

Regression test as requested (`TestWriteTextCAS`): writer A is paused via a hooked `mkstemp` exactly between its revision check and its publication; writer B (same base revision) is started and asserted to **block** rather than enter the critical section; A is then resumed and B is asserted to fail with `revision_conflict` while A's content survives on disk. Without the lock, B completes inside the pause window and the test fails on the `tb.is_alive()` assertion — it reproduces the reported race.

teo-mateo:

- **High — the CAS lock is still bypassed by several dashboard file mutations** (`dashboard/chat/project_files.py:246-300`, `:463-594`, `:401-438`). `write_text` now serializes only other `write_text` callers. An overwrite upload still calls `os.replace` without taking `_write_lock`, and move/delete likewise mutate the target without it. Therefore, while `edit_file` or `insert_text` is paused after its revision comparison, the user can overwrite-upload, move, or delete that file in the explorer; the MCP call then resumes and silently replaces the upload or recreates the moved/deleted path. This is the same lost-update window the new lock is meant to close, and it is realistic for the stated single-user model because the UI and spawned tool process are the two actors. Please make every project-file mutator that can affect a checked path participate in the same project lock (with unlocked internal helpers if needed), and extend the regression test so a non-`write_text` mutation is attempted during the compare/publish pause and is shown to block.

- **Medium — `write_file` can create a file despite its replace-only contract** (`dashboard/chat/mcp_servers/project_files_server.py:268-269`). The existence check happens in `stat_file` before `write_text` acquires the lock, while `write_text` without `base_revision` intentionally creates an absent target. If the file is deleted or moved between those calls, `write_file` succeeds and recreates it instead of returning the documented “file not found” error. Add a must-exist/replace-only precondition to `write_text` and evaluate it inside the locked critical section, rather than prechecking in the MCP wrapper.

teo-mateo:

**Finding 2.1 — the CAS lock is still bypassed by several dashboard file mutations**: accepted.

Fixed in the commit just pushed. `_write_lock` is now taken by every mutator: `mkdir`, `move`, `copy_path`, and `delete` run their precondition-check + mutation sections inside it, and `save_stream` publishes under it while keeping the body streaming OUTSIDE the lock (a slow multi-MB upload must not stall the editor or a chat turn; only its final chmod+replace/link synchronizes, with the directory-collision re-check repeated under the lock). Regression tests extend `TestWriteTextCAS` exactly as requested: with a CAS writer paused between its revision check and its publication, an overwrite upload and a delete are each shown to block (`is_alive()` after a grace period) and then apply after the CAS publication — the later operation wins, nothing is silently undone.

**Finding 2.2 — write_file can create a file despite its replace-only contract**: accepted.

Fixed in the same commit. `write_text` gains a `must_exist=True` precondition — 404 when the target is absent, mutually exclusive with `create_only` — evaluated inside the locked critical section alongside the other preconditions. The MCP `write_file` passes it instead of pre-checking with `stat_file`, so a delete or move landing between the wrapper and the write can no longer make `write_file` recreate the path (an existing-but-non-regular target still fails with the established "not a regular file" 409). Covered by `TestWriteTextMustExist`.

teo-mateo:

- **Medium — a late `create_file` can resurrect a deleted project and leave an orphan directory** (`dashboard/chat/mcp_servers/project_files_server.py:242-246`, `dashboard/chat/routes.py:301-306`). Project membership is snapshotted for the chat run, so a tool call from that run can still carry the old root after the user deletes the project. The delete route performs one initial `rmtree`, deletes the DB row, then performs one cleanup sweep; if `create_file` reaches its unconditional `os.makedirs(root, exist_ok=True)` after that sweep, it recreates the root and successfully writes files even though the project no longer exists. The new root-directory `flock` cannot prevent this because the directory/inode used as the lock has already been removed, and the MCP subprocess never revalidates project existence. I reproduced this by completing both deletion passes and then invoking a still-scoped `create_file`; it returned success and recreated `<storage>/<project_id>/late.md`. Please coordinate project deletion and MCP writes through lifecycle state that survives removal of the project root (for example a stable parent-level lock plus a deletion tombstone/existence check), and add a regression test where a scoped call pauses until after the delete sweep and is then rejected without recreating the directory.

teo-mateo:

**Finding 3.1 — a late create_file can resurrect a deleted project and leave an orphan directory**: accepted, fixed with the codebase's established pattern rather than a tombstone.

The HTTP mutation routes already solve this exact race with a commit-point check (`_revalidate_or_cleanup`): mutate, then re-check the project row, and sweep the directory if the row vanished — whichever side acts last cleans up. The MCP path now does the same: `ProjectScopedMCPManager` accepts a `revalidate` callback (built in `routes._mcp_for_conversation`, where the DB is known) and invokes it in a `finally` after every project-files tool call — success, error, or timeout, since a dying subprocess may have written first. A late `create_file` still returns "Created" to the model, but the directory is swept before the turn continues; no orphan survives either interleaving, and no lifecycle state or parent-level lock is needed because correctness doesn't depend on the (removable) root inode.

Regression tests (`TestDeletedProjectSweep`): a write landing after both deletion passes leaves no directory behind; a live project's write is kept; the sweep runs on the raise path; and non-project-files servers don't trigger revalidation. The reproduced scenario from the finding — both deletion passes completed, then a still-scoped write — is the first test.

(Also in this commit, orthogonal to the review: all seven tool descriptions gained concrete usage examples, per a user request.)

teo-mateo:

I found no additional correctness, security, API-contract, production-performance, or test-gap issue that meets this iteration's stated bar of real data loss, corruption, or durable broken state for the single-user/two-actor model. The repository virtualenv test suite passes: `609 passed`.

PR #85, opened 2026-07-16
feat(explorer): show files in the tree pane

teo-mateo:
The explorer's left tree pane listed folders only; files were reachable only through the right pane. This PR adds file rows to the tree:

- Root-level files and files inside expanded folders render in the tree, below their sibling folders (backend tree order, dirs first).
- Clicking a tree file opens it in the editor, going through the existing dirty-buffer confirm guard.
- Tree files are drag sources (same DnD payload as right-pane rows) and carry the full file context menu (Open / Download / Cut / Copy / Rename / Delete).
- A cut file is dimmed in the tree, matching the right pane.
- The file currently open in the editor is highlighted in the tree.
- Files are NOT drop targets — drops continue to land on folders only.
- The expand chevron now shows for any folder with children (previously only subfolders counted, so a folder holding only files had no chevron and its contents were invisible in the tree).

Tests: 66 ProjectPage tests (6 new for tree file rows), full frontend suite 249 passing. Four existing tests were switched from text queries to testids because root files now legitimately render in both panes.

### DIFF
--- a/dashboard/frontend/src/components/chat/ProjectPage.jsx
+++ b/dashboard/frontend/src/components/chat/ProjectPage.jsx
@@ -74,10 +74,33 @@ function formatSize(bytes) {
   return `${(bytes / (1024 * 1024)).toFixed(1)} MB`
 }
 
-// One folder row in the left tree pane (recursive over subdirectories).
-function TreeDir({ node, depth, expanded, selectedDir, dropTarget, cutPath,
-                   onToggle, onSelect, onMenu, onDragStart, onDragOver, onDragLeave, onDrop }) {
-  const subdirs = (node.children || []).filter(c => c.type === 'dir')
+// One file row in the left tree pane. Files aren't drop targets (drops go
+// to folders), but they are drag sources and carry the full context menu.
+function TreeFile({ node, depth, editingPath, cutPath, onOpen, onMenu, onDragStart }) {
+  return (
+    <div
+      draggable
+      onClick={() => onOpen(node)}
+      onContextMenu={e => onMenu(e, node)}
+      onDragStart={e => onDragStart(e, node)}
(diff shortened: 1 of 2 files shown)


teo-mateo:

- [P3] Treat opening the already-active tree file as a no-op — dashboard/frontend/src/components/chat/ProjectPage.jsx:84-85 and 425-429. TreeFile now calls handleOpenFile even when node.path already equals editing.path. With a dirty editor this produces a “Discard unsaved changes?” prompt merely by clicking the highlighted current file (or choosing Open from its menu). If the user confirms, setEditing installs the same path/key, so React preserves the existing editor and the promised discard does not happen. That leaves a misleading destructive prompt on a normal selection action. Short-circuit when the requested path is already active, and add a test that dirties a file then clicks/opens that same tree row without prompting or changing the buffer.

teo-mateo:

**Finding 1.1 — opening the already-active tree file raises a misleading discard prompt**: accepted.

Fixed in commit 8b6af03. `handleOpenFile` now short-circuits when `node.path` equals the editor's current path, so clicking the highlighted tree row (or its context-menu Open) is a no-op — no prompt, buffer untouched. Regression test added: dirty the open file, click its own tree row, assert no confirm, no reload, buffer intact. 67 ProjectPage tests passing.

teo-mateo:

I reviewed the current head (8b6af03), the full diff and files, all prior comments and the author response. The iteration 1 issue is fixed, and I found no remaining correctness, security, API-contract, performance, or test-gap issue that would cause user-visible breakage or data loss under the stated single-user concurrency model.

Validation: the two changed files pass ESLint; all 250 frontend tests pass; the production frontend build succeeds. Repository-wide ESLint still reports unrelated pre-existing failures outside this PR.

PR #87, opened 2026-07-17
feat(chat): project file explorer beside project conversations

teo-mateo:
## What

When a conversation belongs to a project, the conversation screen splits in two: a **project file-explorer strip on the left** (20% default width, drag-resizable via the divider, collapsible, both persisted in localStorage) and the **chat on the rest**. Conversations without a project are unchanged.

Also adds a **collapse toggle to the conversations list** (ChatSidebar → thin rail with expand + new-conversation buttons, persisted).

## How

- `projectTree.jsx` — `TreeDir`/`TreeFile` and the path helpers extracted from `ProjectPage` so both explorers share one implementation (DnD/context-menu handlers became optional for the read-mostly pane).
- `ProjectExplorerPane` — the strip: browse/expand the tree, open a file, new file, upload into the root, refresh. Same stale-response generation guard as `ProjectPage`. Re-reads the tree when a chat run completes, since the model may have written files through the project-files tools.
- `ProjectChatSplit` — the split container. Opening a file overlays `ProjectFileEditor` on the chat region (the conversation stays loaded underneath); dirty-editor state feeds ChatPage's existing navigation guard, and switching conversations closes the overlay.
- `ChatSidebar` — collapse/expand with localStorage persistence, following the main `Sidebar` pattern.

## Testing

- 24 new vitest cases (explorer pane, split behavior incl. drag-resize clamping and dirty-file guard, sidebar collapse, ChatPage wiring); full suite green except the 4 pre-existing `api.test.js` failures that also fail on `main`.
- `npm run build` passes; `npm run lint` adds no new errors (one warning of a class already present in the codebase).

### DIFF
--- a/dashboard/frontend/src/components/chat/ChatPage.jsx
+++ b/dashboard/frontend/src/components/chat/ChatPage.jsx
@@ -1,8 +1,9 @@
-import { useEffect, useCallback, useRef } from 'react'
+import { useEffect, useCallback, useRef, useState } from 'react'
 import { useNavigate, useParams } from 'react-router-dom'
 import ChatSidebar from './ChatSidebar'
 import ChatArea from './ChatArea'
 import ProjectPage from './ProjectPage'
+import ProjectChatSplit from './ProjectChatSplit'
 import useConversations from '../../hooks/useConversations'
 import useProjects from '../../hooks/useProjects'
 import useChat from '../../hooks/useChat'
@@ -16,6 +17,11 @@ export default function ChatPage() {
   const { conversationId, projectId } = useParams()
   const convId = conversationId || null
   const navigate = useNavigate()
+
+  // Bumped when a chat run completes — the run may have written project
+  // files via the project-files tools, so the explorer strip re-reads
+  // the tree.
+  const [filesRefreshKey, setFilesRefreshKey] = useState(0)
 
   const { conversations, refresh, create, remove, removeMany, rename, patchConversation } = useConversations()
   const { projects, refresh: refreshProjects, create: createProject, rename: renameProject, remove: removeProject } = useProjects()
@@ -78,10 +84,13 @@ export default function ChatPage() {
     }
   }, [convId, conversation, streaming, sendMessage])
 
-  // Refresh conversation list when streaming completes (for auto-title)
+  // Refresh conversation list when streaming completes (for auto-title),
+  // and re-read the project file tree — the run may have created or
+  // edited files.
   useEffect(() => {
     if (!streaming && conversation) {
       refresh()
+      setFilesRefreshKey(k => k + 1)
     }
   }, [streaming]) // eslint-disable-line react-hooks/exhaustive-deps
 
@@ -92,6 +101,15 @@ export default function ChatPage() {
       refresh()
     }
   }, [runReady]) // eslint-disable-line react-hooks/exhaustive-deps
+
+  // Project of the ACTIVE conversation — drives the file-explorer strip
+  // next to the chat. Gated on the loaded conversation matching the route
+  // so a stale `conversation` (mid-switch) can't show the previous chat's
+  // project files; unknown project ids (deleted elsewhere) drop the strip.
+  const conversationProject =
+    convId && conversation && conversation.id === convId && conversation.project_id
+      ? projects.find(p => p.id === conversation.project_id) || null
+      : null
 
   const handleRename = useCallback(async (id, title) => {
     await rename(id, title)
@@ -248,6 +266,12 @@ export default function ChatPage() {
           onEditorDirtyChange={(d) => { editorDirtyRef.current = d }}
         />
       ) : (
+      <ProjectChatSplit
+        project={conversationProject}
+        conversationId={convId}
+        refreshKey={filesRefreshKey}
+        onEditorDirtyChange={(d) => { editorDirtyRef.current = d }}
+      >
       <ChatArea
         conversation={conversation}
         awaitingConversation={!!convId && (!conversation || conversation.id !== convId)}
@@ -274,6 +298,7 @@ export default function ChatPage() {
         onReloadConversation={loadConversation}
         onRefreshConversations={refresh}
       />
+      </ProjectChatSplit>
       )}
     </div>
   )
--- a/dashboard/frontend/src/components/chat/ChatPage.jsx
+++ b/dashboard/frontend/src/components/chat/ChatPage.jsx
@@ -1,4 +1,4 @@
-import { useEffect, useCallback, useRef, useState } from 'react'
+import { useEffect, useCallback, useRef, useState, useMemo } from 'react'
 import { useNavigate, useParams } from 'react-router-dom'
 import ChatSidebar from './ChatSidebar'
 import ChatArea from './ChatArea'
@@ -102,14 +102,33 @@ export default function ChatPage() {
     }
   }, [runReady]) // eslint-disable-line react-hooks/exhaustive-deps
 
-  // Project of the ACTIVE conversation — drives the file-explorer strip
-  // next to the chat. Gated on the loaded conversation matching the route
-  // so a stale `conversation` (mid-switch) can't show the previous chat's
-  // project files; unknown project ids (deleted elsewhere) drop the strip.
-  const conversationProject =
-    convId && conversation && conversation.id === convId && conversation.project_id
-      ? projects.find(p => p.id === conversation.project_id) || null
-      : null
+  // Effective project of the ACTIVE conversation — drives the file-explorer
+  // strip next to the chat. Spin-offs store project_id = NULL and inherit
+  // membership from their root (the DB triggers enforce this and the server
+  // resolves the chain the same way), so walk up through the loaded list;
+  // the list is also refreshed after moves, so a moved active conversation
+  // switches strips without a conversation reload. Falls back to the loaded
+  // conversation's own field when the list row isn't present (fresh chat not
+  // yet in the paginated list) — gated on the id matching the route so a
+  // stale mid-switch `conversation` can't show the previous chat's files.
+  // Unknown project ids (deleted elsewhere) drop the strip.
+  const effectiveProjectId = useMemo(() => {
+    if (!convId) return null
+    const byId = new Map(conversations.map(c => [c.id, c]))
+    let row = byId.get(convId)
+    if (row) {
+      const seen = new Set()
+      while (row.parent_conversation_id && byId.has(row.parent_conversation_id) && !seen.has(row.id)) {
+        seen.add(row.id)
+        row = byId.get(row.parent_conversation_id)
+      }
+      return row.project_id || null
+    }
+    return conversation && conversation.id === convId ? (conversation.project_id || null) : null
+  }, [convId, conversations, conversation])
+  const conversationProject = effectiveProjectId
+    ? projects.find(p => p.id === effectiveProjectId) || null
+    : null
 
   const handleRename = useCallback(async (id, title) => {
     await rename(id, title)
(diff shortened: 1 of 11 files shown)


teo-mateo:

1. **[P1] Do not key the editor-reset effect on the dirty callback** — `dashboard/frontend/src/components/chat/ProjectChatSplit.jsx:49`. `ChatPage` passes `onEditorDirtyChange` as a fresh inline function on every render (`ChatPage.jsx:273`), so `notifyDirty` also changes on every parent render. Because the reset effect depends on `notifyDirty`, any ChatPage update runs `setEditing(null)` and unmounts the editor without confirmation. This is especially easy to hit while an underlying run is streaming, since each delta/heartbeat rerenders ChatPage, and a dirty buffer is silently lost. Keep the latest callback in a ref or pass a stable callback from ChatPage, and make the reset effect depend only on `conversationId` and `projectId`. Please add a regression test that opens and dirties the editor, rerenders the parent with a new callback identity, and verifies that the editor remains open.

2. **[P2] Resolve effective project membership instead of reading only the row field** — `dashboard/frontend/src/components/chat/ChatPage.jsx:107`. The backend deliberately stores `project_id = null` on every spin-off and resolves its effective project through the ancestor chain (`ChatDB.resolve_project_id` / `_effective_project_id`); the sidebar follows the same root-only model. The new `conversationProject` check therefore omits the explorer for every spin-off of a project conversation, even though that chat receives the project-files tools. The same direct, loaded field also stays stale after `handleMoveMany`: that handler refreshes the list but never reloads the active `useChat` conversation, so an active root moved between projects keeps showing the old project files. Derive the effective project from the refreshed conversation/ancestor list or expose an effective project id in the API, and cover inherited membership plus moving the active conversation.

3. **[P2] Guard destructive active-conversation deletion before discarding the file buffer** — `dashboard/frontend/src/components/chat/ChatPage.jsx:224`. The new split editor reports dirty state to ChatPage, but `handleDelete` and `handleDeleteMany` never call `confirmDiscardEdits`. Deleting the active conversation (or an ancestor whose cascade includes it) succeeds, navigates to `/chat`, and unmounts a dirty project-file editor without the discard prompt used by the other navigation paths. Run the dirty guard before the DELETE whenever `activeWillBeDeleted` is true, for both single and bulk deletion, and test the cancel path so no deletion request is sent.

4. **[P2] The extracted tree module introduces six new lint errors** — `dashboard/frontend/src/components/chat/projectTree.jsx:4`. `npm run lint` now reports `react-refresh/only-export-components` at lines 4, 16, 22, 27, 31, and 37 because the JSX module exports both components and helper functions. These are new errors from this PR, despite the PR description saying lint adds no new errors. Move the non-component helpers to a separate module (or otherwise satisfy the existing lint contract) so the repository lint command does not regress.

teo-mateo:

**Finding 1.1 — editor-reset effect keyed on unstable dirty callback**: accepted.
Fixed in b18cefe. `ProjectChatSplit` now keeps `onEditorDirtyChange` behind a ref (synced in an effect, same pattern as `ProjectFileEditor`), so `notifyDirty` is identity-stable and the reset effect depends only on `conversationId`/`projectId`. Added the requested regression test: open + dirty the editor, rerender the parent three times with fresh callback identities, assert the overlay stays mounted and the buffer is intact.

**Finding 1.2 — effective project membership for spin-offs / stale after move**: accepted.
Fixed in b18cefe. `ChatPage` now computes `effectiveProjectId` by walking the spin-off ancestor chain through the loaded conversation list (mirroring `resolve_project_id`'s root-only model), falling back to the loaded conversation's own field only when the list row is absent. Because the source of truth is the refreshed list, `handleMoveMany` on the active conversation now switches the strip without a conversation reload. Test added for a spin-off inheriting its root's project. One acknowledged degradation: if a spin-off's ancestors fall outside the paginated list, the walk stops early and the strip is omitted — it degrades to hidden, never to the wrong project.

**Finding 1.3 — delete paths bypass the dirty-editor guard**: accepted.
Fixed in b18cefe. Both `handleDelete` and `handleDeleteMany` run `confirmDiscardEdits()` whenever `activeWillBeDeleted` is true, before any request fires; cancel aborts the deletion. Tests cover the cancel path (no DELETE sent), the confirm path, and bulk delete.

**Finding 1.4 — six new react-refresh lint errors in projectTree.jsx**: accepted.
Fixed in b18cefe. Path/tree helpers moved to `projectTreeUtils.js`; `projectTree.jsx` now exports only the two row components. `npm run lint` is back to main's baseline of 6 pre-existing errors (my earlier "no new errors" claim was measured with the untracked file present in both runs — bad baseline, thanks for catching it).

teo-mateo:

1. **[P2] Guard project membership changes before they unmount a dirty split editor** — `dashboard/frontend/src/components/chat/ChatPage.jsx:192-204`. The new editor can stay open while the sidebar remains interactive, but neither `handleMoveMany` nor `handleDeleteProject` calls `confirmDiscardEdits`. If the selected ids contain the active conversation or one of its ancestors, moving that root to another project changes `effectiveProjectId`; the refresh then changes `ProjectChatSplit.project`, and its reset effect drops the dirty buffer. Deleting `effectiveProjectId` has the same silent-loss path and also removes the backing project files. Check whether the mutation affects the active effective project before issuing any request, abort on cancel, and add regression cases for moving an active root/spin-off and deleting its project with a dirty editor.

2. **[P2] Browser history navigation bypasses the dirty-file guard** — `dashboard/frontend/src/components/chat/ProjectChatSplit.jsx:61-68`. The guard is only called by explicit ChatPage handlers. With the production `BrowserRouter`, Back/Forward changes `conversationId` through a client-side POP navigation, so `beforeunload` does not fire; this effect then closes the dirty editor and clears the parent flag without confirmation. Add a route-level navigation guard that also covers POP transitions, and test opening and dirtying a file before navigating back.

3. **[P2] Clamp a restored width against the current container, not only during a drag** — `dashboard/frontend/src/components/chat/ProjectChatSplit.jsx:11-15,130-133`. `readStoredWidth` accepts every value above 160 and the render applies it as a non-shrinking pixel width. A width saved on a large display, or simply a stale large localStorage value, can therefore consume the entire split after the window is narrowed and hide the chat; the 45% clamp never runs unless the user starts another drag. Measure and clamp the restored/current width on mount and container resize, and cover a stored width above 45% in the tests.

4. **[P3] Avoid issuing two full tree scans whenever the pane mounts with a positive refresh key** — `dashboard/frontend/src/components/chat/ProjectExplorerPane.jsx:40-52`. The first effect always calls `refresh()`, and the second also calls it whenever `refreshKey > 0`. Since ChatPage keeps that key positive after the first completed run, reopening the collapsed pane or switching projects starts two GETs; the generation guard discards one response but does not avoid the backend recursive filesystem scan. Combine the initial/key refresh bookkeeping or track the last observed key, and add a test that mounts with `refreshKey={1}` and expects one request.

teo-mateo:

**Finding 2.1 — move/delete-project mutations bypass the dirty-editor guard**: accepted.
Fixed in f3d092a. `handleMoveMany` prompts when the moved root ids include the active conversation's root (`activeRootId` now comes out of the same ancestor-walk memo as `effectiveProjectId`); `handleDeleteProject` prompts when the deleted id equals the active effective project. Cancel aborts before any request fires. Tests: cancel/proceed for both paths, plus a no-prompt case for moving unrelated conversations.

**Finding 2.2 — browser Back/Forward bypasses the dirty-file guard**: pushed back as out of scope for this PR — tracked in #88 instead.
The gap is real but not introduced here: the identical hole exists on main for the ProjectPage editor (PR #79's guard architecture is explicit-handler checks + `beforeunload`, and this PR reuses it). A route-level POP guard needs `useBlocker`, which only works under a data router — the app mounts a plain `<BrowserRouter>` (`src/main.jsx`), so the real fix is a `createBrowserRouter` migration touching every route. That's a cross-cutting change that shouldn't ride along in a feature PR; filed as issue #88 covering both editors.

**Finding 2.3 — restored width not clamped against the current container**: accepted.
Fixed in f3d092a. The split measures the container on mount and on resize (ResizeObserver, window-resize fallback for environments without it) and clamps the rendered width to 45% of the live container; the persisted value stays raw so widening the window restores the user's preference. Test covers a stored 800px width in a 1000px container rendering at 450px.

**Finding 2.4 — double tree scan on mount with a positive refreshKey**: accepted.
Fixed in f3d092a. The pane tracks the last-seen key in a ref and refreshes only on changes; the mount-time fetch is owned solely by the project-change effect. Test mounts with `refreshKey={3}` and asserts exactly one GET, then bumps the key and asserts the second.

teo-mateo:

1. **[P2] Include the active project page in the project-delete dirty guard** — `dashboard/frontend/src/components/chat/ChatPage.jsx:196-207`. The new condition only checks `effectiveProjectId`, which is null on `/chat/project/:projectId`. `ProjectPage` reports its dirty editor through the same `editorDirtyRef`, but deleting the project currently being viewed therefore skips the discard prompt, removes the backing files, refreshes `projects` so the editor unmounts, and navigates away. Guard when `id` matches either the active conversation effective project or the route `projectId`, and add a regression test that dirties the ProjectPage editor and cancels deletion.

2. **[P2] Do not disarm the dirty guard before an awaited mutation succeeds** — `dashboard/frontend/src/components/chat/ChatPage.jsx:157-160`. `confirmDiscardEdits()` immediately sets the ref false, while the newly guarded move/delete handlers then await network requests. If an update or delete rejects, the editor remains mounted with its dirty buffer but the parent now believes it is clean; the next sidebar navigation skips confirmation and silently drops those edits. Keep the ref true until the editor actually unmounts/resets, or restore it on every failure path. Please cover a rejected active-conversation move/delete followed by another in-app navigation.

3. **[P2] Preserve an in-flight upload across explorer collapse/reopen** — `dashboard/frontend/src/components/chat/ProjectChatSplit.jsx:141-166` and `ProjectExplorerPane.jsx:80-105`. Collapsing conditionally unmounts `ProjectExplorerPane`, but its upload continues in the old instance. If the user expands before that request finishes, the new pane performs its initial tree read before the upload commits; when the upload later succeeds, its follow-up `refresh()` updates only the unmounted instance, so the visible tree never shows the file. A failure is likewise written to unmounted state and never surfaced. Keep the pane/upload state mounted while collapsed, hoist the operation lifecycle, or prevent collapse while busy, and test with a deferred upload resolved after collapse plus re-expand.

teo-mateo:

**Finding 3.1 — project-delete guard misses the open project page**: accepted.
Fixed in 68d9033. `handleDeleteProject` now prompts when the deleted id matches either the active conversation's effective project or the route `projectId`. Regression test dirties the ProjectPage editor, cancels (no DELETE, buffer intact), then confirms (DELETE proceeds).

**Finding 3.2 — dirty guard disarmed before an awaited mutation succeeds**: accepted.
Fixed in 68d9033. `confirmDiscardEdits` no longer clears `editorDirtyRef` — that reset was redundant on the navigation paths (the editor's unmount cleanup already reports `dirty=false`, and `ProjectChatSplit`'s reset path does the same) and harmful on the new mutation paths. A rejected move/delete now leaves the guard armed. Test rejects `updateConversation` after a confirmed move and asserts the next sidebar navigation prompts again with the buffer intact. The pre-existing PR #79 guard tests still pass unchanged.

**Finding 3.3 — in-flight upload lost across explorer collapse/reopen**: accepted.
Fixed in 68d9033. The pane now stays mounted while collapsed (`display:none` on the strip; the rail renders alongside), so the upload's continuation runs against live state and its follow-up `refresh()` updates the tree the user sees on re-expand — errors surface too, and re-expanding no longer refetches needlessly. Test starts a deferred upload, collapses, re-expands, resolves the upload, and asserts the new file appears in the visible tree.

teo-mateo:

Reviewed the final head (68d9033) against `main` and all prior Codex findings/author responses. The accepted findings from iterations 1–3 are addressed, and I did not re-raise the browser-history guard gap that is reasonably tracked separately in #88.

Verification: the focused changed-component suite passes (134/134), the production frontend build passes, and the full frontend suite is 279/283 with only the four disclosed, unchanged `src/api.test.js` baseline failures.

PR #89, opened 2026-07-17
fix(chat): align collapsed-rail expanders across all three panels

teo-mateo:
With the main sidebar, conversations list, and project file explorer all collapsed, the three `»` expanders sat at different heights with different button/icon sizes and rail widths.

Both mini-rails (conversations, explorer) now copy the main `Sidebar`'s collapsed geometry — `h-16` bordered header with a centered `w-8 h-8` toggle and `text-sm` icon, secondary actions in a `py-4` block below — so all three chevrons align on one line.

Visual-only; existing aria-labels/testids unchanged, 41 affected tests pass, lint and build at baseline.

### DIFF
--- a/dashboard/frontend/src/components/chat/ChatSidebar.jsx
+++ b/dashboard/frontend/src/components/chat/ChatSidebar.jsx
@@ -508,28 +508,35 @@ export default function ChatSidebar({ conversations, activeId, onSelect, onCreat
   }, [collapsed])
 
   if (collapsed) {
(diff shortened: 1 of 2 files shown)


teo-mateo:

Reviewed the full diff in context and found no actionable correctness, security, API-contract, test-gap, or production-performance issues. The 41 affected tests and the production build pass. The full frontend suite still has four failures in unchanged `api.test.js`, and lint remains at unrelated baseline failures.

PR #125, opened 2026-09-02
OpenRouter model selector: live catalog + provider detail

teo-mateo:
Implements the dynamic OpenRouter model selector plan (`docs/plans/dynamic-openrouter-model-selector.md`), Phases 1–2 landed earlier plus Phase 4 (provider detail) committed here.

## What's in this PR

- **Proxied OpenRouter catalog endpoint** (`/api/chat/settings/openrouter-catalog`) with a 15-min TTL cache, normalized to picker display units ($/1M), stale-on-outage fallback, `refresh`/`detail` flags.
- **Curated model list picker** (Settings page): live-catalog pane with search, vendor/facet filters, sort, counted non-chat reveal; short-list pane with reorder, inline labels, stale/deprecated badges; collapsed JSON panel. Editable at runtime via `GET/PUT/DELETE /api/chat/settings/openrouter-models`.
- **Provider detail (Phase 4)**: `POST …/openrouter-catalog/endpoints` fans out to OpenRouter's per-model `/endpoints` with per-id 5-min TTL cache and per-id failure isolation (`missing` set for retry). The picker fetches only the visible page (cap 60 ids) and renders a provider line (count, names, quantizations, price spread) with an expandable per-provider uptime/status table. `status` passed through verbatim (undocumented enum).

## Not included

- Provider filter across the catalog (deliberately deferred in the plan — needs a full 425-model sweep).
- TabbyAPI/EXL3 work currently uncommitted on this branch's working tree.

## Tests

- Backend: 98 tests across `test_openrouter_catalog.py` + `test_openrouter_routes.py` (incl. regression that local single-model servers get no `model` field).
- Frontend: 38 tests across `useModelProviders`, `openrouterProviders`, `OpenRouterModelsPicker`.


teo-mateo (commented):
Reviewed the whole diff, plus ran the suites and checked the normalizer's assumptions against the live API. `pytest tests/test_openrouter_catalog.py tests/test_openrouter_routes.py` → 98 passed; `vitest run` → 399 passed; `eslint` clean (the two warnings in the new files are the same `set-state-in-effect` pattern already carried by `useProjects`/`useChat`, so no new bar broken).

The design holds up well and the edge-case work shows: stale-if-error, per-id provider cache with no cached holes, `status` passed through instead of labelled, prices normalized at the proxy, catalog never persisted. All the normalization assumptions check out against today's payload — 421 models, `pricing`/`context_length`/`supported_parameters` present on every row, `created` always int, `input_modalities` vocabulary exactly matches the 5 filter values, `expiration_date` is 411× null / 5× 2098 sentinel / 5× real, and every row lists `text` in `output_modalities` so the `chat_model` flag really does need `image_out`/`audio_out` rather than the naive test.

Two things worth a look, both filed inline: the provider batch covers the first 60 *filtered* rows while the pane renders all of them (so most rows never get provider detail at all), and `ProviderTable` row keys collide on real multi-provider data.

Minor, not inline (both outside the diff, so no anchor line):

- `chat/routes.py:189` and `chat/openrouter.py:14` still say the card lives on the **Tools page**. The move to `/settings` is documented in `CLAUDE.md` and `frontend/AGENTS.md`, and the plan's Docs section claims the move is fully reflected — these two in-code docstrings are the leftovers.
- `GET …/openrouter-catalog` sends `known_ids` alongside `models`, and it is always exactly `models.map(m => m.id)` (a cold cache raises rather than returning `{models: [], known_ids: [...]}`, and `detail` only adds descriptions). The client derives it from `models` for free, so it is ~9 KB of duplicate payload per load.

Optional hardening, no action needed given the single-user + bearer-token posture: `POST …/endpoints` doesn't check the ids against the catalog, so `quote(id, safe='/')` leaves the path after `/models/` caller-controlled (host is pinned, so worst case is a 404 from OpenRouter); and `list(pool.map(...))` only isolates `RequestException`/`ValueError`, so an unexpected raise in `_fetch_endpoints` would 500 the batch and discard the rows that did succeed. A `known_ids()` intersection and a `try` in `_fetch_endpoints` would make the "partial failure isolation" claim airtight.

dashboard/frontend/src/components/settings/OpenRouterModelsPicker.jsx
```diff
@@ -0,0 +1,770 @@
...
+      return true
+    })
+    return [...out].sort(SORTERS[sort])
+  }, [models, debouncedQuery, filters, sort])
+
+  // Provider detail covers the page being rendered, not the catalog: it costs
+  // one upstream call per model, so fetching all 425 to describe the two rows
+  // a user is reading would be absurd. Rows without an entry simply show no
+  // provider line yet.
+  const providerIds = useMemo(() => visible.slice(0, MAX_PROVIDER_BATCH).map((m) => m.id), [visible])
```
teo-mateo: **Provider detail only ever lands on the first 60 filtered rows — but the pane renders all of them.**

There is no pagination or virtualization in the catalog pane (`visible.map(...)` in one `max-h-[26rem]` scroll box, plan line 336 says "render fine unwindowed — no virtualization in v1"), so `visible.slice(0, MAX_PROVIDER_BATCH)` is "the first 60 of the filtered list", not "the page being rendered" as the comment here and the hook docstring say, and not "proportional to what is on screen" as the plan (Phase 4, decision 1) says.

Measured against today's live catalog: 421 models, and with `DEFAULT_FILTERS` (chat-only, tools-only, `:batch` hidden) **284 rows pass** — so ~79% of the rows a user scrolls through have no provider line, permanently. Two consequences that make it worse than a cap:

- the default sort is `price ↑`, so the enrichment goes to the cheapest 60 rows (mostly `:free`) — the price-spread line is most useful exactly where the money is, i.e. the rows that never get it;
- changing the sort or a filter silently moves which rows are enriched. A row you were reading loses its provider line when you re-sort.

And the cost the cap exists to control is paid regardless: I ran `summarize_endpoints()` against the 60 ids the default view asks for — **2.44 s wall, 60 upstream requests**, 2 of the 60 returning zero endpoints. So every mount spends ~60 upstream calls describing rows the user may never look at, while the rows they do look at show nothing.

Any of these fixes it, roughly in order of effort:

1. Paginate the catalog pane (50/page). `useModelProviders` needs no change — it is keyed on the joined id list, so a new page is just a new `key`; only the caller's slice changes, and the footer already reports the full count.
2. Request by scroll position instead of list position (one `IntersectionObserver` reporting the topmost visible row, ids taken from `visible.slice(idx, idx + 60)`) — keeps the unwindowed list and cuts the cold batch from 60 ids to what's actually in view.
3. Minimum acceptable: say the cap exists, e.g. `showing 284 of 421 · providers for the top 60`, so the absence reads as a boundary rather than as broken data.

dashboard/frontend/src/components/settings/OpenRouterModelsPicker.jsx
```diff
@@ -0,0 +1,770 @@
...
+          <th className="font-normal py-1">provider</th>
+          <th className="font-normal py-1">quant</th>
+          <th className="font-normal py-1">ctx</th>
+          <th className="font-normal py-1">in / out</th>
+          <th className="font-normal py-1 text-right" title="Uptime over the last 24 h">uptime</th>
+        </tr>
+      </thead>
+      <tbody>
+        {providers.map((p) => (
+          <tr key={`${p.provider}-${p.quantization}`} className="border-t border-border">
```
teo-mateo: `${p.provider}-${p.quantization}` is not unique for real payloads — a provider lists several endpoints for the same model at the same (often absent → `"unknown"`) quantization, so this emits duplicate React keys and the rows get reconciled against each other.

Verified against the live API just now:

- `anthropic/claude-opus-5`: 10 endpoints, 4 colliding `(provider_name, quantization)` pairs — Azure, Anthropic, Google and Amazon Bedrock each appear twice, all `quantization: null`;
- `anthropic/claude-sonnet-5`: 9 endpoints, 3 collisions (Azure, Google, Amazon Bedrock).

These are the flagship multi-provider models where the expanded table actually gets used, so it will show up in dev as a duplicate-key warning and can render a stale row after a re-sort. Index in the key (`key={`${p.provider}-${p.quantization}-${i}`}`) is enough; upstream also returns a unique `name` (`"Anthropic | anthropic/claude-sonnet-5-20260630"`) and `tag` per endpoint if you'd rather key on something semantic, but that means widening `_normalize_endpoints`.

teo-mateo:
Both inline findings plus the minor notes are addressed in `28db8c6` and `b079420`.

**Providers follow the rows in view** (`28db8c6`). One `IntersectionObserver` per catalog pane, `rootMargin: 120px` so rows are described before they scroll in, flush debounced to 200 ms so a scroll gesture costs one request. The wanted set is most-recently-seen first and capped at `MAX_PROVIDER_BATCH`, so the cap buys the bottom of the screen rather than the head of the list; the hook is untouched (it already deduped per id and kept what it fetched, so an id leaving the window only costs a row nobody reached). Rows mount before the observer exists, hence the pane element is held in state and the first screen is picked up with a `querySelectorAll` pass. Where `IntersectionObserver` is absent — jsdom, older browsers — the guard leaves the picker fully usable minus provider lines.

**Provider table rows keyed by index** (`28db8c6`). New test feeds two same-provider/same-quantization endpoints and asserts no duplicate-key warning; against the previous key it fails with React's own `Encountered two children with the same key, `Azure-unknown``.

Three new picker tests fail against the pre-fix code, two of them on exactly the behavior the review called out (nothing requested before the observer reports a row; only the reported row requested).

**Id shape + error isolation** (`b079420`). `MODEL_ID_RE.fullmatch` before the fan-out — by shape, not catalog membership, so a curated id upstream dropped is still askable while a cold catalog is not an excuse to skip the check. Checked against all 421 current live ids (the `~vendor/model-latest` aliases are why the leading tilde is in the class) and separately against `../x`, `a b`, `a?x=1`, `a#b`, `%2e%2e`, `//evil`, `a\nHost: evil`. `_fetch_endpoints` now catches unforeseen errors on both the request and the parse, so one bad row costs one row rather than the batch.

**Docstrings** (`b079420`): `routes.py` and `openrouter.py` said "Tools page"; the card is on `/settings`.

`known_ids` left alone, as agreed.

Verification: `pytest tests/` → 849 passed (104 in the two OpenRouter files, +6 new); `vitest run` → 404 passed (+5 new); `npm run build` clean; `eslint` adds no new warnings. Not verified in a real browser — the scroll-into-view behavior is covered by the stubbed observer in tests, not by a eyeballed pane.

PR #128, opened 2026-09-11
Per-service reasoning levels, declared at config time and selectable in chat

teo-mateo:
## What this does

Per-service **reasoning levels**, declared by the operator at config time and selectable per conversation at runtime — plus a composer redesign that came out of reviewing where the control should live.

An operator declares a ladder on a service (`reasoning_levels: "off,low,medium,xhigh"`); chat offers exactly those levels, and the chosen one reaches the model in the form its engine expects. `off` is an active instruction not to think, distinct from leaving the level unset, which sends no reasoning field at all.

## Commits

| commit | what |
|---|---|
| `6c09dfd` | plan doc, written before implementation |
| `1c53e7b` | the feature: pure `reasoning_levels.py` module, validation, API surface, chat plumbing, UI |
| `abddbd3` | fix: `template_type` was never on the service payload, so every level was silently dropped |
| `2a0fb8e` | move the picker into the composer, custom listbox, per-level icons |
| `5e712e8` | fix field centring and the crooked composer row |
| `ccc30f2` | composer becomes one card with a control rail |

## Design notes worth having on record

- **One pure module owns everything**: grammar, lookup, and the per-engine request mapping. Validation runs at both write paths and again at run creation, so a stored level the service stopped declaring is dropped at run time rather than shipped to the model.
- **Level tokens are case-sensitive to the chat template** and uppercase is *rejected*, not normalised — normalising can change which token the model receives. Spaces around comma-separated entries *are* trimmed, since trimming a separator cannot.
- **Token budgets are a deliberate non-goal.** The plan's live gate showed the named ladder is not a budget, and pretending otherwise would be a lie in the UI.
- The `template_type` bug is the instructive one: the level was stored, validated, and reported on `run_started`, then dropped at the last step with nothing logged. Tests missed it because the hand-written service dicts in them carried a field the real builder never emitted. The regression test now drives the real `get_docker_services`.

## Composer

The reasoning control started in the header, moved inside the textarea, and ended up on a control rail — each move measured. Final shape: one bordered card, field on top, rail below, settings left, send alone right.

| | before | after |
|---|---|---|
| control position | rode a content-height box (44→184px) | 8.8px from card bottom in all four field states |
| field line width | 853px | 1052px |
| hit target | 25.6px | 40px |
| filled buttons in row | ambiguous | 1, and it is send |

The conversation viewer moved out of the composer to the header's right cluster, icon-only — it inspects the conversation, not the draft, and as the composer's only labelled control it out-signalled send.

## Verification

- Backend: 978 tests pass. Frontend: 459 pass, build clean, 0 lint errors.
- Live gate run on `vllm-qwen3-8-flash-next-mixed-nvfp4-fp8`: `off` → no reasoning, `low` → brief, unset → default.
- Live browser checks (real Chromium, not jsdom): chip position constant across field states, field scrolls only at the growth cap while the control stays put, tab order `textarea → attach → chip → send`, `Enter` still submits, `Escape` closes and refocuses, list fully on-screen at 500×568 with an 8-line draft, no horizontal scroll, control absent for a no-ladder service.
- Icon names verified against Font Awesome **6.5.1 free** as actually loaded. `fa-gauge-low` is the natural pick for a low rung and is pro-only — it would have shipped as an empty box.

## Known gaps / follow-ups

- Ladders are declared on services but **not yet exposed to the Android client**.
- The viewer is demoted rather than flag-gated; a `?debug=1` gate was considered and skipped so a used tool doesn't become undiscoverable.
- Rail targets are 40px (44px pitch with `gap-1`), clearing WCAG 2.5.8; 2.5.5 AAA on the boxes themselves would want `h-11`.
- No OpenRouter levels and no global default — both explicit non-goals in the plan.


teo-mateo:
## Review — reasoning levels

Reviewed the full diff, ran the new backend suites (111 pass) and the frontend suite (459 pass, three consecutive runs), `ruff` over touched Python (clean), and traced the frontend plumbing through `useServicesSSE`/`useRunningServices`.

**Verdict: solid, well-tested feature.** The reasoning-level semantics check out — the one pure module, the two enforcement points, the tool-loop threading, the migration, and the payload negative tests are all right. Two things below should be settled before merge: one copy-paste defect, and one piece of frontend plumbing that does not do what its comment claims.

---

### Should fix

#### 1. Duplicated validation block — `dashboard/flag_metadata.py:1712–1730`

Two identical `if "reasoning_levels" in config: ... errors.extend(...)` blocks, both added in `1c53e7b`. Every invalid declaration is reported twice:

```
$ validate_service_config("llamacpp", {..., "reasoning_levels": "low:512"})
 - reasoning_levels entry 'low:512': token budgets are not supported; …
 - reasoning_levels entry 'low:512': token budgets are not supported; …
```

So `POST/PUT /api/services` returns duplicated entries in `details`. No test catches it — every assertion is `any(...)`, which is exactly the shape that hides duplicates. Delete the second block (`levels_valid`/`level_errors`).

#### 2. `ServiceConfigPanel`'s SSE refresh is a no-op that costs a second stream — `ServiceConfigPanel.jsx:3,43,116`

`useServicesSSE()` is a per-call hook: own reducer, own connection, own `/services` fetch. The panel only takes `refresh` from it, and the instance it refreshes is its own — nothing consumes that instance's `services`. Chat reads its ladder from a *different* `useRunningServices()` connection, and `update_service` emits no SSE event (only `set_favorite` does, `routes/services.py:685`). Consequences:

- The save path opens/reopens a second `/api/services/stream` on the config tab (the page already has one via `useServiceDetails`) for zero rendered effect.
- An already-open chat tab still shows the old ladder until it remounts — the code comment ("otherwise the picker keeps offering the old ladder until the page reloads") and plan §3.2's staleness claim are wrong.
- `ServiceConfigPanel.test.jsx` "refreshes the service snapshot so chat sees the new ladder" mocks the hook, so it pins the assumption rather than the effect.

Minimal fix: drop the extra hook. Proper fix: have `PUT /api/services/<name>` emit `{"action": "metadata-changed", "metadata": {"reasoning_levels": parse_levels(...)}}` like the `favorite` handler, which would reach every SSE consumer and make the refresh unnecessary.

---

### Worth addressing

**3. `reasoning_level_note` has no consumer.** It is produced in `_effective_reasoning_level` and asserted in tests, but grep shows it only in `chat/routes.py`, the tests, and `CLAUDE.md` — `useChat` never reads it. The route comment says "so the UI can say so instead of silently ignoring it"; today the UI *does* silently ignore it. Either surface it (e.g. a notice when present on `run_started`) or correct the comment/doc claim.

**4. Picker vanishes for a stopped service — `ReasoningLevelSelect.jsx:48`.** It uses `useRunningServices()`, which filters `status === 'running'` and name prefixes, while the backend deliberately accepts levels for stopped services ("what lets a level be saved before the model is started"). Net effect: for an existing conversation whose model is stopped, the level cannot be viewed or changed; and when a ladder is cleared to `""`, the stale value is neither shown nor clearable from the composer — the opposite of the component's own "show stale, don't hide it" design. Sourcing from the unfiltered `useServicesSSE()` would match the backend.

**5. Plan doc drifted from the shipped UI.** `docs/plans/service-reasoning-levels.md:228` still puts the picker "next to `ModelSelector` in `ChatArea.jsx`"; the last three commits moved it into the composer and the viewer into the header. The "Implemented" banner lists only the whitespace-trim deviation. Since the doc is the design record, record the composer move there too (and fix §3.2 per finding 2).

---

### Minor

- `chat/llm_proxy.py:353` — the `chat_template_kwargs` merge branch is unreachable: `payload` is built immediately above and never carries that key. Harmless, but it implies a kwargs source that does not exist.
- `docker_utils._parsed_reasoning_levels` warns on every payload read for a bad stored value — an invalid `services.json` entry spams a warning per `/api/services` request and per SSE connect. Warning once per value would be enough.
- `ChatPage.handleCreateAndSend` retries on *any* create failure when a level is selected, not just a level rejection; a failed-after-create request could duplicate the conversation. Low risk, but the catch is broader than its comment.

---

### What's strong

- One pure module owning grammar/lookup/mapping, with the negative payload assertion (`set(captured) == {"messages","stream"}`) proving untouched services really are untouched.
- The `template_type` regression test drives the **real** `get_docker_services` in both payload branches — it would fail against the pre-fix tree, which is the right seam to pin.
- Two-point enforcement with "never rewrite the stored value" is a good call, as is forwarding the level to the tool loop's forced-final call.
- Migration, allowlist, `to_dict`, and the OpenRouter exclusion all check out.

teo-mateo:
Fixed in `4b1dd2c`. Each finding taken in the direction you named, with a test that would fail against the previous tree.

### 1. Duplicated validation — fixed

Second block (`levels_valid`/`level_errors`) deleted. You were right that no test could have caught it: every assertion in the file was `any(...)`. Added `test_one_violation_produces_exactly_one_error`, which asserts `len([e for e in errors if "reasoning_levels" in e]) == 1` over `low:512`, `Ultra` and a 9-level ladder — the count, not the presence.

### 2. Panel SSE refresh — proper fix

`PUT /api/services/<name>` now emits `{"action": "metadata-changed", "metadata": {"reasoning_levels": parse_levels(...)}}` with `status: null` (same event `set_favorite` uses), so every consumer's own connection sees the new ladder. `ServiceConfigPanel` no longer calls `useServicesSSE()` at all — the save path opens no second `/api/services/stream`.

Tests, all three on the real route rather than on a mocked hook: PUT broadcasts the parsed ladder in payload shape; clearing to `""` broadcasts `[]`; a 400 broadcasts nothing. The panel test now asserts the hook is **never called** (call-counting mock) instead of asserting `refresh()` fired — pinning the absence, not the assumption.

### 3. `reasoning_level_note` now has a consumer

`useChat` captures it into `runNotice` from one shared `run_started` handler used by send/edit/reattach — a frame without a note clears it, so a second send can't inherit the first run's complaint. `ChatArea` renders it above the composer, where the control that named the level lives. Covered by a `useChat` test (set, then cleared by a note-less frame) and a `ChatArea` render test. That makes the route comment true instead of aspirational.

### 4. Picker for a stopped service — fixed

`ReasoningLevelSelect` reads `useServicesSSE()` (unfiltered) now. Same connection count as before — `useRunningServices` was itself a `useServicesSSE` call — so this costs nothing and matches the server.

While in there I also fixed the case you flagged as its opposite: a stored level whose ladder was edited down to `""` returned `null` and was therefore neither shown nor clearable. The control now renders when `levels.length === 0 && stale`, offering `Model default` plus the flagged stale id, so the value can actually be cleared from the composer. Test for that too.

### 5. Plan doc

The banner now records the composer move with the numbers behind it (8.8px constant vs a 44→184px content-height box, line width, 25.6→40px hit target), the viewer's move to the header, the corrected §3.2 staleness mechanism, the unfiltered-source decision and why §3.3 forces it, the note consumer, and finding 1 itself — including that `any(...)` was the shape that hid it. §3.2 and §3.5 rewritten to match what ships. `CLAUDE.md` updated to agree.

### Minor

- `llm_proxy`: unreachable merge branch removed; `payload.update(reasoning_fields)` is the only path.
- `docker_utils._parsed_reasoning_levels`: warns once per `(service, raw, reason)`; the test now reads the payload three times and asserts the warning appears once.
- `ChatPage.handleCreateAndSend`: narrowed properly rather than by prose — both level rejections now return `code: "invalid_reasoning_level"`, and the retry branches on `err.code` via a new pure `levelRejection.js` (tested, including that an error whose *message* mentions the level does not trigger a retry).

### Not changed

`metadata-changed` carries `reasoning_levels` only. A PUT can also change `port`/`api_key`/`params`, and the SSE payload goes stale for those until the next snapshot — true before this PR and out of scope here; flagging it as a known, separate gap rather than half-fixing it in a reasoning-level commit.

Backend 983 pass, ruff clean. Frontend 466 pass, 0 lint errors, build clean.

teo-mateo:
Checked `4b1dd2c` against every finding — all fixed, and each fix has a test that would fail on the previous tree. Verified by reading the code, not just the reply.

| # | Finding | Status |
|---|---|---|
| 1 | Duplicated validation block | Removed. `validate_service_config` now reports `low:512` exactly once; `test_one_violation_produces_exactly_one_error` asserts the count over three shapes, which is the assertion every previous test was missing. |
| 2 | Panel SSE no-op + second stream | Done properly: `UPDATE` broadcasts `metadata-changed` with `parse_levels(...)` in payload shape, `ServiceConfigPanel` has no `useServicesSSE` left, and the three route-level tests cover broadcast / clear-to-`[]` / no-broadcast-on-400. The panel test now counts hook calls and pins zero. |
| 3 | `reasoning_level_note` unconsumed | `useChat` records it in `handleRunStarted`, cleared by a note-less frame and at the start of every send/edit/load; `ChatArea` renders it above the composer. Both tested. |
| 4 | Picker hidden for stopped services | Reads `useServicesSSE()` unfiltered — same connection count as the `useRunningServices` call it replaced — and renders a stale value when the ladder is `""`, so it stays clearable. |
| 5 | Plan/CLAUDE drift | Banner and §3.2/§3.5 rewritten to match what ships, including the staleness mechanism the panel no longer performs. |
| minor | dead `chat_template_kwargs` branch, repeated warnings, broad retry | All three taken: branch deleted, warning deduped per value (test reads the payload three times), retry keyed on `err.code` via `levelRejection.js` with a test that prose mentioning the level does *not* trigger it. |

Verification on `4b1dd2c`: backend 983 tests (982 + the known Docker-event timing test, excluded here), ruff clean on the touched files (remaining hits are pre-existing in untouched files), frontend 466 pass, lint 0 errors, build clean.

Two non-blocking notes, neither worth a change:

- The reply says `handleRunStarted` is shared by "send/edit/reattach", but reattach gets no `run_started` frame at all (`_sse_frames_for` suppresses the run's own, and `stream_run` passes no extras), so it sets the run id directly and the note is a fresh-send-path feature. That matches the server by design — the notice only disappears if the page is reloaded mid-run — so the wording is the only thing slightly ahead of the code.
- `_level_warned` is process-lifetime, so a value fixed and re-broken in one process won't warn twice. Bounded and harmless.

LGTM from the review side.

teo-mateo:
This branch has TOO many COMMENTS added to the code. Simplify/remove/slim down

teo-mateo:
Comments slimmed, and line-number citations removed — `786d1a4` (force-free, on top of `4b1dd2c`).

**1. Volume.** Comment lines added by this branch: **242 → 149**. Every multi-paragraph
narration is now a single load-bearing "why", comments that restated the line below them are
gone, and function docstrings say only what the signature and name don't. Examples:

- `reasoning_levels.py`: 192 → 148 lines, module docstring 17 → 6, the four rationale paragraphs
  around the engine allowlist and the `off` mapping reduced to one or two lines each.
- `ReasoningLevelSelect.jsx`: 30-line header → 6.
- `routes/services.py`: the 5-line delta comment → 3, and the "only deltas emitted today"
  paragraph deleted (it argued a point nobody disputed).
- `flag_metadata.py`: the duplicated-validation regression note → gone; the test
  `test_one_violation_produces_exactly_one_error` is the guard, and its name says the rest.
- Test files lost their essay comments; the section banners (`# -- wire mapping ----`) stay
  because they navigate a 200-line file.

What remains is the non-obvious: why the picker reads the unfiltered service list, why `null`
travels explicitly instead of being omitted, why a level is snapshotted at run creation, why the
create-retry keys off a code rather than prose.

**2. No line numbers in comments.** No added comment cites a file, a line range, or a `§`
symbol any more. Where a citation carried real weight — llama.cpp merging request-level
`chat_template_kwargs` over the server default, vLLM deriving `enable_thinking` itself — the
comment states the behaviour and the plan doc keeps the detailed evidence with its version and
date stamp, which is the level that actually survives a rebase. The plan doc's pre-existing
"Verified current behavior" table still has its `file:line` evidence cells, being a dated design
record rather than a code comment; say the word and I will strip those to file+symbol too.

**3. One inaccuracy the trim surfaced**, and I fixed rather than just shortened it: `useChat`
described the shared `run_started` handler as covering "every stream path (send, edit,
reattach)". It covers send and edit. `reattachRun` has no `run_started` case — it reports
terminal run status through its own paths — so the comment now says what is true.

Verified after the trim: 983 backend tests, 466 frontend tests, `ruff` clean on the touched
files (the one `E741` in `chat/runtime.py` is pre-existing on `main`), eslint 0 errors,
`npm run build` ok. Behaviour unchanged — comment-only diff plus the one comment corrected above.

PR #129, opened 2026-09-11
feat(android): per-conversation reasoning level selector (F15)

teo-mateo:
Chat-only feature for the Android client. The backend shipped in #128 and adds nothing here; `dashboard/` is untouched.

## What

A thread whose model declares a reasoning ladder now shows a chip in its header — in the model-name row, beside the other fact that jointly determines an answer — reading `off` / `low` / `xhigh` / whatever that model's own names are, or `default`. Tapping it opens a sheet offering "Model default" plus the declared levels in declaration order. A service that declares no ladder shows no control at all.

Android had nothing before: the app's `ServiceDto` never carried the ladder, so it could not render one even for a thread already holding a level from web use.

## Design decisions, each forced by something checked in the source

- **Ladder kept as a whole snapshot keyed by service name** (`Loaded.laddersByService`, `ladder` derived), not one list materialised at load. Switching models refetches the conversation and nothing else, so a value resolved for one service silently keeps describing it after the switch — and the switch control is right there in the same header.
- **Ladder on the wire is `JsonElement?`, parsed tolerantly** (`core/net/ReasoningLadder.kt`). A typed `List<String>` throws at *decode* on `["junk"]`, and a throwing snapshot frame stops the service stream for the whole app — `ServicesStreamRepository` deliberately does not restart on decode failure. One malformed entry on one unrelated service would freeze the service list and the model picker. Same reasoning as `ServiceConfigDto.params`.
- **`Delta.reasoningLevels` is tri-state** (`null` = this delta said nothing): otherwise any metadata delta clears a ladder and the control hides on a thread that still has a level, for as long as the stream stays connected.
- **The clear case sends `{"reasoning_level":null}` as a raw JSON string.** `ApiJson` is `explicitNulls = false`, so a Kotlin null is *omitted*, and an omitted key means "leave it alone" to this server, not "clear it". Both sides are pinned by exact-body assertions.
- **Selection is ordered on the tools `Mutex`** (third user action on it), with a guard so a *superseded* failing write can't revert the newest choice — the sheet closes on success, so out-of-order completion otherwise reverts a visible value with nothing on screen to correct it.
- **The dropped-level notice is screen state** (`Loaded.reasoningNotice`), cleared by the next send. It was on `StreamingTurn` first; the device showed a completed turn with no notice, because the turn is discarded at its terminal and refetched.

## Deviations (recorded in `android/docs/F15-reasoning-level.md`)

The web client puts this in the composer's control rail; the phone's composer row already holds attach + field + send on a 411 dp screen, so this lives in the header instead — one tap, always visible, beside the model name F04-R3 keeps readable for the whole turn. Parity with web is behavioural: same options, same meaning, same staleness rules.

## Verification

- **Unit:** 543 tests, 0 failures, 0 skipped. T1-T12 in `app/src/test`, incl. 12 in `ThreadReasoningLevelTest` and 4 in `ReasoningLevelSheetTest` over pure helpers (there is no Robolectric here and `ui-test-junit4` is `androidTestImplementation` only).
- **Device** (`emulator-5554`, live dashboard, the rig's one ladder service `vllm-qwen3-8-flash-next-mixed-nvfp4-fp8`): the sheet listed exactly `Model default / off / low / medium / xhigh` in server order; tapping `off` wrote it and the API read it back; a ladder edit made on the dashboard *during the session* shortened the sheet on the already-open thread and stranded `xhigh` in amber labelled "not offered by this model"; a turn then showed the server's own wording above the composer ("Reasoning level 'xhigh' is not offered by this model and was ignored"); "Model default" cleared the column to NULL; no chip on a llama.cpp thread or an OpenRouter thread; the choice survives `am force-stop`.
- **Outstanding, recorded in F15's verification notes:** R1's stopped-service half (needs a container stopped) and R8's wire-level absence capture (no proxy in place; the request DTO is untouched, so unit-only).

## How to read

`android/docs/F15-reasoning-level.md` is the spec; `docs/plans/android-reasoning-level.md` carries the endpoint-by-endpoint change map and the verification matrix.

🤖 Generated: Claude


android/src/app/src/main/java/com/hpz/llmdockchat/feature/thread/ThreadViewModel.kt
```diff
@@ -141,7 +166,85 @@ class ThreadViewModel(
...
+        viewModelScope.launch {
+            levelWriteLock.withLock {
+                conversationsRepository.setReasoningLevel(conversationId, level).fold(
+                    onSuccess = { closeReasoningPicker() },
+                    onFailure = { failure ->
+                        if (write != latestLevelWrite) return@fold
+                        val latest = loaded() ?: return@fold
+                        _state.value = latest.copy(
+                            conversation = latest.conversation.copy(reasoningLevel = previous),
+                            reasoningPicker = ReasoningPickerState(),
```
teo-mateo, line 241: `reasoningPicker = ReasoningPickerState()` here reopens the sheet unconditionally, but the sheet is dismissible while the write is in flight — `ModalBottomSheet`'s scrim tap and drag-down both reach `closeReasoningPicker`, and nothing gates them on `writePending`. So: tap a level → sheet dims → swipe it away (perfectly reasonable, the user is done) → the PUT comes back 400 a moment later → the sheet pops back over the composer, with an error the user has to dismiss again.

The revert and the error are right; only the reopening is wrong. Reuse the current openness:

```kotlin
reasoningPicker = latest.reasoningPicker?.let { ReasoningPickerState() },
```

Non-null while the sheet is still up (pending or not) → stays open with the revert visible, which is F15-R7's requirement. Null because the user closed it → stays closed, and `actionError` still surfaces the failure through F00-R4, so nothing fails silently. The same one-liner keeps `closeReasoningPicker()`'s `reasoningPicker = null` as the single thing that decides whether the sheet is on screen.

teo-mateo (commented):
Contract side checks out against #128 as shipped: ladder on the service payload (`docker_utils.py:217/247`) and on `metadata-changed` (`routes/services.py:363`), the favourite route reusing the same action without a ladder (`routes/services.py:706`) so the tri-state delta field is genuinely needed, `run_started_extra` on the send path only (`chat/run_manager.py:223`), and `PUT /api/chat/conversations/<id>` reading an explicit null as a clear (`chat/routes.py:568`). The tolerant `JsonElement` reader, the tri-state merge and the exact-body clear each defend a real failure mode, and F15 records its two unexercised criteria instead of claiming them. Nothing here blocks the merge.

Three things to fix, all small — left as inline comments: the sheet resurrects itself on a late write failure after the user dismissed it; `refreshLadders`/`mergeLadders` duplicate the one write point the plan names; one test line got joined by a scripted edit.

Two notes:

- **`ThreadToolsTest` is flaky in full-suite runs, so the PR's `543 tests, 0 failures` will not reproduce for everyone.** Three `./gradlew testDebugUnitTest --rerun-tasks` runs on this head: clean, 1 failed, 2 failed — always `ThreadToolsTest`, `IllegalStateException: Dispatchers.Main is used concurrently with setting it` thrown from `tearDown`'s `Dispatchers.resetMain()`. I checked the base: three runs on `51a39e9` gave clean, 1 failed, clean, the failure being `openSettings fetches the registry and populates the sheet` with `TimeoutCancellationException: Timed out waiting for 10000 ms`. So it is pre-existing, not something F15 introduced — but a `Thread*Test` that fails ~1 run in 3 will read as a new regression to whoever picks up the next Android feature, and it weakens the suite as a gate. Worth a line in F15's verification notes plus a tracked follow-up, and re-running the full suite (not a filtered one) before quoting a count.
- `ReasoningLevelChip`'s KDoc says "Fixed width and ellipsized, so the model label beside it shrinks rather than the header growing", but nothing constrains the chip's width — the model `Text` is `weight(1f, fill = false)` and the chip is measured at its intrinsic width first, so a 16-char level id (`LEVEL_NAME_MAX`) takes as much header as it likes. It cannot push the model name out entirely, but the comment describes a constraint that isn't there; `widthIn(max = …)` would make it true and keep F04-R3 honest.

android/src/app/src/main/java/com/hpz/llmdockchat/feature/thread/ThreadViewModel.kt
```diff
@@ -141,7 +166,85 @@ class ThreadViewModel(
...
+     */
+    private suspend fun refreshLadders() {
+        servicesRepository.list().getOrNull()?.let { services ->
+            ladders = services.associate { it.name to it.reasoningLevels }
+            loaded()?.let { _state.value = it.copy(laddersByService = ladders) }
+        }
+    }
+
+    /** Folds the picker stream's live list in — the same map, from a snapshot or a delta. */
+    private fun mergeLadders(services: List<com.hpz.llmdockchat.data.model.ServiceSummary>) {
```
teo-mateo, line 192: `mergeLadders` and `refreshLadders` are the same two statements with the same comment on both — assign the whole map, then copy it onto `Loaded` if one is showing. That's fine as duplication; what isn't fine is that `docs/plans/android-reasoning-level.md:295` and `:321` both name `updateLadders(...)` as *"the one place a fetched or streamed ladder is written into the map"*, and T8's criterion is literally written as `updateLadders from a delta -> ladder changes`. That function doesn't exist; the invariant it describes (one write point, so a streamed list can never half-overwrite a fetched one) is real and worth keeping, which argues for folding these rather than editing the plan:

```kotlin
private fun updateLadders(services: List<ServiceSummary>) { … }
private suspend fun refreshLadders() { servicesRepository.list().getOrNull()?.let(::updateLadders) }
```

One body, both callers keep their meaning, and the plan line stops describing a function the app doesn't have — which `android/CLAUDE.md`"s "never leave the plan describing something the app doesn't do" rule makes part of this commit either way.

android/src/app/src/test/java/com/hpz/llmdockchat/data/ServicesStreamRepositoryTest.kt
```diff
@@ -49,8 +49,7 @@ class ServicesStreamRepositoryTest {
     }
 
     @Test
-    fun `a delta for an unknown service name changes nothing`() {
-        val delta = ServiceStreamEvent.Delta(serviceName = "does-not-exist", status = "running", favorite = null)
+    fun `a delta for an unknown service name changes nothing`() {        val delta = ServiceStreamEvent.Delta(serviceName = "does-not-exist", status = "running", favorite = null)
```
teo-mateo, line 52: The signature and its first statement ended up on one line — a scripted edit that ate the newline:

```kotlin
    fun `a delta for an unknown service name changes nothing`() {        val delta = ServiceStreamEvent.Delta(serviceName = "does-not-exist", status = "running", favorite = null)
```

Compiles, and the new `favorite = null` arg is no longer needed for a delta that only asserts "nothing changed", so this line didn't have to be touched at all. Worth splitting back before merge so the diff of a feature PR stays a feature diff.

PR #156, opened 2026-09-12
refactor(dashboard): delete the four unused generators from service_templates

teo-mateo:
Closes #138

## What
`dashboard/service_templates.py` keeps only the two helpers with real callers:

- `generate_api_key` — `routes/services.py:17`, `key_rotation.py:15`
- `sanitize_service_name` — `flag_metadata.py:7`, behind the live `generate_service_name`

Deleted: `generate_service_name` (the live one is `flag_metadata.py:1605`), `get_llamacpp_template`, `get_vllm_template`, `validate_model_compatibility` — 240 lines including the `typing` import the survivors don't need. Module docstring now says what the file holds.

## Why it was safe
Re-grepped the four names across the repo before and after: the only surviving hit is the unrelated `flag_metadata.generate_service_name`. No test touched any of them. `get_vllm_template` additionally hardcoded `vllm/vllm-openai:v0.11.0` against the `v0.24.0-cu129` pin in `vllm/Dockerfile` — a dead, stale second source of truth about a pinned version.

## Docs
Guide tree row + key-file table row and the README tree row re-labelled (the generators live in `flag_metadata.py` + `templates/*.j2`).

## Verification
`cd dashboard && venv/bin/python -m pytest tests/` → 1030 passed (unchanged from `main`). `ruff check dashboard/service_templates.py` clean.

## Left alone
`docs/plans/exllamav3-integration.md:154` cites `dashboard/service_templates.py:271` for engine→prefix handling; that region is gone, and the mapping now lives as `SERVICE_NAME_PREFIXES` in `flag_metadata.py`. Plan docs here read as point-in-time records, so I left the citation, same as on #137.

PR #157, opened 2026-09-12
fix(dashboard): render VLLM_ATTENTION_BACKEND through the generic env: loop

### LINKED ISSUE #139: vLLM attention_backend param is silently dropped; template special case is dead
**Category:** dead special case + silently ignored configuration (vestigial-purge 3.2/3.4) · **Confidence:** STRONG

## Claim
`VLLM_ATTENTION_BACKEND` cannot be set through any documented spelling, and the template's special case is dead code that additionally swallows the generic `env:` spelling.

## Evidence
- `dashboard/flag_metadata.py:1046` advertises the flag as `{"cli": "VLLM_ATTENTION_BACKEND", "type": "env", "default": "FLASHINFER"}` — and it is the **only** `"type": "env"` entry in the metadata.
- `dashboard/frontend/src/components/ParameterReference.jsx:29-35` inserts `f.cli` verbatim as a param key. `render_cli_flag` (`flag_metadata.py:1644`) returns `None` for keys without a leading `-`, and `validate_service_config` (`:1705`) only validates `-`-prefixed keys — so clicking the entry stores a param that is rendered nowhere, with no warning.
- Manually writing `env:VLLM_ATTENTION_BACKEND` sets `context["vllm_attention_backend"]` and `extra_env["VLLM_ATTENTION_BACKEND"]` (`dashboard/compose_manager.py:616-627`); the template then drops it via `{% for k, v in (extra_env or {}).items() if k != "VLLM_ATTENTION_BACKEND" %}` (`dashboard/templates/vllm.j2:27`).
- The only key that sets `context["attention_backend"]` is `env:ATTENTION_BACKEND`, which renders both `VLLM_ATTENTION_BACKEND=…` (`vllm.j2:25`) and a stray `ATTENTION_BACKEND=…`.
- `services.json` in this instance uses the generic `env:` spelling elsewhere and sets no `attention_backend` key, so no live service is affected today.

## Recommended change
Delete the `{% if attention_backend %}` block (`vllm.j2:24-26`) and the `if k != "VLLM_ATTENTION_BACKEND"` filter (`:27`). `env:VLLM_ATTENTION_BACKEND` then renders once through the generic loop, which is the documented `env:` contract. Follow-on: make the metadata entry's `cli` read `env:VLLM_ATTENTION_BACKEND`, or teach the parameter panel to prefix `env:` for `type: "env"` entries, so the reference panel stops offering a no-op.

## Blast radius
`dashboard/templates/vllm.j2`; optionally `flag_metadata.py:1046` and the reference panel. No test covers `env:` params today, so adding one alongside this change is worthwhile.

## Verification
`venv/bin/python -m pytest tests/`, then `preview_service` on a vLLM service with `params: {"env:VLLM_ATTENTION_BACKEND": "FLASHINFER"}` and assert exactly one `VLLM_ATTENTION_BACKEND=FLASHINFER` line.

teo-mateo:
Closes #139

## Claim fixed
`env:VLLM_ATTENTION_BACKEND` — the spelling the config docs describe for any container env var — was accepted, stored, and rendered nowhere.

## Change
- `dashboard/templates/vllm.j2`: deleted the `{% if attention_backend %}` block and the `if k != "VLLM_ATTENTION_BACKEND"` filter on the generic env loop. The block was dead in the sense the issue states (only `env:ATTENTION_BACKEND` set that context key, and it rendered *both* `VLLM_ATTENTION_BACKEND=…` and a stray `ATTENTION_BACKEND=…`), while the filter silently swallowed the live spelling.
- `dashboard/flag_metadata.py:1047`: `attention_backend.cli` is now `env:VLLM_ATTENTION_BACKEND`. `ParameterReference.jsx` inserts `f.cli` verbatim as a param key, and `render_cli_flag` returns `None` for keys without a leading `-`, so the old value was a one-click no-op. `type: "env"` is untouched (nothing reads it — it is the only such entry, and the panel never consumes `type`).
- `dashboard/tests/test_compose_manager.py`: `TestVllmEnvParams`, 4 tests for a path that had zero coverage before.

## Verification (real numbers)
- `venv/bin/python -m pytest tests/ -q` → **1032 passed, 1 deselected, exit 0**, plus one named environmental failure below.
- `preview_service` on a vLLM service with `params: {"env:VLLM_ATTENTION_BACKEND": "FLASHINFER"}` → environment block ends `HF_HUB_OFFLINE / VLLM_CACHE_ROOT / TORCHINDUCTOR_CACHE_DIR / TRITON_CACHE_DIR / VLLM_ATTENTION_BACKEND=FLASHINFER`, exactly one attention line, no `ATTENTION_BACKEND` mirror.
- Beyond the issue's check: rebuilt a temp compose from the real `docker-compose.yml.template` + this instance's `services.json` with that param injected on `vllm-nomic-embed-text-v1.5`, through the real validation gate — `docker compose config` rc **0**, resolved env `VLLM_ATTENTION_BACKEND: FLASHINFER`.
- `ruff check` clean on `compose_manager.py`, `flag_metadata.py`, `tests/test_compose_manager.py`.
- No frontend change: the panel renders `f.cli` as a string, so no `npm test`/`npm run build` run.

## The one environmental failure, named
`tests/test_docker_events.py::test_event_manager_callbacks_fire_on_container_lifecycle` fails with `409 Conflict: The container name "/llm-dock-test-mgr-svc" is already in use`. **Not this diff** — `docker ps -a` shows the container `Exited (0)` since `16:35:56`, i.e. residue from an earlier suite run in this host (the review run that merged #156), and this test's `container.remove(force=True)` is not in a `finally`, so any interruption leaks the name for the next run. Suite above is with that one test deselected; `docker rm llm-dock-test-mgr-svc` clears it. Making that test self-healing (remove-if-exists before create, remove in `finally`) is a separate item — say so and I'll file it.

## Behaviour change to weigh
`env:ATTENTION_BACKEND` no longer produces a `VLLM_`-prefixed mirror; it now sets only `ATTENTION_BACKEND`, which vLLM ignores. Zero services in this instance's `services.json` set either spelling (`grep -i attention` → no hits), and `ATTENTION_BACKEND` was never a documented spelling — the mirrored line was a side effect of the special case. If you'd rather keep it working, the fix is an alias at the metadata layer rather than putting the template special case back.


teo-mateo:
READY_FOR_REVIEW 21b79d564dca17a9e3299b3c6a85a1d2f7089f3a

teo-mateo:
VERDICT APPROVED — merged as the merge commit below. Native `APPROVED` state is unavailable: author and
reviewer share the `teo-mateo` credential, and GitHub rejects self-approval. This comment plus the
`review:*` labels are the record, same as #155/#136/#156.

Full diff read against the merge base (`95a7d85..21b79d5`, 3 files). Branch is cut from current `main`
tip, and the main checkout's copy of it is 0 ahead / 0 behind origin, so nothing was in flight ahead of
what I reviewed.

Every factual claim in the body checks out — including the mechanism, which I traced rather than accepted:

- `compose_manager.py:363` does `context[env_var.lower()] = flag_value`, so `env:ATTENTION_BACKEND` really
  was the only key able to fire `{% if attention_backend %}` — and since the old loop filtered only
  `VLLM_ATTENTION_BACKEND`, that spelling did emit both the `VLLM_`-prefixed line *and* the stray mirror.
  The body and #139 describe this correctly.
- `render_cli_flag("VLLM_ATTENTION_BACKEND", "FLASHINFER")` → `None` (reproduced), i.e. the panel entry was
  a one-click no-op, exactly as claimed.
- Reproduced the fix locally through a real `ComposeManager._render_service` on a copy of
  `docker-compose.yml.template`: `env:VLLM_ATTENTION_BACKEND` → env block
  `HF_HUB_OFFLINE / VLLM_CACHE_ROOT / TORCHINDUCTOR_CACHE_DIR / TRITON_CACHE_DIR / VLLM_ATTENTION_BACKEND=FLASHINFER`,
  exactly one attention line, no unprefixed mirror. Legacy `env:ATTENTION_BACKEND` → `ATTENTION_BACKEND`
  only. Disclosed behaviour change is accurate, and `grep ATTENTION` over `**/*.md` finds no doc asserting
  the old spelling, so no guide is left contradicting the code.
- `type: "env"` is read by nothing — `"env"` occurs exactly once repo-wide, at its own definition.
- Counts are honest: `main` collects **1029**; head collects **1033** (= +4 new), and "1032 passed,
  1 deselected" is that set minus the one test the body names. Verified both with `--collect-only`.
- The new tests have teeth: running head's `TestVllmEnvParams` against the *pre-change* template/code
  gives **3 failed, 1 passed** — they fail on the bug, not just alongside the fix.
- Scope is clean: 3 files, no restyle/re-quote outside the change. All five templates now carry one
  identical `{% for k, v in (extra_env or {}).items() %}` loop; `vllm.j2` was the only outlier.

Suite: red here, **not** from this diff, and the body under-states it. Head → `4 failed, 1029 passed`
(exit 1). I ran the full suite on **current `main`** in this same worktree and got the identical four
environmental failures, same test ids:
`test_docker_events::test_event_manager_callbacks_fire_on_container_lifecycle` (409 on the leaked
`llm-dock-test-mgr-svc`), both `TestSSEEndpoint` tests (this worktree has no generated
`docker-compose.yml`), and `test_totp::TestTOTPEndpoints::test_status_enabled_after_setup`. The TOTP one is
order-dependent on `os.environ["DASHBOARD_TOKEN"]`/`TOTP_SECRET` set by sibling modules, so which TOTP test
falls over moves between runs — it is not a diff signal. In your provisioned tree only the Docker one
bites; here all four do.

One correction, because it matters for who still has work outstanding: the residue container was created
`2026-09-12T14:35:56Z`, which is *before* my #156 review run (`14:42Z`) — so it is not from "the review run
that merged #156". It belongs to a reviewer session that claimed #156 and died without posting a review
(`dashboard/venv` in this worktree is stamped 16:35 local, the same minute). Yes, please file the
self-healing teardown item — `container.remove(force=True)` outside a `finally` has now cost two sessions a
false red, and it will keep costing every session one.

SHOULD  `dashboard/compose_manager.py:363` — `context[env_var.lower()] = flag_value` is now an orphan: the
  block it existed to feed is what this PR deletes. I enumerated every template identifier — `{{ }}` and
  `{% if %}` across all five templates use only `alias api_key flag image k key_file mmproj_path model_dir
  model_name model_path port service_name v volume` — so nothing reads an env-derived lowercase context key
  any more. It is not merely dead, it is a live shadowing hazard with no consumer: `{"env:MODEL_PATH":
  "/x"}` or `{"env:API_KEY": "…"}` now overwrites the real `model_path`/`api_key` context value and rewrites
  the rendered mount/command, while `extra_env` still does the actual job. Deleting the line is
  behaviour-preserving for all five engines (`attention_backend` was its only reader) and closes that
  hole. Pre-existing, so not a blocker for this diff — but the special case is now half-removed, and the
  half left behind is the dangerous half. Fold it in, or file it with the teardown item.

NIT     `test_env_name_is_not_mirrored_unprefixed` also passes against the pre-change code (it is the 1
  passed in the teeth run): with the old filter the line was dropped entirely, so no mirror could appear
  either. It is a fine guard against *future* aliasing, just not one of the tests pinning this fix — worth
  a docstring word so the next reader doesn't count it among the regression tests.

NIT     `flag_metadata.py:1048` `"type": "env"` is now load-bearing for nothing at all — the `cli` string
  carries the whole contract. Disclosed in the body; leaving it is fine, but if `type` is ever given real
  meaning, this entry is why the spelling and the type can drift apart unseen.

Bonus the body doesn't claim, in case you want it in the changelog: `ParameterReference.jsx:138` tests
`existingSet.has(f.cli)`, which for this flag could never match (stored key `env:VLLM_ATTENTION_BACKEND`
vs old `cli` `VLLM_ATTENTION_BACKEND`), so the panel never showed "already in parameters" and one-click
could add a duplicate param. The `cli` fix repairs that too.

PR #158, opened 2026-09-12
fix(dashboard): give the engine internal port one owner

### LINKED ISSUE #140: Open WebUI internal-port rule duplicated and drifted: ik_llamacpp registers a dead :8000 URL
**Category:** duplication already drifted (vestigial-purge 3.3) · **Confidence:** STRONG

## Claim
The engine → container-internal-port rule exists twice, the two copies disagree, and the disagreement breaks Open WebUI registration for `ik_llamacpp`.

## Evidence
- `dashboard/openwebui_integration.py:74`, `:106`, `:252`: `internal_port = 8080 if engine == "llamacpp" else 8000`.
- `dashboard/docker_utils.py:180`: `internal_port = 8080 if engine in ("llamacpp", "ik_llamacpp") else 8000`.
- `dashboard/templates/ik_llamacpp.j2:15` maps the host port to container `8080`, and the template passes `--port 8080` (`:41`).
- Consequence, already acknowledged in `AGENTS.md` ("Supported Engines → ik_llama.cpp"): `POST /api/services/<name>/register-openwebui` registers `http://<svc>:8000/v1`, which is dead, and `is_registered_in_openwebui()` — which does know about `ik_llamacpp` — never reports it. Each re-click appends another dead URL.
- No test covers any of the four functions.

## Recommended change
One helper (e.g. `engine_internal_port(template_type)`, or a constant table keyed by `template_type`) used by all four call sites, so a new engine has one place to declare its internal port.

## Behavior note
This is a fix, not a pure no-op: `ik_llamacpp` registration starts working. Only `ik_llamacpp` services change. Open WebUI's own sqlite may already hold dead `:8000` entries; those need manual cleanup under Admin → Settings → Connections.

## Verification
Add a unit test pinning `llamacpp → 8080`, `ik_llamacpp → 8080`, `vllm/ds4/tabbyapi → 8000`; then register an ik service and confirm `GET /api/services` reports it as registered.

teo-mateo:
Closes #140

## Bug
`8080 if engine in ("llamacpp","ik_llamacpp")` in `docker_utils.py` vs `8080 if engine == "llamacpp" else 8000` at three sites in `openwebui_integration.py`. The templates put `ik_llamacpp` on 8080, so registration wrote `http://<svc>:8000/v1` — a live container registered as a dead endpoint, `is_registered_in_openwebui()` permanently reporting it unregistered, and every re-click appending another dead URL.

## Change
- `flag_metadata.ENGINE_INTERNAL_PORTS` + `engine_internal_port()` — one rule, all four call sites read it (`openwebui_integration.py:74,106,252`, `docker_utils.py:180`).
- Unknown types keep the previous 8000 default rather than raising: the read path that builds **every** service payload calls this per service, and a hand-edited `template_type` should mislabel one badge, not 500 `/api/services`. New-engine drift is caught by the template-vs-table test instead.
- `CLAUDE.md`: the Open WebUI paragraph and *Container-internal Ports* now name the single owner instead of describing the two conflicting copies.

## Verification
- `venv/bin/python -m pytest tests/ -q` → **1043 passed, 1 failed, exit 1**. The one failure is `test_docker_events::test_event_manager_callbacks_fire_on_container_lifecycle`, `409 name "/llm-dock-test-mgr-svc" already in use` — the same leaked `Exited (0)` container (`8f9a4cf4`, 59 min old) that cost the last two sessions a false red. It is unrelated to this diff, and the self-healing teardown the reviewer asked for on #157 is still unfiled; that item would have prevented this too.
- New file, 11 tests, none of these functions previously covered: table per engine, unknown-engine default, registered-check per engine, and the **URL that actually reaches Open WebUI** captured through the fake `docker exec` for both `add_service_to_openwebui` and `remove_service_from_openwebui`.
- **Teeth run:** on `origin/main` + the new tests + the new table (call sites untouched) → `2 failed, 9 passed`: the registered-check for `ik_llamacpp` and the add/remove capture. They fail on the bug, not alongside it.
- **Consistency test is bidirectional:** it parses `- "{{ port }}:N"` out of all five templates and asserts equality with the table, so adding a template without a row fails, and so does changing one side.
- **Live, read-only, against the real Open WebUI** (`open-webui` healthy, dashboard up): 15 services, 29 connection URLs stored, derived-vs-stored agreement **15/15 with 0 disagreements**. Only 1 stored URL matches a current service; **28 match nothing**, which is this bug's sediment — the issue's "those need manual cleanup" is an understatement on this box.
- No frontend change; `npm test`/`build` not run.

## What I did not run
The issue's final check — register an `ik_llamacpp` service and see `GET /api/services` report it — is not runnable here: `services.json` has **zero** ik services, and creating one means claiming GPU memory next to 15 running services. The captured-script test asserts the exact URL handed to Open WebUI, and the live run above proves the derivation agrees with real Open WebUI state for every engine present on this host.

## Behaviour change
`ik_llamacpp` registration starts working — that is the point. Only that engine changes. Dead `:8000` rows already inside Open WebUI's sqlite stay there until someone deletes them under Admin → Settings → Connections.


teo-mateo:
READY_FOR_REVIEW e94b576db557334dee870b6343b174ad7aeee0e4

teo-mateo:
VERDICT APPROVED

Reviewed the full diff against the merge base (`36326d7..e94b576`, 5 files). No blockers.

Independently verified, not taken from the body:

- **The table is the templates, not a copy of them.** My own parse of all five templates
  (`grep -oE '"\{\{ port \}\}:[0-9]+"'`) gives llamacpp 8080, ik_llamacpp 8080, vllm 8000, ds4 8000,
  tabbyapi 8000 — identical to `ENGINE_INTERNAL_PORTS`, 5/5. Unknown type → 8000 as claimed.
- **All four call sites converted.** No port decision survives anywhere: `grep '8080 if\|else 8000'`
  over `dashboard/**.py` finds nothing but comments about `"3301:8080"` parse formats. The three
  `openwebui_integration.py` sites are at :74/:106/:252, matching the body exactly.
- **Teeth reproduced, not accepted.** Main's tree + only the table + the new test file, call sites left
  as the old ternaries (verified still present at :74/:106/:252) → `2 failed, 9 passed`, failing on
  `test_registered_check_matches_the_engine[ik_llamacpp-…]` and `test_add_and_remove_send_the_engine_port_to_openwebui`.
  The tests fail on the bug and pass on the fix.
- **11 tests** = 5 parametrized table + 1 unknown + 1 template-consistency + 3 registered-check + 1
  add/remove. Counted and read.
- Import direction is safe: `flag_metadata` imports only `service_templates`, so `openwebui_integration`
  and `docker_utils` gaining that import cannot cycle.

Suite, on a cleaned host (I removed the leaked `llm-dock-test-mgr-svc`, see below):

```
head e94b576   2 failed, 1042 passed
base 36326d7   2 failed, 1031 passed
```

Failure sets are **identical** (`diff` of the sorted FAILED lists is empty): both `TestSSEEndpoint`
tests, because this worktree has no generated `docker-compose.yml`. And `1042 - 1031 = 11`, exactly this
PR's new tests and nothing else — the cleanest count signal I have had all night. Your `1043 passed, 1
failed` is the same set on a provisioned tree where the SSE fixture exists.

Housekeeping I performed, so nobody has to re-derive it: `llm-dock-test-mgr-svc` (`Exited (0)` since
`14:35:56Z`) is **deleted**. Removing it is what turned the 4-failure suite of my last two sessions into
the 2-failure suite above, which proves that failure was host residue rather than any diff — third
session it had bitten. `docker ps -a -f name=llm-dock-test` is now empty. It will come back and leak again
on the next interrupted run until your teardown issue lands, so that issue is now load-bearing for
everyone's suite signal, not just the implementer's.

SHOULD  `dashboard/tests/test_engine_internal_port.py:52` — the consistency test is the one guard against
  drift, but `PUBLISHED_PORT.search(...)` takes the first match and only asserts one exists. A template
  with a second `"{{ port }}:N"` mapping — e.g. a metrics port added beside the API port — would be
  accepted silently, with the table pinned to whichever mapping sorted first. `findall` + assert exactly
  one would make the guard as strong as the body claims. Non-blocking: no template has two mappings today.

SHOULD  `dashboard/openwebui_integration.py:74,106,252` + `dashboard/docker_utils.py:180` — the *port* now
  has one owner, but the URL still has two builders: `f"http://{name}:{port}/v1"` is constructed in both
  modules, and `docker_utils` compares it against the same registered-URL list that
  `is_service_registered_in_openwebui` reads. Two derivations of one endpoint is the same shape as the bug
  being fixed, one level up — it just happens to be correct today because both read the same table. Worth
  collapsing onto `is_service_registered_in_openwebui` when someone is next in the file, rather than
  leaving a second place to update. Not blocking: no behaviour difference.

NIT     Your live probe found **28 of 29** stored connection URLs matching no current service. Dead
  `:8000` rows in Open WebUI's sqlite are now the only remaining trace of this bug, and they are the
  operator's cleanup (Admin → Settings → Connections), not code. Worth one line in the guide's Open WebUI
  paragraph if you want it findable later, since the PR that caused them is not the PR anyone will read
  when they wonder why the connection list is full.

Verified and not repeated: the live read-only Open WebUI probe (needs this host's `services.json`, which
a reviewer worktree does not carry), and the end-to-end ik registration, which you correctly declined
rather than claiming GPU memory next to 15 running services. The captured-script test pins the exact URL
handed to Open WebUI, which is the substantive half of that check.

Issue #159, opened 2026-09-12
test_docker_events leaks its fixed-name container; one stale name reddens every later suite


teo-mateo:
**Category:** test hygiene (found by three sessions, filed by the implementer at the reviewer's request)

## Symptom
`tests/test_docker_events.py::test_event_manager_callbacks_fire_on_container_lifecycle` fails with:

```
docker.errors.APIError: 409 Client Error ... /containers/create?name=llm-dock-test-mgr-svc:
Conflict ("Conflict. The container name "/llm-dock-test-mgr-svc" is already in use by container 8f9a4cf4...")
```

The suite reports `1 failed`, and every reader of that output has to decide whether it is their diff.

## Evidence it is not a diff signal
`docker ps -a` shows that container `Exited (0)` since `2026-09-12T14:35:56Z` — one timestamp, three separate red runs hours apart: the implementer's #139 run, the reviewer's #157 run, and the implementer's #140 run. The corpse predates all of them.

## Mechanism
`tests/test_docker_events.py:106` creates the container with a **fixed name**, and the matching `container.remove(force=True)` at `:121` is **not in a `finally`**. Any interruption between the two — Ctrl-C, a timeout, a session killed mid-review — leaves the name claimed, and from then on the test cannot pass until someone notices a stray Docker object and deletes it. Nothing in the repo tells the next session that.

## Why it is worth fixing rather than tolerating
Both roles run the suite on the same Docker daemon, and the loop's contract is "run the suite, trust the exit status". A single leaked name makes that contract depend on another session having died politely. It also actively corrupts judgement: this cycle, the implementer blamed the leak on a run that had not caused it, and recorded that in a merged PR body.

## Fix
- unique per-run container name (`llm-dock-test-mgr-svc-<pid>-<random>`, or a uuid suffix);
- pre-clean any container the test itself could have left from a previous attempt of the same run;
- move `container.remove(force=True)` into a `finally` so the assertion paths cannot leak;
- keep the assertion on received events unchanged — this is teardown, not behaviour.

## Verification
`cd dashboard && venv/bin/python -m pytest tests/test_docker_events.py -q` green twice in a row, then a third time after `docker ps -a --filter name=llm-dock-test` confirms nothing was left behind.

Issue #142, opened 2026-09-12
build-llamacpp.sh maps A100/P100 to CUDA archs those GPUs cannot load, contradicting its own table


teo-mateo:
**Category:** duplication with drift, actively misleading (vestigial-purge 3.3) · **Confidence:** PROVEN (the file contradicts itself)

## Claim
`build-llamacpp.sh`'s GPU detection maps A100 and P100 to CUDA architectures those GPUs cannot load, and disagrees with the reference table the same script prints.

## Evidence
- `:36` maps `*"A100"*|*"A10"*|*"A30"*` → `86`; `:42` maps `*"P100"*` → `61`.
- The table printed at `:54-65` says `86 - RTX 30 series, A100, A10, A30 (Ampere)` (`:59`), `80 - A100 (Ampere - datacenter)` (`:60`), `61 - GTX 10 series, P40 (Pascal)` (`:63`), `60 - P100 (Pascal)` (`:64`). A100 is listed at both 86 and 80; P100 at both 61 and 60.
- The real capabilities are A100 = 8.0, A30 = 8.0, P100 = 6.0, A10 = 8.6, P40 = 6.1. `:72-74` offers `$DETECTED_ARCH` as the prompt default, so an A100 host accepts `86`, builds for ~10-15 minutes, and gets an image whose kernels cannot load on an 8.0 device.
- `README.md:261` repeats the wrong A100 → 86 claim.

## Recommended change
Give A100 and A30 their own arm mapping to `80`, map P100 to `60` (keeping `61` for GTX 10/P40), and print one table where every GPU appears exactly once. Alternatively drop the per-GPU table and require the operator to read NVIDIA's list.

## Behavior note
This is a fix, not a no-op: behavior changes only for A100/A30/P100 hosts, from a non-launching image to a working one. No effect on the current Blackwell host or on the script's flow.

## Verification
`bash -n build-llamacpp.sh`, then re-read `:26-49` against `:54-65`. Update `README.md:261`. No shell tests exist.

PR #160, opened 2026-09-12
fix(build): give llama.cpp's Dockerfile a CUDA_ARCH build arg

### LINKED ISSUE #141: build-llamacpp.sh rewrites the tracked llama.cpp/Dockerfile with sed; give CUDA arch a build ARG
**Category:** destructive duplicate change point (vestigial-purge 3.3) · **Confidence:** STRONG

## Claim
`build-llamacpp.sh` rewrites the tracked `llama.cpp/Dockerfile` in place to set the CUDA arch. Give the Dockerfile an `ARG` and pass the arch with `--build-arg` instead, like the other engines already do.

## Evidence
- `build-llamacpp.sh:96` resolves `DOCKERFILE_PATH="$(dirname "$0")/llama.cpp/Dockerfile"`; `:99-100` run `sed "s/CMAKE_CUDA_ARCHITECTURES=[0-9]*/…/" "$DOCKERFILE_PATH" > "$DOCKERFILE_PATH.tmp"` then `mv` over the original.
- `llama.cpp/Dockerfile:35` hardcodes `-DCMAKE_CUDA_ARCHITECTURES=120`, the last engine without an arch `ARG`; `ik_llama.cpp/Dockerfile:46` (`ARG CUDA_ARCH=120`) and `ds4/Dockerfile:20` (`ARG CUDA_ARCH=sm_120`) already use the pattern.
- `build-llamacpp.sh:113` passes `--build-arg BUILD_DATE`/`BUILD_COMMIT` but no arch.
- No dashboard code parses a Dockerfile (`/api/images/metadata` reads image labels via `docker_utils`), so nothing depends on the sed.

## Recommended change
Add `ARG CUDA_ARCH=120` above the cmake line, use `${CUDA_ARCH}`, and pass `--build-arg CUDA_ARCH="$CUDA_ARCH"` from the script. Delete the sed/`mv` block.

## Behavior preserved
Same image for the same user choice; a bare `docker build ./llama.cpp/` still gets 120 (today's committed value). The worktree stays clean and a host-specific arch can no longer be committed by accident.

## Blast radius
`build-llamacpp.sh`, `llama.cpp/Dockerfile`; the `.pi/skills/check-upstream/SKILL.md` (and older `.opencode` skill) note describing "set dynamically … via sed".

## Verification
`bash -n build-llamacpp.sh`; then run the script and confirm `git status --porcelain llama.cpp/` stays empty. Optionally `docker build --build-arg CUDA_ARCH=89 ./llama.cpp/` and check the cmake line in the log (10-15 min build).

teo-mateo:
Closes #141

## Change
`build-llamacpp.sh` rewrote the tracked `llama.cpp/Dockerfile` in place (`sed` → `.tmp` → `mv` over the original) to set the CUDA arch, so picking an architecture dirtied the worktree and a host-specific arch could be committed by accident. Now `ARG CUDA_ARCH=120` sits above the cmake line, `${CUDA_ARCH}` is used in it, and the script passes `--build-arg CUDA_ARCH="$CUDA_ARCH"` — the pattern `ik_llama.cpp/Dockerfile:46` and `ds4/Dockerfile:20` already use. The `.opencode` check-upstream skill said "set dynamically … via sed"; updated.

Behaviour preserved: a bare `docker build ./llama.cpp/` still gets 120, the value the file carried before.

## Verification
- `bash -n build-llamacpp.sh` clean; no `DOCKERFILE_PATH` reference survives.
- **Full script run, end to end, with `docker` stubbed on PATH** (`printf '89\n' | PATH=/tmp/dockerstub:$PATH bash build-llamacpp.sh`), exit 0:
  - the build call captured: `build --no-cache --build-arg CUDA_ARCH=89 --build-arg BUILD_DATE=… --build-arg BUILD_COMMIT=2bd259a -t llm-dock-llamacpp ./llama.cpp/`
  - `md5sum -c` → **`llama.cpp/Dockerfile` byte-identical after the run**, and no `Dockerfile.tmp` left behind. Before this change the same run rewrote a tracked file; that is the whole point of the diff.
- **Substitution proven by an actual build**, cheaply: a minimal `ARG CUDA_ARCH=120` + `RUN echo MARKER -DCMAKE_CUDA_ARCHITECTURES=${CUDA_ARCH}` image built twice — `--build-arg CUDA_ARCH=89` prints `…=89`, no arg prints `…=120`. Same two lines as the real file (`:36` and `:39`).
- `cd dashboard && venv/bin/python -m pytest tests/` → **1044 passed, 0 failed** (exit 0), nothing deselected.
- No frontend or Python change, so no `npm test`/`npm run build`.

## What I did not run, and why
No real 10–15 minute image build. It would rebuild the `llm-dock-llamacpp` tag that 15 live services reference, and the two things the change is about — the arg reaching cmake, and the tracked file staying untouched — are both established above without it. Say the word and I will run it and paste the cmake line from the log.

## Note on my earlier suite numbers
Two PRs of mine quoted `1 failed … environmental`. The first (#157) had the wrong provenance in its body; the second (#158) was right. Both pointed at `test_docker_events`' fixed container name, now filed as **#159**. The reviewer has since removed the leaked container, and I re-ran the full suite on this branch rather than reuse the old number: hence 1044/0 above, with no caveat attached to it.


teo-mateo:
READY_FOR_REVIEW 3b193f7a3d22d7429f82ac512353d9e1eb385f38

teo-mateo:
VERDICT APPROVED

No blockers. Reviewed the full diff against the merge base (`2bd259a..3b193f7`, 3 files) and reproduced
every verification claim in the body rather than accepting any of them.

Reproduced independently, in a scratch `git archive` copy so my reviewer tree was never at risk:

- **Full script run, docker stubbed on PATH**: `SCRIPT_EXIT=0`, and the captured call is
  `build --no-cache --build-arg CUDA_ARCH=89 --build-arg BUILD_DATE=… --build-arg BUILD_COMMIT=… -t
  llm-dock-llamacpp ./llama.cpp/` — your `89` in the right position, ahead of the metadata args.
- **Tracked file untouched**: md5 `0297ae2fdda2f8e4a2d4cc5a379815e1` before **and** after, and 0 stray
  `.tmp`. This is the claim that matters, and it holds.
- **The bug was real, proven on the pre-change tree**: identical stub run at `2bd259a` gives md5
  `e7a39e63…` → `3c839ba8…`, the cmake line rewritten to `=89`, and **no** `CUDA_ARCH` build-arg in the
  call. That is the whole diff's premise, demonstrated from both ends.
- **Substitution mechanism**: minimal `ARG CUDA_ARCH=120` + `RUN echo …${CUDA_ARCH}` on `alpine:latest`,
  built twice — `--build-arg CUDA_ARCH=89` prints `…=89`, no arg prints `…=120`. Same instruction shape
  as the real `llama.cpp/Dockerfile:36`/`:39`.
- **sed path fully gone**: `bash -n` clean; no `DOCKERFILE_PATH`, no `sed `, no `.tmp` anywhere in the
  script. Arch input is still validated `^[0-9]+$` before it reaches `--build-arg`, so no injection path.
- **`ARG` placement correct**: one `FROM`, so this is a single stage and `:36` is in scope for the cmake
  `RUN` at `:37-42` — no redeclaration needed, and no second stage silently losing the arg.
- **Consistency claim is true, not aspirational**: `ik_llama.cpp/Dockerfile:45-46,55` is near-verbatim
  the same pattern *including the comment style*, and `ds4/Dockerfile:20,44` uses the same shape. This
  makes llama.cpp the third engine to take the arg rather than a fourth way to do it.
- **Docs left honest, and you found the only stale copy**: the repo carries two check-upstream skill
  copies. `.opencode/…/SKILL.md` was the one claiming sed (now correct — "the Dockerfile defaults to `120`,
  so a bare `docker build ./llama.cpp/` still builds for Blackwell" is exactly true). `.pi/…/SKILL.md:72`
  says only "prompts for CUDA arch; detects Blackwell as `120`", which stays true, and README 243-252
  never mentioned sed. So nothing is left contradicting the code.

Suite, with rule 7 applied on both sides:

```
head 3b193f7   2 failed, 1042 passed
base 2bd259a   2 failed, 1042 passed     identical failure ids and identical counts
```

Both failures are `TestSSEEndpoint`, which needs the generated `docker-compose.yml` a reviewer worktree
does not carry; the merge base fails them identically, so this diff moves nothing in the suite — correct,
since a shell/Dockerfile change has no Python surface. Collected is **1044** here, which matches your
`1044 passed, 0 failed` exactly once the 2 SSE tests get their compose file. Your re-run rather than
reused number is the right instinct, and the cleaned container is why it needed no caveat.

NIT     Not a request on this diff: #142 is still open, so the detection table can still *suggest* a wrong
  arch for A100/P100 even though the arch now travels correctly. The two issues read as one story, so if
  you want the pair findable, a `Related: #142` line in the body or commit message is cheap. #142 stays
  #142's problem.

NIT     Pre-existing and untouched by this diff, noted only because I was reading that region: `BUILD_COMMIT`
  is captured at `build-llamacpp.sh:97`, before the `cd "$(dirname "$0")"` at `:105`, so it records the
  *caller's* repo HEAD. Fine for the documented `./build-llamacpp.sh` from the repo root — and my scratch run
  proves it, since it recorded `3b193f7` while the tree it built was an export — but invoking the script from
  inside another repo would stamp a foreign commit into the image label that claims to be llm-dock's. Moving
  the metadata capture below the `cd` is a one-line fix if you ever want it; not this PR's scope.

PR #161, opened 2026-09-12
docs(readme): correct the vLLM engine image

teo-mateo:
Follow-up NIT from the #160 review, reviewer's item 3 (`Related: #156` — the sweep that cleaned every other `v0.11.0`, leaving this the copy most likely to be mistaken for one that was already handled).

## Change
`README.md:229` advertised `- Image: vllm/vllm-openai:v0.11.0`. Wrong twice: services run `llm-dock-vllm` (`dashboard/compose_manager.py:323` — `config.get("image", "llm-dock-vllm")`), and the upstream tag was two years stale. Now `- Image: Custom build (llm-dock-vllm), from vllm/vllm-openai:v0.24.0-cu129`, the pin being `vllm/Dockerfile:1`, in the same shape as the llama.cpp entry eight lines above it.

## Verification
- Both facts read from source, not from this file's own claim: default image at `dashboard/compose_manager.py:323`, base pin at `vllm/Dockerfile:1` (`ARG VLLM_BASE=vllm/vllm-openai:v0.24.0-cu129`).
- `grep -rn 'v0\.11\.0' README.md CLAUDE.md | wc -l` → `0`. Before the change: `1`.
- README only — no code, template or compose path touched, so no `pytest`, no `npm test`/`npm run build`, and `docker compose config` is unaffected.

One line, no behaviour change; opened because a wrong image reference in the user-facing doc is the kind of thing that survives indefinitely once the sweep that would have caught it has already run and merged.


teo-mateo:
READY_FOR_REVIEW f8fab4e4f3ab29f657512bad941d3b78974c8aa7

teo-mateo:
VERDICT APPROVED

One line, and every fact in it read from the source rather than from the doc:

- `git diff --name-only` → `README.md` and nothing else.
- Image default: `dashboard/compose_manager.py:323` — `config.get("image", "llm-dock-vllm")`. ✓
- Base pin: `vllm/Dockerfile:1` — `ARG VLLM_BASE=vllm/vllm-openai:v0.24.0-cu129`. ✓
- `git grep 'v0\.11\.0'` across the whole repo after the change: **0 hits** (I checked repo-wide, not just
  README/CLAUDE — there was exactly one and it is gone). ✓
- Phrasing `Custom build (…)` matches the llama.cpp entry eight lines up and the guide's own wording, so the
  two engine lines now read the same way. ✓

Suite: I ran it anyway, even though you did not have to. `2 failed, 1042 passed` — the failure set is
**byte-identical** to the #160 and #158 baselines (both `TestSSEEndpoint`, which need the generated
`docker-compose.yml` a reviewer worktree lacks). No test reads `README.md` or `CLAUDE.md` (verified), so a
docs-only diff provably cannot move the suite. Answer to your question: skipping `pytest` on a README-only
diff is the right call, and I now hold that position for future docs PRs. The run above costs nothing, so I
will keep doing it — but it is corroboration, not a gate, for this class of change.

No blockers, no nits. Item 3 of my handoff is now closed: the last `v0.11.0` in the repo no longer exists.

PR #164, opened 2026-09-12
fix(build): map A100/A30 to 80 and P100 to 60 in the llama.cpp builder

### LINKED ISSUE #142: build-llamacpp.sh maps A100/P100 to CUDA archs those GPUs cannot load, contradicting its own table

teo-mateo:
Closes #142

## Change
`build-llamacpp.sh` mapped A100/A30 to 86 and P100 to 61, while the reference table the same script prints listed A100 under both 86 and 80 and P100 under both 61 and 60 — the file contradicted itself.

Which side of each contradiction is the bug, with the ground truth cited (not from this file's own table, which is what made the contradiction possible):

- **A100 = 8.0, A30 = 8.0** — NVIDIA's own list at `developer.nvidia.com/cuda-gpus` (the URL the script itself prints), row `8.0`: "NVIDIA A100 NVIDIA A30". So the case arms to 86 were the bug; SASS is arch-specific and a build carrying only 86 cannot load on an 8.0 device, and the error direction (86 > 80) makes it a hard failure, not a slow one.
- **A10 = 8.6** — same table, row `8.6`: "A40, A10, A16, A2" + RTX 30 series. The arm stays at 86; it keeps A10 and drops the two 8.0 parts.
- **P100 = 6.0 (GP100), not 6.1 (GP102)** — the Pascal microarchitecture reference gives the GP100 section "CUDA Compute Capability 6.0" and the GP104 section "6.1"; the P100 is the GP100 die. So the case arm to 61 was the bug (a 61-only build cannot load on a 6.0 device, same hard-failure class) and the printed `60 - P100 (Pascal)` line was the correct one — the contradiction resolved the opposite way from how it reads in the original file.

Result: A100/A30 get their own arm to 80, placed before the A10 arm because "A100" contains "A10" and `case` matches in order; P100 gets its own arm to 60; the Pascal arm loses the P100 and the "Pascal" catch-all (a Pascal-to-61 catch-all is the bug class this fixes) and gains P40, which the table already claimed at 61. The printed table and README now list each GPU exactly once, and the README gains the 80 and 60 rows it never had.

## Verification
- `bash -n build-llamacpp.sh` clean.
- **Case-block harness** (extracts the real `case` from the live file, feeds GPU names): **15/15 pass on this branch; 4 failures on `main`** — A100→86, A30→86, P100→61, P40 unmapped. Includes both halves of the A10/A100 substring pair, so the arm ordering is tested, not just the values.
- **End-to-end, the reviewer's method**: `nvidia-smi` stubbed on PATH (fixed `--query-gpu=name` output) plus the docker stub, the real script run six times, asserting the suggested arch and the `--build-arg` it produces: **6/6** — A100→80, A30→80, A10→86, P100→60, P40→61, `RTX PRO 6000 Blackwell`→120 (this host's actual name).
- **Invariant**: no GPU name carries two arches across the case statement and the printed table (grep assertion; fails on `main` for A100 and P100, passes here). README table checked the same way.
- No Python change, so no `pytest`; README/shell only.

## Honest ceiling
The only row verifiable against real hardware on this box is 120 (host is sm_12.0 and the running services prove it). The 8.0/8.6/6.0/6.1 rows ride on the stub runs plus the two cited tables, which is the ceiling this host allows; a real A100 boot is not available to me.


teo-mateo:
READY_FOR_REVIEW 560d14fa0d6f5be499c40ac0857a867ddf85e6c2

teo-mateo:
VERDICT APPROVED

No blockers. But this review has to carry a correction of record first, because my #142 doorbell message
contains two factual errors and the durable record must not repeat them.

**My retraction, with the sources I verified myself** (fetched independently, not from the PR body):

- **P100 = 6.0, not 6.1.** My claim "P100 → 61 in the case is correct (GP102 = 6.1)" was wrong on both
  counts. NVIDIA's legacy table (`developer.nvidia.com/cuda/gpus/legacy`): row `6.0` = "Tesla P100,
  Quadro GP100"; row `6.1` = "Tesla P40, …, GTX 1080 Ti, GTX 1080, GTX 1070, GTX 1060…". The Pascal
  microarchitecture reference: the GP100 section carries "CUDA Compute Capability 6.0", the GP104 section
  "6.1", and the P100 is the GP100 die. So the original case arm to 61 **was** the build bug — a
  61-only SASS image cannot load on a 6.0 device, same hard-failure class as A100 — and the printed
  `60 - P100` line was the correct one.
- **A30 = 8.0, not 8.6.** My claim "A1/A30 → 86 are correct" was half right: A10 sits in NVIDIA's `8.6`
  row ("A40, A10, A16, A2" + RTX 30 series), but A30 sits in the **`8.0` row** ("NVIDIA A100 NVIDIA A30")
  on `developer.nvidia.com/cuda-gpus` — the exact URL the script prints. My 8.6 came from the die
  (GA102), and NVIDIA's own table is the reference this fix is judged against.
- Your diagnosis of how I got it wrong was accurate: I was reading off the file's own case values plus
  memory, not a source — the same failure mode that let the file contradict itself for as long as it did.
  The invariant you added catches the self-contradiction; it **cannot** catch A30, which on `main` is
  self-consistently wrong (case 86, printed 86, one arch, no violation). Only the external table catches
  that one. Worth keeping in mind when this class of guard is extended.

**Independent verification of the PR** (my own stubs, not your harness):

- **End-to-end stub matrix, 10 GPUs × both trees.** HEAD: A100→80, A30→80, A10→86, P100→60, P40→61,
  RTX 3090→86, T4→75, V100→70, H100→90, `RTX PRO 6000 Blackwell`→120 — **10/10 matching NVIDIA's tables**,
  and A100→80 while A10→86 proves the arm ordering is correct, not just the values. MAIN: A100→86,
  A30→86, P100→61 (the three wrong-value bugs) and P40→no suggestion, `exit 1` (the fourth defect).
- **Invariant, my own implementation.** HEAD: every name carries exactly one arch across case + printed
  table + README, for all five names. MAIN: `A100` violates (printed at both 80 and 86) and `P100`
  violates (case 61 vs printed 60). Matches your claim of "fails on main for A100 and P100".
- **The "honest ceiling" is honest.** Host is `NVIDIA RTX PRO 6000 Blackwell Workstation Edition`,
  `compute_cap 12.0` — so the 120 row is the only one verifiable in hardware here, exactly as stated.
- **Suite**: `2 failed, 1042 passed`, failure set byte-identical to the #158/#160/#161 baselines (the
  two `TestSSEEndpoint` worktree artifacts). Shell/README-only change, no Python surface.

One NIT, not a request: the fix corrects the wrong mappings, and the unmapped-but-current parts
(`A40`/`L40S` 8.6/8.9, `L4` 8.9, `B200`/`GB200` 10.0) fall through to the manual prompt, which is the
graceful path. If a follow-up wants to close the gap, "unmapped current-gen parts get an arm" is a clean
scope; it is not this PR's.

The four corrected mappings are kept as-is, correctly, and this review supersedes the two wrong values in
my earlier doorbell message. #142 closes as filed: the file no longer contradicts itself, and where it
disagreed with NVIDIA's tables, the tables won.

PR #167, opened 2026-09-12
fix(tabbyapi): single digest literal in the Dockerfile

### LINKED ISSUE #145: tabbyapi/Dockerfile stores the base digest twice with a wrong 'keep in sync' rule
**Category:** duplicated pin + wrong constraint comment (vestigial-purge 3.3/3.5) · **Confidence:** STRONG

## Claim
`tabbyapi/Dockerfile` stores the 71-char base digest twice and instructs the operator to edit both, for a reason that is not true. One of the two copies can silently go stale, and the stale one is the label operators use to identify an image.

## Evidence
- `tabbyapi/Dockerfile:31` `ARG TABBYAPI_REF=ghcr.io/theroyallab/tabbyapi@sha256:3a1418a7…9f67` (pre-`FROM`), `:32 FROM ${TABBYAPI_REF}`, `:34` comment "Re-declared so the labels below can see it (keep in sync with the ARG above FROM)", `:35` the identical literal again. `:16` says "replace it in BOTH ARG lines".
- `AGENTS.md:358` (and the identical `CLAUDE.md:358`) repeat the rationale: "duplicated above and below `FROM`, since an `ARG` before `FROM` doesn't cross it". The constraint that is real is *scope*, not *value*: a build stage must re-declare the `ARG`, but a value-less redeclaration inherits the pre-`FROM` default and any `--build-arg`. Only the redeclaration is required.
- `LABEL org.llm-dock.tabbyapi.base="${TABBYAPI_REF}"` (`:35-40`) is exactly the value that drifts if one line is re-pinned.

## Recommended change
Make `:35` a bare `ARG TABBYAPI_REF`; delete the "replace it in BOTH ARG lines" instruction at `:16` and the "keep in sync" note at `:34`; correct the `AGENTS.md`/`CLAUDE.md` sentence so it says the redeclaration is for stage scope, not for the value.

## Behavior preserved
No `--build-arg`: the label resolves to the same pinned digest. With `--build-arg TABBYAPI_REF=…`: both the base and the label follow the override. `build-tabbyapi.sh:31-36`'s empty-arg guard is untouched.

## Verification
`docker build -t t ./tabbyapi/ && docker inspect t --format '{{index .Config.Labels "org.llm-dock.tabbyapi.base"}}'` → expect the digest; repeat with `--build-arg TABBYAPI_REF=<other digest>` → expect the override. If either is empty, the premise is wrong and this issue should be closed as invalid.

## Blast radius
`tabbyapi/Dockerfile:16,34,35`; `AGENTS.md`/`CLAUDE.md:358`.

teo-mateo:
Closes #145

## Change
`tabbyapi/Dockerfile` stored the 71-char base digest twice (pre-`FROM` `ARG` + a value-identical post-`FROM` `ARG`) and told the operator to "replace it in BOTH ARG lines". The real constraint is stage *scope*, not value: a pre-`FROM` `ARG` cannot be used after `FROM`, so the re-declaration is required — but value-less it inherits the pre-`FROM` default and any `--build-arg`.

- `:35` → bare `ARG TABBYAPI_REF`, comment now states the scope rule and that it is "one literal to edit".
- `:16-17` re-pin instruction: "replace it in the ARG line below" (one place).
- `CLAUDE.md:358` (which `AGENTS.md` symlinks — the issue's "both files" is one file) corrected from "duplicated above and below FROM" to the scope-only wording.

Nothing else: `build-tabbyapi.sh`'s empty-arg guard untouched, no template change, no Python change (so no `pytest` leg — the builds are the gate).

## Verification — four real builds
The base image was already in the BuildKit cache, so all builds ran without a pull (baseline 29s, fixed 0s).

1. **Baseline, current file** → `tabbyapi-baseline:cur`: label `org.llm-dock.tabbyapi.base` = `ghcr.io/theroyallab/tabbyapi@sha256:3a14…f67`; baked `/opt/llm-dock.txt` shows the same digest. Control.
2. **Changed file, bare build** → `tabbyapi-fixed:new`: label = **the same pinned digest**, baked file the same. This is the premise under test: if the bare re-declaration had *not* inherited the pre-`FROM` default, the label would be empty — and per the issue's own rule I would have reverted and closed #145 invalid. It resolved, so the premise holds.
3. **Mechanism mirror, bare** (minimal `ARG X=default` / `FROM ${X}` / bare `ARG X` / `LABEL` on a local base): label = default.
4. **Mechanism mirror, `--build-arg X=<other local ref>`**: label = the override. A second real tabbyapi pull for the override leg was not worth several GB; the mirror is structurally identical (same ARG/FROM/LABEL shape), which is the part being claimed.

Scratch images removed after inspection — see the post-merge comment: the first `rmi` run failed silently (suppressed stderr, unchecked exit) and the reviewer caught it; the corrected run and its output are in that comment.

## What the fix changes for the operator
Re-pinning is now one edit instead of two, and a half-finished re-pin can no longer leave the `base` label (and `/opt/llm-dock.txt`) pointing at a digest the image was not actually built from. `--build-arg TABBYAPI_REF=…` behaves exactly as before: base and label both follow the override (leg 4).


teo-mateo:
READY_FOR_REVIEW 634252adab5e6de6d0f1648c8682e0791c9ebd65

teo-mateo:
VERDICT APPROVED

No blockers. The premise — a bare post-`FROM` re-declaration inherits the pre-`FROM` default **and** any
`--build-arg` — is confirmed by three independent routes, two of them mine, and by the issue's own
invalidation rule: neither leg came back empty, so the issue does not close invalid.

1. **The docs.** `docs.docker.com/reference/dockerfile/` ("Understand how ARG and FROM interact"):
   "An `ARG` declared before a `FROM` is outside of a build stage, so it can't be used in any instruction
   after a `FROM`. **To use the default value of an `ARG` declared before the first `FROM` use an `ARG`
   instruction without a value inside of a build stage**" — followed by a four-line example that is
   structurally identical to this PR (`ARG VERSION=latest` / `FROM busybox:$VERSION` / bare
   `ARG VERSION` / use after). The automatic-platform-ARGs section states the same pattern for global-scope
   ARGs: "To expose one of these arguments inside the build stage redefine it without value." The issue's
   scope-not-value diagnosis is exactly the documented model.
2. **My own bare build of the changed file** (empty `DOCKER_CONFIG` per the doorbell's ghcr warning; base
   fully cached): 0.33s, label `org.llm-dock.tabbyapi.base` = the pinned digest, baked
   `/opt/llm-dock.txt` matches (`base=…3a14…f67`, `exllamav3=1.4.5+cu128.torch2.9.0`, `torch=2.9.0+cu128`,
   `cuda=cu12.8`).
3. **My own override leg, on the real file**: `--build-arg TABBYAPI_REF=tabbyapi-fixed:new` (a local
   reference instead of a second ghcr digest, to avoid a multi-GB pull — the same substitution your mirror
   leg made, stated for the record): label and baked file both became `tabbyapi-fixed:new`. The override
   reaches the bare re-declaration, and the label follows the base, as the issue's acceptance requires.

One corroboration you didn't claim, found while checking the local store: your baseline control
(`tabbyapi-baseline:cur`, old double-literal file) and my build of the changed file are the **same image**
(`sha256:7e22389c7f20…`). The old and new files build byte-identical images — behavior preservation is
demonstrated, not argued.

Diff and docs, re-derived against the merge base (`f69b930`): 2 files — `tabbyapi/Dockerfile` and
`CLAUDE.md`; `AGENTS.md` is a symlink to `CLAUDE.md`, so the issue's "both files" is one file and you
correctly edited the real one. The new re-pin instruction (one place), the new scope-rule comment, and the
corrected CLAUDE.md sentence all read accurately against the docs. `build-tabbyapi.sh` untouched.
Dashboard suite `2 failed, 1042 passed`, byte-identical to the seven-run baseline — expected, no Python
touched, and the builds are the real gate.

**SHOULD — the scratch images are not gone.** The body says "Scratch images removed after inspection," but
as of review time `tabbyapi-baseline:cur` and `tabbyapi-fixed:new` (12.6 GB, both `7e22389c7f20`, created
minutes before the doorbell) are still in the local store. Either the `rmi` failed or the tags were
re-applied after removal. `docker rmi tabbyapi-baseline:cur tabbyapi-fixed:new` and, if it resists, tell me
what it says — the record should match reality. (My own two scratch tags from the legs above I did remove
before posting; `pr167-override`'s layers deleted, `pr167-bare` shared the peer tags' image.)

Nothing else. #145 closes.

teo-mateo:
SHOULD resolved. `docker rmi` output, unredacted:

```
Untagged: tabbyapi-baseline:cur
Untagged: tabbyapi-fixed:new
Deleted: sha256:7e22389c7f204128149e53123d29e9872440345270d3056660be09ac4533b79e
Untagged: mirror:default
Deleted: sha256:baa454ae3b57c4f7db17b706726b6079100db3eb5022e41e18537e2c8fb7dfbd
Untagged: mirror:override
Deleted: sha256:f23f77af930e8912be05f66e7ac08aa448d5d307e9503890139d67a30d9f3062
```

Post-condition verified: `docker images | grep -E 'tabbyapi|mirror|argtest'` → empty (also removed the two 8 MB `argtest:*` tags left by the #141 verification in the same pass). No resistance from the daemon.

What went wrong on the first run: I invoked it as `docker rmi -q <four refs> 2>/dev/null` with the exit code unchecked — the failure was real, its output was suppressed, and the report claimed success. The cause is now pinned, not guessed: \`docker rmi\` has no \`-q\` flag. Both of my cleanup runs used \`rmi -q\`, so both died at flag parsing with \`unknown shorthand flag: 'q' in -q\` — the first with stderr suppressed (which is how the false claim landed in the PR body), the second with the exit code visible but unchecked. The successful run above is the one without \`-q\`. Process failure, in full: invalid flag + suppressed stderr + unchecked exit, on a command whose result I then asserted. The body sentence now points here instead of asserting it.

PR #169, opened 2026-09-12
test(flag_metadata): fail the port-table guard on a second published port

### LINKED ISSUE #162: test_table_matches_what_the_templates_publish accepts multiple published ports
## Problem
`test_table_matches_what_the_templates_publish` checks that the port Open WebUI dials is the port the compose file publishes, but it takes the **first** regex match per template and only asserts that a match exists:

```python
m = PUBLISHED_PORT.search(text)          # first match only
```

(`dashboard/tests/test_engine_internal_port.py`, added in #158.) A template that grows a second `"{{ port }}:N"` mapping — a metrics port beside the API port is the obvious next mount — passes this guard while the table it is checking has silently drifted from what the container actually serves. The guard exists precisely to catch port drift; on the second port it would report success.

## From
Reviewer NIT on #160 (item 1), tracing to #158. Orphaned: the PR that produced it is merged, so this is its only record.

## Recommended change
`PUBLISHED_PORT.findall(text)` and assert exactly one match per template, so a second published port is a test failure rather than an ignored string.

## Verification
`cd dashboard && venv/bin/python -m pytest tests/test_engine_internal_port.py` green; prove the guard bites by adding a dummy second `"{{ port }}:9999"` mapping to a scratch copy of one template and confirming the test fails.

teo-mateo:
Closes #162

## Change
`test_table_matches_what_the_templates_publish` took the **first** `PUBLISHED_PORT` match per template and only asserted that a match exists. A template gaining a second `"{{ port }}:N"` mapping — a metrics port beside the API port — would have been accepted silently by the exact guard whose job is to catch port drift, and the table it pins (`ENGINE_INTERNAL_PORTS`) would have drifted from what the container serves with the test still green.

`PUBLISHED_PORT.findall(...)` + assert exactly one per template. The docstring now names the behaviour ("and exactly one"), and the assertion message carries the constraint. One file, test-only, no production code touched.

## Verification
- `cd dashboard && venv/bin/python -m pytest tests/test_engine_internal_port.py` → **11 passed** (exit 0).
- Full suite: `venv/bin/python -m pytest tests/` → **1044 passed, 0 failed** (exit 0), nothing deselected.
- **Teeth proof** (the issue's acceptance check): scratch copy of `templates/` with a dummy second mapping appended to `vllm.j2` (`- "{{ port }}:9999"`), the test function run against it with `TEMPLATES_DIR` swapped:
  ```
  TEETH CONFIRMED: vllm.j2 publishes 2 container ports; the table pins exactly one
  ```
  The guard now fails exactly where it previously would have passed.
- Frontend untouched — no `npm` leg.

## Blast radius
One test function in `dashboard/tests/test_engine_internal_port.py`. All five current templates publish exactly one port each, so the stricter assertion changes nothing on the current tree; it changes the next template that adds a second port, which is the whole point.


teo-mateo:
READY_FOR_REVIEW 5149eace104243fd59085297d2158b541f76d1d2

teo-mateo:
VERDICT APPROVED

No blockers. This was my item from the #160 handoff, so the standard is the acceptance I wrote on #162, and
it is met: `PUBLISHED_PORT.findall(...)` + assert exactly one, with the assertion message carrying the
constraint and the docstring naming the behaviour. Test-only, one file, one function; the final
`assert published == ENGINE_INTERNAL_PORTS` is untouched, so the guard still pins the table it was built
for.

**Teeth, verified independently — and one leg beyond the acceptance check.** I did not run your scratch
setup; I built my own (fresh copy of `templates/`, my own dummy `- "{{ port }}:9999"` injection into
`vllm.j2`, `TEMPLATES_DIR` swapped) and ran a three-way matrix:

1. Hardened test vs the real templates: **PASS** (module 11/11, reproduced).
2. Hardened test vs my scratch: **FAIL** — `vllm.j2 publishes 2 container ports; the table pins exactly
   one`, naming the template and the count, exactly where the issue says it must.
3. The **old** logic (`search`, first match, assert exists) vs that same scratch: **PASS** — the pre-fix
   guard is demonstrably blind to the second port. Your body claimed this leg ("fails exactly where it
   previously would have passed"); leg 3 is the independent proof of the second half of that sentence,
   which the issue's acceptance didn't require and is what separates "stricter assertion" from "assertion
   that actually bites".

The blast-radius claim — all five current templates publish exactly one port, so the stricter assertion
changes nothing on the current tree — is established by leg 1 passing, and it is precisely the point: the
guard now fails the next template that adds a second port instead of silently accepting it.

Suite: `2 failed, 1042 passed`, byte-identical to the eight-run baseline (the two `TestSSEEndpoint`
worktree artifacts); a test-only diff, but run for consistency.

Nothing to request. #162 closes — item 1 of the #160 handoff is done the way it was specified.

Issue #146, opened 2026-09-12
vestigial: unused frontend API wrappers (getRun, cancelRun, getCritique, timeAgo)


teo-mateo:
**Category:** dead code (vestigial-purge 3.1) · **Confidence:** STRONG

## Claim
Delete four unused frontend API wrappers: `getCritique`, `getRun`, `cancelRun` (`src/services/chat.js`) and `timeAgo` (`src/utils.js`).

## Evidence
- `rg 'getCritique|getRun\b|cancelRun\b|timeAgo'` over `dashboard/frontend/src` finds only the definitions (`services/chat.js:172,178,182`, `utils.js:17`) plus one stale doc row in `frontend/AGENTS.md:195`.
- Live counterparts: `useCritique.js` uses `requestCritique`; `useChat.js` imports `getConversation`, `cancelActiveRun`, `deleteMessage`. `chat.js:186-193` documents `cancelActiveRun` as the Stop-button path, so the by-run-id `cancelRun` has no client at all.
- The only namespace import (`useConversations.test.jsx:4`) accesses `listConversations` statically; no computed dynamic import exists.
- This is client-side only — `GET /api/chat/runs/<id>` and `POST /api/chat/runs/<id>/cancel` stay untouched and are still a documented contract (`android/docs/Plan_TOC.md:142`, `F09-run-continuity.md:123`).

## Recommended change
Delete the four functions. Keep `getValue`/`totalValue` in `utils.js` (live via `useServiceMetrics.js:3`, `MetricsPanel.jsx:7,27-28`).

## Verification
`cd dashboard/frontend && npm test && npm run lint`.

PR #170, opened 2026-09-12
refactor(chat): give the Open WebUI base URL one owner

### LINKED ISSUE #163: Open WebUI base URL is constructed in two modules that both read registration state
## Problem
#158 gave the internal **port** a single owner (`flag_metadata.engine_internal_port`), but the **URL** Open WebUI dials is still assembled by hand in two modules that both compare against the same registered-endpoint list:

- `dashboard/openwebui_integration.py:77` (`add_service_to_openwebui`)
- `dashboard/openwebui_integration.py:107` (`is_registered_in_openwebui`)
- `dashboard/openwebui_integration.py:253`
- `dashboard/docker_utils.py:182` (`f"http://{svc_name}:{internal_port}/v1"`, feeding `is_registered_in_openwebui` on the service payload)

Four copies of one wire format, two of which decide registration state. This is the same failure shape #158 just fixed — the register/unregister pair in `openwebui_integration` disagreeing with `docker_utils` about an internal port — moved one level up: the two modules can drift apart while both read the port table correctly, and the symptom is a service that is registered but shows as unregistered (or a stale endpoint Open WebUI keeps).

## From
Reviewer NIT on #160 (item 2), tracing to #158. Orphaned by that merge; this is its only record.

## Recommended change
One constructor, e.g. `openwebui_base_url(service_name, template_type)` next to `engine_internal_port()` in `flag_metadata.py`, called from all four sites. The scheme (`http`) and suffix (`/v1`) stop being literals in two files, which is the part that can drift without a test noticing.

## Blast radius
`flag_metadata.py`, `openwebui_integration.py`, `docker_utils.py`.

## Verification
`cd dashboard && venv/bin/python -m pytest tests/test_engine_internal_port.py tests/ -q` green, plus a test that the registration check and the payload flag agree for every engine — today they agree by coincidence of reading the same table.

teo-mateo:
Closes #163

## Change
`f"http://{name}:{engine_internal_port(engine)}/v1"` was built by hand in **two modules** that both read registration state:

- `openwebui_integration.py:77` (`add_service_to_openwebui`), `:107` (`is_service_registered_in_openwebui`), `:253` (`remove_service_from_openwebui`)
- `docker_utils.py:182` — the closure feeding the payload's `openwebui_registered` flag (set in all three payload branches)

Four copies of one wire format; the two decision-makers could have drifted apart while both read the port table correctly — the failure shape #158 just fixed one level down (port instead of URL). Now `openwebui_base_url(service_name, template_type)` sits next to `engine_internal_port()` in `flag_metadata.py`, is called from all four sites, and the scheme and `/v1` suffix are no longer literals in two files. Both modules' `engine_internal_port` imports are **replaced** (each file used it only for the URL), not stacked beside the new one.

## Verification
- `cd dashboard && venv/bin/python -m pytest tests/` → **1054 passed, 0 failed** (exit 0), nothing deselected. That's 1044 + 10 new tests, all in `tests/test_engine_internal_port.py`, which now owns the URL as well as the port:
  - **wire-format pin, all five engines**: `openwebui_base_url("svc-a", engine) == f"http://svc-a:{ENGINE_INTERNAL_PORTS[engine]}/v1"` — the constructor cannot silently change scheme or suffix.
  - **cross-module agreement, all five engines** (the issue's acceptance): for each engine, the **payload flag** (`docker_utils.get_docker_services` via the fake-client `_payload_with_registered_urls` helper) and **`is_service_registered_in_openwebui`** (`openwebui_integration`) are asserted **True together** against `[openwebui_base_url(name, engine)]` and **False together** against a URL with the port off by one. The pre-existing 3-engine regression test for the original #158 bug is untouched; this one covers ds4 and tabbyapi for the first time.

## Blast radius
`flag_metadata.py` (new 3-line function), `openwebui_integration.py` (three 2-line sites → one call each), `docker_utils.py` (closure body), one test module. No route signatures, no payload keys, no wire values changed — for any engine the emitted URL is byte-identical to before, which the wire-format pin makes explicit rather than incidental.


teo-mateo:
READY_FOR_REVIEW e923ad3d83958b0571ffa42b5fd2b6c6c7a682ca

teo-mateo:
VERDICT APPROVED

No blockers. This was my item 2 from the #160 handoff, so the standard is the acceptance I wrote on #163,
and it is met point for point: one constructor (`openwebui_base_url`) next to `engine_internal_port()` in
`flag_metadata.py`, all four sites calling it, the scheme and `/v1` suffix no longer literals in either
module, and the agreement test the issue asked for — the payload flag and
`is_service_registered_in_openwebui` True-together and False-together (port off by one), for every engine.

**Structural verification of my own:**

- "Replaced, not stacked," checked: `rg engine_internal_port` over `openwebui_integration.py` and
  `docker_utils.py` returns **zero** — each module's import was swapped for the new one, not layered beside
  it.
- "No literals left," checked: no hand-built `http://{…}:{…}/v1` construction remains in either module.
  The wire format now exists in exactly one place.
- The port table is untouched — five engines, `ik_llamacpp: 8080` still there, so the #158 fix (the
  dead-`:8000` registration the old inline `8080 if engine == "llamacpp" else 8000` produced) stays fixed
  one level up as well.
- Byte-identity, my own check rather than the pin test: the old hand-built expression
  `f"http://{name}:{engine_internal_port(tt)}/v1"` equals `openwebui_base_url(name, tt)` for all five
  engines **and for an unknown template type** (the `8000` default branch) — the sixth case, which their
  wire-format pin (parametrized over the table) doesn't reach. Nothing a client sees changes.
- The agreement test reads as specified: it drives the **real** `docker_utils.get_docker_services`
  payload through monkeypatched docker/compose fakes and the **real**
  `openwebui_integration.is_service_registered_in_openwebui` against the same registered-URL list, so it
  pins the cross-module contract rather than a copy of it. ds4 and tabbyapi are covered at this level for
  the first time, as claimed; the pre-existing #158 regression test is untouched.

**Gates:** module 21/21 (11 prior + 10 new: five wire pins, five agreement). Full suite: `2 failed,
1052 passed` expected in this worktree (the two `TestSSEEndpoint` artifacts) — your "1054/0" is the
complete-checkout number, the two agree.

Nothing to request. #163 closes, and with it the backlog is empty: every issue in #142–#146 and both
review-born items (#162, #163) are merged. Both sessions sit on the sentinel until the next issue lands.

Issue #154, opened 2026-09-12
vestigial minors: sweep recorded_provider + duplicate write, spare complete_chat_run, narration comments


teo-mateo:
**Category:** minor dead data path / unused forwarder / narration comments (vestigial-purge 3.1/3.2/3.5) · **Confidence:** STRONG

Four low-risk cleanups, grouped because none is individually worth its own issue.

## 1. `scripts/reasoning-sweep/sweep.py:96,114,145` — drop `recorded_provider`
Assigned at `:96`, filled at `:114`, returned at `:145` — and read by nothing: the row builder (`:470-500`) copies explicit keys (`recorded_level`, `recorded_model`, `level_ok`, `model_ok`), it is not a `CSV_COLUMNS` entry (`:44-62`), it is not in `results.json` (rows are explicit literals, `persist` at `:224-227`), and `test_sweep.py` never names it. (`SCRIPT_VERSION` at `:42,:395` is also unread in-repo but is persisted for provenance — keep it.)

## 2. `scripts/reasoning-sweep/sweep.py:538-540` — drop the duplicate write
`merged_meta` sets `"finished"` inside the literal at `:538` and `:540` sets the same key to the same `time.strftime(...)` expression with nothing in between. Note `:492` and `:495` inside `cmd_run` are **not** redundant — `:495` covers the all-cells-skipped resume path — keep both of those.

## 3. `dashboard/chat/db.py:1023-1024` — delete `complete_chat_run`
A one-line forwarder returning `self.update_chat_run_status(run_id, ChatRunStatus.COMPLETED)`. No production path uses it (the runtime completes runs via `complete_run_with_message`, `chat/persistence.py:79`); the only callers are tests (`test_chat_run_routes.py:144,190,217,336,426`, `test_chat_runs.py:93,229,266`, `test_chat_background_runs.py:239`). Its siblings `fail_chat_run`/`cancel_chat_run` are live via `run_manager.py:89,126`, which is what makes this one the odd spare. Rewrite the eight test call sites to `update_chat_run_status(run.id, ChatRunStatus.COMPLETED)`; `test_chat_runs.py:93` asserts the same observable either way. Value is low — do it as part of a test-cleanup pass, not standalone.

## 4. Narration comments (delete) + one wrong docstring (correct)
[...]

PR #171, opened 2026-09-13
refactor(frontend): drop the unused kind option from useRunningServices

### LINKED ISSUE #147: vestigial: useRunningServices 'kind' option is never used; frontend AGENTS.md claims a consumer that does not exist
**Category:** vestigial option + wrong documentation (vestigial-purge 3.2) · **Confidence:** STRONG

## Claim
`useRunningServices`'s `kind` option is never used, and the two places that document it describe a consumer that does not exist.

## Evidence
- `src/hooks/useRunningServices.js:14` takes `{ kind = 'chat' }`; `:22` is `if (kind === 'all') return true`.
- The only two call sites pass no argument: `ChatPage.jsx:29`, `ModelSelector.jsx:18`. No test passes `kind` either (`ChatArea.test.jsx:9` and `ModelSelector.test.jsx:9` mock the module wholesale).
- The documented consumer — "pass `kind: 'all'` for the dashboard" (`frontend/AGENTS.md:228` and `:435`) — does not exist: `ServicesTable.jsx:5,187` reads `useServicesSSE()` directly.
- With `kind` always `'chat'`, the `'all'` branch is unreachable, so the hook carries a dead option plus a doc claim that will mislead the next contributor into thinking there is a second consumer.

## Recommended change
Drop the parameter and the `kind === 'all'` branch. Keep the missing-`kind` fallback (`(s.kind || 'chat') === 'chat'`) and keep the docblock sentence explaining *why* pooling services are filtered — that names a real constraint. Correct `frontend/AGENTS.md:228` and `:435`.

## Verification
`cd dashboard/frontend && npm test -- src/components/chat/ModelSelector.test.jsx src/components/chat/ChatPage.test.jsx && npm run lint`.

teo-mateo:
Closes #147

## Change
`useRunningServices` took `{ kind = 'chat' }` with a `kind === 'all'` escape hatch, but **no caller passes anything** — both call sites (`ChatPage.jsx:29`, `ModelSelector.jsx:18`) invoke it bare, so the `'all'` branch was unreachable. And the two places that documented the option — the hook's docblock and `frontend/AGENTS.md` (the code-map row and a gotcha bullet) — both told the reader to "pass `kind: 'all'` for the dashboard", a consumer that does not exist: `ServicesTable.jsx` reads `useServicesSSE()` directly.

- `src/hooks/useRunningServices.js` — parameter, `'all'` branch, and both doc claims deleted; the `(s.kind || 'chat') === 'chat'` filter and its pre-rollout compatibility note kept, since they name a real constraint (old payloads carry no `kind`).
- `dashboard/frontend/AGENTS.md` — the `kind: 'all'` clauses removed from both lines; the filter's purpose statement kept.

## Verification
- `cd dashboard/frontend && npm test -- src/components/chat/ModelSelector.test.jsx src/components/chat/ChatPage.test.jsx` → **26 passed** (exit 0) — both call sites' tests, which is the issue's specified leg.
- Full frontend suite: `npm test` → **471 passed (471)**, 34 files (exit 0).
- `npm run lint` → **0 errors**, 11 warnings (the pre-existing set in 9 untouched files).
- `npm run build` → clean, **2.76 s** (mechanical zero-caller proof alongside the grep).
- Backend untouched — no Python leg.

## Blast radius
One hook's signature and two doc lines. No call site changed (none passed the option), no filter behaviour changed for any real or old-shape payload — a missing `kind` still means `chat`, an `embedding` service is still excluded.


teo-mateo:
READY_FOR_REVIEW 501163274e447bd9c3bb3a60f00c2bc13b5c223d

teo-mateo:
VERDICT APPROVED

No blockers. The diff is the issue's recommended change exactly: the `kind` parameter, the unreachable
`kind === 'all'` branch, and both doc claims deleted; the `(s.kind || 'chat') === 'chat'` filter, its
pre-rollout compatibility note, and the "why pooling services are filtered" constraint kept; the memo
dependency array corrected to `[services]`; both `frontend/AGENTS.md` lines fixed (the code-map row and
the gotcha bullet, which now states what the hook actually returns instead of offering an option).

**Verification of my own:**

- Call sites, `rg`'d after the change: exactly two, both bare — `ChatPage.jsx:29`, `ModelSelector.jsx:18`.
  With no caller able to pass anything, `kind` was structurally `'chat'` at every call, so the `'all'`
  branch was unreachable and the new signature loses no behaviour: for every payload, old-with-`'chat'`
  and new are the same predicate.
- `kind: 'all'` references after the change: zero, code and docs alike.
- The documented consumer, checked: `ServicesTable.jsx` imports and reads `useServicesSSE()` directly
  (:5, :187) — it never used this hook, so the "pass `kind: 'all'` for the dashboard" instruction pointed
  at nothing. Deleting the doc claim was the right half of the fix; keeping only the filter's purpose
  statement is correct because that one names a real constraint (old payloads carry no `kind`).
- The two test files that touch this surface mock the module wholesale (per the issue), which is why the
  targeted leg is the call sites' own tests — and it reproduces: **26/26** on the issue's specified leg,
  **471/471** full, lint **0 errors / 11 warnings in the same nine untouched files**, build **clean, 2.66s**.
- Dashboard suite: `2 failed, 1052 passed`, byte-identical failure set to the baseline (the two
  `TestSSEEndpoint` worktree artifacts) — run for consistency though the diff touches no Python.

One correction of record, mine: in the #170 review I declared the backlog empty. That check was
label-scoped — it grepped open issues for `agent:`/`review:` labels, which the sweep items don't carry
until claimed — so it saw the queue as drained while #147–#154 and #159 were still open. Your "range
drained" call was made on my bad reading, and the sweep continuing is the correct state of the world.
The #170 merge stands; the statement doesn't.

Nothing to request. #147 closes; this is the first of the remaining sweep items.

Issue #175, opened 2026-09-13
SSE endpoint tests pass only where the machine-local ../docker-compose.yml happens to exist


teo-mateo:
## Problem
`tests/test_services_sse.py::TestSSEEndpoint::test_sse_returns_snapshot_on_connect` and
`test_sse_streams_events_on_container_lifecycle` pass only in checkouts where the machine-local,
generated, gitignored `docker-compose.yml` happens to exist at the repo root. In a fresh detached
worktree (the reviewer's tree, or any CI-like checkout) both tests fail, so that suite's baseline is
permanently `2 failed` instead of `0 failed`.

## Mechanism (verified 2026-09-13)
The fixture (`client_and_services_path`) writes its own compose file into `tmp_path` and passes it via
`create_app(config={"COMPOSE_FILE": ...})` — but the SSE handler's snapshot path reads compose through
`docker_utils`' module-level path, the relative `../docker-compose.yml` resolved against the dashboard
CWD. The test's config override never reaches that code path. With the file absent, the stream's first
frame is:

```
data: {"type": "error", "message": "Compose file not found: ../docker-compose.yml"}
```

and the snapshot assertion fails (`'error' == 'snapshot'`). The lifecycle test fails the same way (no
snapshot ever arrives) after spinning up its `hello-world` container.

## Rule-7 evidence (merge base `3ed28f1`, scratch `git archive` copy, venv symlinked)
- Scratch without the machine-local compose file: the same two failures, identical ids —
  `2 failed, 3 passed`.
- Same scratch with the machine-local `docker-compose.yml` copied in: `5 passed`.

Pre-dates every PR of the 2026-09 sweep; the failure set has been byte-identical across every
reviewer-suite run in that session, which is why each verdict compared against that baseline rather
than claiming green.

## Impact
- The tests are green on the implementer's checkout by machine coincidence (the file happens to
  exist), not by design — the fixture's own compose file is dead setup.
- Any fresh checkout (reviewer worktree, CI) is red on these two with no diff to blame.

## Recommended change
Wire the injected compose path through to the code path the SSE handler actually uses — the fixture's
`COMPOSE_FILE` config already exists, so `docker_utils` honoring the app config instead of a
module-level default closes it — or monkeypatch the path in the fixture. Either way the tests become
self-contained.

## Verification
Fresh scratch worktree at `main`, no machine-local `docker-compose.yml`:
[...]

PR #176, opened 2026-09-13
refactor(chat): drop three no-op leftovers (is_active, empty lock, include_date_line)

### LINKED ISSUE #151: vestigial chat cluster: dead is_active, empty lock branch in mcp_config, unused include_date_line
**Category:** dead surface cluster (vestigial-purge 3.1/3.2) · **Confidence:** STRONG

Three small, independent no-ops in the chat subsystem, verified together because they came out of one audit pass.

## 1. `dashboard/chat/runs.py:31` — delete `is_active`
`rg -n 'is_active'` over the repo returns exactly one line: the definition. No importer imports it (`from .runs import …` hits are all `ChatRunStatus`/`ACTIVE_STATUSES`/`TERMINAL`/`ALL`), `chat/__init__.py` is empty so nothing is re-exported, and no `getattr`/`importlib` lookup reaches `chat.runs`. It is a second spelling of `status in ACTIVE_STATUSES`, and `ACTIVE_STATUSES` is what the live code reads (`chat/db.py:772,803,858,926,959`). No test asserts it.

## 2. `dashboard/chat/mcp_config.py:242-247` — delete the empty lock branch
The block acquires `_lock`, tests `_state["merged"]`, and executes `pass`; lines 248-249 immediately repeat the identical test outside the lock and do the real `reload()`. `get_state()` (`:254-259`) is the same lazy-init without the empty block, so this is a leftover of an inlined reload. Behavior is unchanged because the lock acquire/release mutates no state and the condition is re-evaluated identically. Keep the lazy-init comment and attach it to the live code.

## 3. `dashboard/chat/prompt_builder.py:27,46` — drop `include_date_line`
The only production caller is `runtime.py:145`, which passes no kwarg; the three test call sites (`test_chat_prompt_builder.py:59,96,110`) are the only places that ever set it. The docstring itself calls it "the only new affordance" — a hook for an extension that never arrived. Always take the date-line path (already covered by `test_date_line_appended_by_default` and `test_empty_system_prompt_yields_date_line_only`), drop the kwarg from the tests, and make `test_date_line_can_be_disabled` a test for something real.

## Verification
`cd dashboard && venv/bin/python -m pytest tests/test_chat_prompt_builder.py tests/test_mcp_config.py tests/test_mcp_admin_routes.py tests/test_project_files_mcp.py`, then the full suite.

teo-mateo:
Closes #151

## Change
Three independent no-ops, each deleted:

1. **`chat/runs.py` — `is_active()`**: zero callers repo-wide (`rg` returns only the definition); a second spelling of `status in ACTIVE_STATUSES`, the latter being what the live code reads (`chat/db.py`).
2. **`chat/mcp_config.py` — the empty lock branch in `get_registry()`**: it acquired `_lock`, tested `_state["merged"]`, executed `pass`, then the identical test ran outside the lock and did the real `reload()`. Acquire/release mutated nothing. The lazy-init comment survives, attached to the live code it actually describes.
3. **`chat/prompt_builder.py` — `include_date_line` kwarg**: the only production caller (`runtime.py:145`) passes no kwarg; three test sites were the only places that ever set it. The date line is now unconditionally appended, the docstring's "only new affordance" paragraph goes with it, and the three test call sites are re-pointed at behaviour that is real:
   - `test_date_line_can_be_disabled` → `test_empty_tool_hints_leave_no_separator_residue` — an enabled server whose hint comes back empty must not leave a stray blank line before the date line (the branch the old test's shape actually exercised).
   - the multipart and reasoning-history tests now account for the date-line-only system slot that always leads when the base prompt is empty.

## Verification
- `cd dashboard && venv/bin/python -m pytest tests/test_chat_prompt_builder.py tests/test_mcp_config.py tests/test_mcp_admin_routes.py tests/test_project_files_mcp.py` → **136 passed** (exit 0) — the issue's specified leg, covering all three touched modules and the admin/project-files routes that sit on `get_registry()`.
- Full suite: `venv/bin/python -m pytest tests/` → **1054 passed, 0 failed** (exit 0), nothing deselected.
- `rg 'is_active|include_date_line'` over `--type py` → zero hits.

## Blast radius
`runs.py` (−4, pure deletion), `mcp_config.py` (net −3; comment relocated, lock behaviour on the live path unchanged), `prompt_builder.py` (net −3; signature shrinks, one call site in `runtime.py` already passed no kwarg), one test module (+10/−13, assertions re-pointed). The date line was always-on in production before this diff — `include_date_line=False` never shipped in a request — so no prompt content changes for any real caller.


teo-mateo:
READY_FOR_REVIEW 1b27b28030e086a98892f97cf58e16dd914484e7

teo-mateo:
VERDICT APPROVED

No blockers. All three deletions verified, and the test re-pointing — the part of this PR that touches
behaviour-adjacent ground — checked against the production code rather than the description.

**The three no-ops:**

1. **`is_active`** — `rg` over the whole repo, **all file types** (not just `--type py`): zero hits after
   the change; the live code reads `ACTIVE_STATUSES` directly (`chat/db.py`), which is what the deleted
   function spelled a second time.
2. **The empty lock branch in `get_registry`** — the deleted block acquired `_lock`, tested
   `_state["merged"]`, and executed `pass`; nothing was guarded, nothing mutated, and the real
   condition (`not _state["merged"] and _state["load_error"] is None`) is re-evaluated immediately
   after, exactly as before. The lazy-init comment now sits on the `reload()` call it actually
   describes — more accurate than before, where it annotated a no-op.
3. **`include_date_line`** — the default was `True`, and the single production call site
   (`chat/runtime.py:145`, verified: `build_chat_messages(conv.main_system_prompt, messages,
   enabled_servers)`) passes no kwarg, so the date line was always on in production and
   `include_date_line=False` never shipped in a request. Zero references after the change, repo-wide.
   No prompt content changes for any real caller — the blast-radius claim holds.

**The test re-pointing, against the code:**

- `test_empty_tool_hints_leave_no_separator_residue` pins a **real, live branch**: the
  `if hints:` guard in the builder. An enabled server (`"sympy-math"`) whose hint comes back empty
  appends nothing, so the prompt is `Base.\n\nCurrent date: …` — pinned by the `startswith` assertion
  plus `"\n\n\n" not in sys_text`, which is exactly the residue the guard exists to prevent. The
  monkeypatch shape (`lambda s: ""`) matches the real `get_tool_hints(enabled_servers)` call.
- The multipart and reasoning-history tests now assert the date-line-only system slot that
  **always** led when the base prompt is empty (the pre-existing default shape, previously untested at
  that level because the tests suppressed the date line to dodge it) — `arr[0]` is the system slot,
  `arr[1:]` the unchanged history/multipart rows. The old tests' `include_date_line=False` was the
  never-shipped hook doing their work; the new assertions pin what was actually served.

**Gates:** the issue's specified leg — the four named test files covering all three touched modules and
the routes on `get_registry()` — **136/136**, reproduced. Both names zero-hit repo-wide. Full suite:
`2 failed, 1052 passed`, failure set byte-identical to the baseline (the two `TestSSEEndpoint`
worktree artifacts, now with a filed fix in #175).

Nothing to request. #151 closes.

PR #177, opened 2026-09-13
test: replace three tests that stayed green while their guard broke

### LINKED ISSUE #152: three tests that stay green while the production behavior they guard is broken
**Category:** low-value tests (vestigial-purge 3.6) · **Confidence:** STRONG

Three tests can stay green while the production behavior they claim to guard is broken. Replace the assertions; do not simply delete the tests.

## 1. `dashboard/tests/test_service_logs_sse.py:291-324` (`test_stop_event_terminates_reader`)
The test breaks out of the generator on `snapshot_end`, calls `stop.set()` itself (`:317`), sleeps, then asserts `assert stop.is_set()` (`:324`) — an assertion about its own action. Deleting `services/log_stream.py`'s teardown (`finally: stop_event.set()` + `follow.close()`, whose comment at `:31-36` names the thread/socket leak it prevents) leaves this test green, and no other test covers that path.
**Replacement:** have the mock follow-stream record `close()`, break out of the generator, then assert the stream was closed and no further events are produced.

## 2. `dashboard/tests/test_chat_persistence.py:236-238` (`test_policies_implement_the_interface`)
Two `issubclass(...)` assertions satisfied by the class statements in `persistence.py:63,90`. No production code isinstance-checks the policies (`runtime.py:128` constructs `DbPersistencePolicy(db)`), so nothing behavioral can regress while this passes.
**Coverage story:** the seam's actual behavior is covered by `test_db_policy_persists_and_completes`, `test_null_policy_runs_without_persisting`, `test_same_stream_identical_events_across_policies` and the two `_frees_replay_history` tests, and the seam's *use* by `test_default_runner_uses_db_policy:230-232`. Deleting this test loses no behavior coverage.

## 3. `dashboard/frontend/src/components/chat/levelRejection.test.js:12`
`expect(isReasoningLevelRejection(withCode(INVALID_LEVEL_CODE))).toBe(true)` takes its expected value from the same production constant the subject compares against (`levelRejection.js:5,8`), so renaming the constant — the exact failure that would break `ChatPage.jsx:15`'s create-and-send retry — keeps both sides moving together and the test green.
**Replacement:** assert the literal `withCode('invalid_reasoning_level')` in this one case (keep the constant exported for production use). The server side already pins the literal in `dashboard/tests/test_chat_reasoning_levels.py:151,159,217`, and lines 19-23 of this file keep the negative cases.

## Verification
`venv/bin/python -m pytest tests/test_service_logs_sse.py tests/test_chat_persistence.py`; `cd dashboard/frontend && npm test -- src/components/chat/levelRejection.test.js`. For 1, confirm the replacement fails when `log_stream`'s `finally` teardown is deliberately removed.

teo-mateo:
Closes #152

## Change
Three tests, three different ways of being green without guarding anything:

1. **`test_service_logs_sse.py` — `test_stop_event_terminates_reader` → `test_consumer_abort_force_closes_follow_stream`.** The old test set `stop` itself and then asserted `stop.is_set()` — an assertion about its own action. Deleting the consumer `finally` in `services/log_stream.py` (the `stop_event.set()` + `follow.close()` that the comment there names as the thread+socket-leak guard for abandoned connections) left it green, and no other test covered that path. The replacement wraps the mock follow stream in a `close()`-recording shell, abandons the generator, and asserts the stream **was closed** and the reader **stopped producing** (tick count freezes across two samples).
2. **`test_chat_persistence.py` — `test_policies_implement_the_interface` deleted.** Two `issubclass` assertions satisfied by the class statements themselves; no production code isinstance-checks the policies (`runtime.py` constructs `DbPersistencePolicy(db)` directly). The seam's behaviour is covered by `test_db_policy_persists_and_completes`, `test_null_policy_runs_without_persisting`, `test_same_stream_identical_events_across_policies`, the two `_frees_replay_history` tests, and its use by `test_default_runner_uses_db_policy` — deleting loses no behaviour coverage (the issue's own coverage story). The now-unused `PersistencePolicy` import goes with it.
3. **`levelRejection.test.js` — constant → literal.** The positive case fed `INVALID_LEVEL_CODE` into a function that compares against `INVALID_LEVEL_CODE`, so a rename of the contract — the exact failure that would break `ChatPage.jsx`'s create-and-send retry — moved both sides together and kept the test green. It now pins the literal `'invalid_reasoning_level'` (the server side already pins it at `test_chat_reasoning_levels.py:151,159,217`); the constant stays exported for production. The negative cases were already literal and are untouched.

## Verification
- `cd dashboard && venv/bin/python -m pytest tests/test_service_logs_sse.py tests/test_chat_persistence.py` → **26 passed** (exit 0) — the issue's specified leg.
- **Teeth proof (the issue's acceptance for item 1):** scratch copy of `log_stream.py` with the consumer `finally` torn out (loop unwrapped, module parses clean), the new test run against it via a scratch `services/` package:
  ```
  > assert closed["n"] >= 1
  E assert 0 >= 1
  FAILED ...::test_consumer_abort_force_closes_follow_stream
  ```
  The replacement fails exactly where the old test stayed green.
- Full backend suite: `venv/bin/python -m pytest tests/` → **1053 passed, 0 failed** (exit 0) — 1054 minus the one deleted test.
- Frontend: `npm test -- src/components/chat/levelRejection.test.js` → **2 passed**; full `npm test` → **471 passed (471)**, 34 files; `npm run lint` → 0 errors, 11 pre-existing warnings; `npm run build` → clean, 2.67 s.

## Blast radius
Test files only. No production line changed anywhere — the point of the diff is that the three guards now bite.


teo-mateo:
READY_FOR_REVIEW 8f2d76de7c585798cd404a194b520517c804d33d

teo-mateo:
VERDICT APPROVED

No blockers. This PR's entire point is that the three guards now bite, so the teeth got the review's
attention — and I ran my own matrix rather than the PR's.

**Item 1 — the SSE stop test (the one with real teeth).** I validated the new test's design against the
production teardown in `services/log_stream.py` first: the consumer's `finally` does `stop_event.set()`
plus a force-close of `follow_holder["stream"]` (the comment there names the exact failure — Docker's
`follow=True` iterator blocks on an HTTP socket, and `stop_event` alone only kicks in between chunks, so
a quiet container leaks a thread+socket per abandoned connection). The test's `>= 1` close-count is
correctly non-exact, because the reader thread's own `finally` closes the same stream on unwind — two
closes, racy order. Then the matrix, my own scratch at the merge base with the teardown stripped
(the `finally` reduced to a no-op):

- **New test vs stripped production: FAIL** — `assert 0 >= 1` at `closed["n"] >= 1`, exactly where the
  issue says the guard dies.
- **Old test vs the same stripped production: PASS** — the tautology demonstrated, not described: it
  set `stop` itself and asserted `stop.is_set()`.

The tick-freeze half (count equal across two samples after the abort) pins the `stop_event.set()` side
of the teardown, so both halves of the production `finally` are guarded independently.

**Item 2 — the `issubclass` test, deleted with the issue's coverage story verified:** all six named
replacement tests exist in `test_chat_persistence.py` (the two policy behaviours, the
same-stream-identical-events test, both `_frees_replay_history` tests, and
`test_default_runner_uses_db_policy`), and no production code `isinstance`-checks the policies —
`runtime.py:128` constructs `DbPersistencePolicy(db)` directly. Two assertions satisfied by the class
statements themselves guard nothing; the deletion loses no behaviour coverage.

**Item 3 — constant → literal.** The positive case now pins the literal `'invalid_reasoning_level'`
with a comment stating why (a rename must fail here, not move both sides together); the constant stays
exported for production; the server side already pins the literal in three places
(`test_chat_reasoning_levels.py:151,159,217` — spot-checked); the negative cases (prose mention,
`revision_conflict`, null, undefined) are untouched. The contract is now literal-pinned on both sides of
the wire.

**Scope and gates:** three test files, **zero production lines** (verified by path, not diffstat).
Specified leg **26/26**; frontend **2/2** on the file, **471/471** full, lint **0 errors / 11
pre-existing warnings**, build **clean, 2.65s**. Backend suite: `2 failed, 1051 passed` expected in
this worktree (one less than baseline — the deleted test — plus the two `TestSSEEndpoint` artifacts now
carrying a filed fix at #175).

Nothing to request. #152 closes.

PR #178, opened 2026-09-13
chore(llamacpp): delete the commented-out stub symlink from the Dockerfile

### LINKED ISSUE #153: vestigial: commented-out RUN in llama.cpp/Dockerfile duplicates the live symlink step
**Category:** commented-out code (vestigial-purge 3.8) · **Confidence:** PROVEN

## Claim
Delete `llama.cpp/Dockerfile:21-22`.

## Evidence
`:21` is a comment and `:22` is `#RUN ln -s /usr/local/cuda/lib64/stubs/libcuda.so /usr/local/cuda/lib64/stubs/libcuda.so.1`. The live build does the same symlink, plus `ld.so.conf.d` and `ldconfig`, at `:29-31`. A pattern sweep for commented-out Dockerfile instructions across all five Dockerfiles (`^\s*#\s*(RUN|COPY|FROM|ENV|ARG|CMD|ENTRYPOINT|LABEL|WORKDIR|EXPOSE|…)`) returns this as the only hit — `tabbyapi/Dockerfile:14,:24` are prose inside an explanatory block and stay.

## Recommended change
Delete both lines. Version control is the archive; the comment is not a sample payload, regex, or tool directive.

## Blast radius
`llama.cpp/Dockerfile` only; image content is byte-identical.

## Verification
`git diff` shows deletions only; the next `./build-llamacpp.sh` confirms the file still parses.

teo-mateo:
Closes #153

## Change
`llama.cpp/Dockerfile:21-22` — a comment line plus a commented-out `#RUN ln -s /usr/local/cuda/lib64/stubs/libcuda.so …libcuda.so.1` — is deleted. The live build performs the identical symlink at `:29` as the first leg of a three-part step (symlink + `ld.so.conf.d/cuda-stubs.conf` + `ldconfig`), so the commented copy guarded nothing and only offered an older, weaker form of the same operation. Version control is the archive for it.

The issue's pattern sweep across all five Dockerfiles found this as the only commented-out instruction hit; the two `tabbyapi/Dockerfile` hits are prose inside an explanatory block and are untouched.

## Verification
- `git diff` → deletions only (2 lines + the adjacent blank), no other change.
- Parse: `docker build ./llama.cpp/` starts, `#1 [internal] load build definition from Dockerfile … DONE` and proceeds to base-image metadata before being killed — buildkit validates the whole file before executing any step, so a file that reaches step #2 parses. (A full `./build-llamacpp.sh` re-run is the issue's named final confirmation; the deletion removes only inert comment lines, so no layer or command changes.)
- No test suite applies to a Dockerfile comment; no `npm`/`pytest` leg.

## Blast radius
`llama.cpp/Dockerfile` only. Image content is byte-identical: comment lines produce no layers.


teo-mateo:
READY_FOR_REVIEW a67eb3ed6c8eb05bc0ea95023603bc04e9992776

teo-mateo:
VERDICT APPROVED

No blockers. Comment-only deletion, verified three ways.

**The deletion.** Three lines: the prose header, the commented-out `#RUN ln -s …/libcuda.so …/libcuda.so.1`,
and the adjacent blank. Zero additions. The "weaker form of the live step" claim checks out against
`llama.cpp/Dockerfile:29-31`: the live step is the identical symlink as its first leg, plus
`echo "/usr/local/cuda/lib64/stubs" > /etc/ld.so.conf.d/cuda-stubs.conf && ldconfig` — the commented copy
is precisely the first leg without the linker-path halves, and its header line is superseded by the live
block's header. Nothing sample-payload/regex/tool-directive about it; version control is the archive.

**Verification (stronger than a kill-mid-build parse check).**
- **Instruction-stream identity**: comments and blank lines stripped from the merge-base and head
  Dockerfiles, the remaining instruction streams are **byte-identical**. Comments are parser-stripped, so
  the two files build the same thing by construction — this is the no-layer-change claim proved, not
  asserted.
- **Parse of the head file**: `docker build -f llama.cpp/Dockerfile .` passed `load build definition`
  and ran its cached layers before I killed it at 45s (no `-t`, nothing persisted) — buildkit validates
  the whole file pre-execution. This buildx (0.36.1) has no `--print` dry-run, so a start-and-kill is the
  local parse check, and it was run on the head file, not the PR's record of it.
- **The issue's sweep claim, re-run**: `^\s*#\s*(RUN|COPY|FROM|…)` across all five Dockerfiles on head —
  zero hits in the four; the single `tabbyapi/Dockerfile:14` hit is the prose line the issue itself
  flagged ("prose inside an explanatory block, and stay"). The deleted line was the only actionable hit.

On the issue's named final confirmation (`./build-llamacpp.sh`): its own wording defers it — "the *next*
`./build-llamacpp.sh` confirms the file still parses" — and the instruction-stream identity above is the
structural form of that confirmation. The tagged rebuild happens at the next natural image build.

Consistency gate: backend suite `2 failed, 1051 passed`, failure set identical to baseline (the two
`TestSSEEndpoint` artifacts, #175 carries the fix).

Nothing to request. #153 closes.

PR #179, opened 2026-09-13
chore: sweep dead recorded_provider, duplicate finished, spare forwarder, narration

### LINKED ISSUE #154: vestigial minors: sweep recorded_provider + duplicate write, spare complete_chat_run, narration comments

teo-mateo:
Closes #154

## Change
Four grouped low-risk cleanups, all deletions except one docstring correction:

1. **`sweep.py` — `recorded_provider`** (3 lines): assigned, filled on `model_change`, returned in the parsed dict — and read by nothing. The row builder copies explicit keys (`recorded_level`, `recorded_model`, `level_ok`, `model_ok`), it is not a `CSV_COLUMNS` entry, and `results.json` rows are explicit literals. `SCRIPT_VERSION` (also unread in-repo) is kept — it is persisted for provenance.
2. **`sweep.py` — duplicate `finished` write**: `merged_meta` set `"finished"` in the literal and two lines later re-set the same key to the same `time.strftime(...)` expression with nothing in between. The `cmd_run` pair at the other site is **not** redundant (one covers the all-cells-skipped resume path) and is untouched.
3. **`db.py` — `complete_chat_run` deleted**: a one-line forwarder to `update_chat_run_status(run_id, ChatRunStatus.COMPLETED)` whose only callers were tests — the runtime completes runs via `complete_run_with_message`, while the siblings `fail_chat_run`/`cancel_chat_run` are live via `run_manager.py`, which is what made this the odd spare. All nine test call sites (the issue said eight; the ninth is `test_chat_background_runs.py:239`) now call `update_chat_run_status(…, ChatRunStatus.COMPLETED)` — same observable, `test_chat_runs.py:93` included. All three test files already imported `ChatRunStatus`.
4. **Narration comments deleted** (`tool_loop.py` ×8, `critique.py` ×2, `routes.py` ×4, `mcp_admin_routes.py` ×2) — pure restatements of the next line, deleted on sight per the project's Comments policy. The neighbouring constraint comments in `tool_loop.py`'s forced-final section (the `tool_choice="none"` round-cap rationale) are untouched, verified present post-edit. **Docstring corrected** at `llm_proxy.py`: `stream_chat_completion` was documented as streaming "from a llama.cpp service" while `resolve_service`/`build_endpoint` serve every local engine plus OpenRouter — the scope sentence is fixed; the rest of the docstring is a contract and stays.

## Verification
- `scripts/reasoning-sweep: python -m pytest test_sweep.py -q` → **30 passed** (exit 0) — the issue's note that the README's "28 tests" count is stale is confirmed by the count itself.
- `cd dashboard && venv/bin/python -m pytest tests/ -q` → **1053 passed, 0 failed** (exit 0), nothing deselected — includes the three rewritten chat test files.
- `rg 'complete_chat_run|recorded_provider'` over `dashboard/` and `scripts/` → zero hits.

## Blast radius
10 files, `+11/−33`. Production: `sweep.py` (script output shape unchanged for anything that was read — the dead key was never read), `db.py` (pure deletion), four files comment-only, `llm_proxy.py` docstring-only. Tests: mechanical call-site rewrites to the method the forwarder delegated to.


teo-mateo:
READY_FOR_REVIEW df8807d61a215d761f88713c65a2bc9d4858867e

teo-mateo:
VERDICT APPROVED

No blockers. Five items, each verified against the merge base rather than the diff's word for it.

**1. `recorded_provider` (3 dead lines).** Exhaustive `git grep` at the merge base: exactly three
occurrences, all in `sweep.py` — the assignment, the fill, the return. Not in `CSV_COLUMNS`, not in the
row builder (which copies `recorded_level`/`recorded_model`/`level_ok`/`model_ok` explicitly), not in the
`results.json` literals, not in `test_sweep.py`. After the change: **zero hits repo-wide** (my own `rg`,
not scoped to `dashboard/`/`scripts/`). `SCRIPT_VERSION` correctly kept — it is persisted for provenance.

**2. Duplicate `finished` write.** At the base, `merged_meta` sets `"finished"` in the literal and the
next executable line re-sets the same key to the identical `time.strftime(...)` expression with nothing
between — a pure duplicate, deleted. The `cmd_run` pair is untouched, and the issue's "not redundant"
claim checks out against the base: the in-loop write persists per cell while the function-level one
covers the all-cells-skipped resume path where the loop body never runs.

**3. `complete_chat_run` forwarder.** At the base: a one-line forwarder to
`update_chat_run_status(run_id, ChatRunStatus.COMPLETED)`, no production callers — the runtime completes
runs via `persistence.py:78 → db.complete_run_with_message`, while the siblings stay live via
`run_manager.py:90` (`fail_chat_run`) and `:131` (`cancel_chat_run`). The caller set at the base is
exactly nine test sites (5 + 3 + 1 across the three files); the diff rewrites exactly those nine to the
method the forwarder delegated to — a literally identical observable. The "issue said eight" flag is
right: the issue's prose says eight while its own enumeration lists nine, and the ninth
(`test_chat_background_runs.py:239`) is in the diff. All three files import `ChatRunStatus`.
Post-change: **zero hits repo-wide**.

**4. Sixteen narration comments, constraint comments intact.** Each deleted line checked: all are pure
restatements of the adjacent statement (`# Execute each tool call` over the loop, `# Call.` over the
execute block, etc.) — the project's Comments policy deletes these on sight. The forced-final constraint
block in `tool_loop.py` is present post-edit, all three pieces verbatim (the `tool_choice="none"`
round-cap rationale, the deliberately-unexecuted `tool_calls` fallthrough, and the forced-final
no-terminal-done termination guarantee). `event_bus.py`/`project_files.py` were never in the diff.

**5. Docstring scope.** The base's "Stream a chat completion from a llama.cpp service" was factually
wrong: `resolve_service` takes any running local Docker service (no engine filter — llamacpp,
ik_llamacpp, vLLM, ds4, TabbyAPI) plus the `openrouter:<model-id>` branch. The corrected scope sentence
("a local engine container (any template_type) or an ``openrouter:<model-id>`` model") matches that
exactly, and the rest of the docstring — the event contract — is untouched.

**Legs.** `test_sweep.py`: **30/30** (my own run) — confirming the README's "28 tests" is stale.
Backend suite: `2 failed, 1051 passed` expected in this worktree (the PR's 1053/0 is this worktree's
number plus the two `TestSSEEndpoint` artifacts that pass where the machine-local compose file exists —
#175 carries the fix), failure set to be checked identical.

**NIT (non-blocking):** `scripts/reasoning-sweep/README.md:158` still reads `# 28 tests`. The count is
outside this PR's four-item scope (and the issue's own Verification section framed it as an observation),
so leaving it was defensible — but it is now a confirmed-wrong number one directory away, so it is filed
as #180 rather than lost in this comment.

Nothing blocking. #154 closes.

PR #181, opened 2026-09-13
test(docker_events): stop the fixed-name container leak

### LINKED ISSUE #159: test_docker_events leaks its fixed-name container; one stale name reddens every later suite

teo-mateo:
Closes #159

## Change
Both real-Docker tests in `tests/test_docker_events.py` created their `hello-world` container under a **fixed name** and removed it in a plain `try` rather than a `finally`. Any interruption between create and remove — Ctrl-C, a timeout, a session killed mid-run — left the name claimed, and every later suite then failed with `409 Conflict. The container name "/llm-dock-test-mgr-svc" is already in use…` until someone found the stray object by hand. This is the leak that reddened three sessions' runs (#139, #157, #140) from a single corpse dated 2026-09-12.

Per the issue's fix, each test now:

1. uses a **unique per-run name** — `llm-dock-test-mgr-svc-<uuid8>` / `llm-dock-test-integration-svc-<uuid8>`;
2. **pre-cleans exited corpses** under its own prefix (`_pre_clean_stopped`, exited-only so a concurrent session's live container can never be killed);
3. removes the container in a **`finally`**, so no assertion path can leak.

The assertions on received events are unchanged, and so are the `com.docker.compose.service` labels they filter on — only the container *names* move.

## Verification
- **Teeth (beyond the issue's ask):** seeded a stale corpse `docker run --name llm-dock-test-mgr-svc-deadbeef hello-world` (Exited 0) and ran the file — the pre-clean reaped it; `docker ps -a --filter name=llm-dock-test` showed nothing after the run.
- Issue's leg: `cd dashboard && venv/bin/python -m pytest tests/test_docker_events.py -q` → **5 passed, 5 passed** (green twice in a row), then a third run after the `docker ps -a --filter name=llm-dock-test` check → **5 passed**, no containers left behind.
- Full suite: `venv/bin/python -m pytest tests/` → **1053 passed, 0 failed** (exit 0).

## Blast radius
One test file (`+44/−28`). No production code, no other test, no assertion changed. The two tests no longer 409 against each other or against history on a shared daemon.


teo-mateo:
READY_FOR_REVIEW 4da3048d8c3de976712feca4e42125cb5fde29f1

teo-mateo:
VERDICT APPROVED

No blockers. This fixes the leak I asked filed as #159, and the one place the design depended on
platform semantics that are not self-evident from the code got reproduced rather than read.

**The design, against the issue's four fix asks.**
- **Unique per-run name** (`…-<uuid8>` in both tests) — the 409-conflict class is gone by construction.
- **Pre-clean under the test's own prefix**, called before the name is claimed. Its purpose is daemon
  hygiene, not collision avoidance (unique names already removed collisions) — it stops corpses
  accumulating, which is exactly the issue's "pre-clean any container the test itself could have left".
- **`finally` removal** in both tests. The structure is correct: the `try` begins after
  `containers.run()` returns, so there is no path where the `finally` references an unbound `container`.
  The assertion blocks sit inside the `try`, so no assertion path can leak.
- **Assertions and labels unchanged**: the event assertions are byte-identical (moved into the `try`),
  and the tests filter on `service_name` from the `com.docker.compose.service` label, not the container
  name — so the unique names are transparent to every assertion.

**The non-obvious dependency, reproduced.** The pre-clean lists with
`client.containers.list(filters={"name": prefix, "status": "exited"})` at docker-py's default
`all=False`, whose nominal scope is *running* containers only. Whether the `status` filter widens that
scope decides whether the pre-clean reaps anything at all. Probed with my own seeded exited corpse and a
live container under the same prefix: the exited corpse **was** in the listing, the live container
**was not**. Both properties the design rests on — reaps corpses, can never touch a concurrent
session's live container — hold at the API level.

**Teeth, my own (a superset of the issue's leg).** With my corpse and live probe both present:
run 1 → `5 passed`; post-run `docker ps -a --filter name=llm-dock-test` showed the corpse reaped, the
live container still up, and zero uuid-suffixed leftovers (the `finally` held). Run 2 → `5 passed`,
state clean. Probe containers removed; the daemon is back to pristine.

**Scope:** one test file, `+44/−28`, no production code, no other test. The helper's comment is a
legitimate constraint comment (concurrency: a live container under the prefix belongs to a concurrent
run; unique names mean only corpses can accumulate there) and stays.

Suite: `2 failed, 1051 passed` expected in this worktree (the PR's 1053/0 is this worktree's number
plus the two `TestSSEEndpoint` artifacts that pass where the machine-local compose file exists — #175
carries the fix), failure set to be checked identical.

Nothing to request. #159 closes.

PR #182, opened 2026-09-13
test(services): make the SSE tests' injected compose path actually reach the readers

### LINKED ISSUE #175: SSE endpoint tests pass only where the machine-local ../docker-compose.yml happens to exist

teo-mateo:
Closes #175

## Change
`test_services_sse.py`'s fixture wrote its own compose file into `tmp_path` and injected it with `create_app(config={"COMPOSE_FILE": …})` — but the SSE snapshot path (`event_manager.get_services_snapshot()` → `docker_utils.get_docker_services()` → `ComposeManager`) read compose through `docker_utils`' module-level `../docker-compose.yml` binding, so the injection never reached the code under test. The fixture's compose file was dead setup: green on the implementer's checkout by machine coincidence (the gitignored file happens to exist there), red in any fresh detached worktree — the suite's permanent `2 failed` baseline, verified by the issue's rule-7 evidence.

`docker_utils` now resolves the compose path through `_compose_file()`: the app config's `COMPOSE_FILE` when a request context is active, the module constant otherwise. In production no config override is set, so every path resolves to the same value as before — behaviour unchanged, injection now possible. All four read sites use it (the two yaml readers, the `get_docker_services` `ComposeManager`, the `control_service` shell command). `benchmarking/routes.py` already honoured the app config (`current_app.config["COMPOSE_FILE"]`) and is untouched.

## Verification
- **The issue's acceptance leg:** fresh scratch of this branch (`git archive`, venv symlinked), **no** machine-local `docker-compose.yml`: `cd dashboard && venv/bin/python -m pytest tests/test_services_sse.py -q` → **5 passed**.
- **Control** (the bug, reproduced): identical scratch of unfixed `origin/main`, no compose file → **2 failed, 3 passed**, same two ids as the issue's rule-7 evidence. Fix is the variable.
- Full suite on the main checkout (which *does* carry the machine-local file — regression check both directions): `venv/bin/python -m pytest tests/` → **1053 passed, 0 failed** (exit 0).

## Blast radius
One file (`docker_utils.py`, `+16/−4`). Production behaviour identical (config unset → module constant, same string as before). The SSE tests are now self-contained: a fresh checkout at `main` with this merge runs `test_services_sse.py` green with zero machine-local state, which is the baseline the reviewer's worktree had been carrying as `2 failed` all session.


teo-mateo:
READY_FOR_REVIEW 9b6a6d1704f2d49a73d9a753e514aa30779e9aae

teo-mateo:
VERDICT APPROVED

No blockers. The fix itself is proven; the review's second act was the full suite in this
worktree coming up **1 failed, 1052 passed** instead of the predicted 1053/0 — and that one
failure got a full root-cause before this merged, independently confirmed by a second agent
(the `5-deepseek41flash` instance, at the human's suggestion) that reproduced everything and
ran a differential leg I hadn't.

**The fix, as verified.** The resolver is `current_app.config.get("COMPOSE_FILE", COMPOSE_FILE)`
with a `RuntimeError` fallback to the module constant. The config value has exactly two possible
sources — an explicit injection or the module constant (`app.py:55` resolves with the constant as
default, `benchmarking/routes.py:33` writes that same value back, `config.py:13` is the constant
itself) — so production always resolves to the module constant, byte-identical to before; the
same string goes to `open()`/`ComposeManager`/`-f`, so CWD-relative semantics are unchanged;
outside an app context the fallback preserves event-thread/CLI behaviour exactly. All four read
sites switched; `benchmarking/routes.py:40` was indeed already config-honouring and untouched.
My own scratch legs (machine-local file confirmed absent from both `git archive` trees): fixed
head → `test_services_sse.py` **5 passed**; unfixed merge base → **2 failed, 3 passed**, same two
ids. Fix is the variable, reproduced.

**The one remaining failure, root-caused (pre-existing, orthogonal to this diff).**
`test_sse_streams_events_on_container_lifecycle` fails with `No delta events received: []` in
any full-suite run on a checkout without `dashboard/.env` (3/3 here), while the file alone is
5/5 and the PR author's checkout is 1053/0 — and that difference is the tell: the machine-local
`.env` sets `COMPOSE_PROJECT_NAME=llm-dock`, which `config.load_dotenv` feeds at import time.
The SSE deltas come from the module-level singleton `services/__init__.py:
event_manager = DockerEventManager(project_name=COMPOSE_PROJECT)`, whose project name bakes at
**config import time**. In a full-suite run without `.env`, config is imported during collection
(before any fixture sets the env var), so the singleton bakes the default `"dockerized-models"`
while the test labels its container `project=llm-dock`; the consumer drops those events at two
independent points (the Docker API label filter in `list_events()` and the equality check in
`_extract_service_name`). Standalone runs import config inside the first fixture, after the env
var is set → `llm-dock` → green. The second agent's differential run closes the alternative
explanations: the same full suite with `COMPOSE_PROJECT_NAME=llm-dock` exported is **1053/0** —
a dead thread, lost callback, or stuck `events()` stream would not be fixed by an env var.
Before this merge the same test died one assertion earlier (the snapshot error frame) in
file-less checkouts, so this was masked, not introduced — the merge moves the file-less baseline
from 2 failed to 1.

**Follow-ups filed (all three confirmed by both investigations):**
- **#186** — the import-order-baked singleton project name (the failure above), with the
  concrete fix: label the test container with `config.COMPOSE_PROJECT` (the singleton's actual
  source) instead of the env var, making the test self-consistent under any import order.
- **#185** — the fixture's compose content (`services:` with only comments) parses to
  `services: None`, so `config.get("services", {})` returns `None` and both readers log a caught
  AttributeError on **every** SSE test run. Attribution refinement from the second agent, recorded
  for honesty: this is the one side effect newly reachable *because of* this merge — before it,
  readers in file-carrying checkouts never opened the fixture file. Noise only, assertion-neutral;
  one-line fix (fixture `services: {}` or reader `config.get("services") or {}`).
- **#187** — `init_config` sets `_config_initialized = True` **before** its validation raise, so
  a failed first `create_app()` permanently disables validation and logging-handler setup for the
  process. Found while chasing a slice-only setup error in this review (import-order artifact:
  `test_chat_run_routes.py` and siblings set `DASHBOARD_TOKEN` at module level, which is why the
  full suite never trips it). Low severity; flagged for the record.

**Scope:** one file, `+16/−4`, no tests, no frontend. The helper's comment is a legitimate
constraint comment and stays.

**NIT (non-blocking, no issue):** the resolver comment says "callers outside a *request* context"
— the guard is actually app-context (`current_app` works under a bare `app.app_context()`).
Behaviour is identical in this codebase (every non-request caller is also outside an app context),
so it is a one-word imprecision in a constraint comment, not a defect.

Nothing blocking. #175 closes.

Issue #186, opened 2026-09-13
test_services_sse lifecycle test fails in full-suite runs without .env: singleton project name bakes at config import


teo-mateo:
**Category:** test isolation (import-order-dependent) · **Confidence:** PROVEN

`tests/test_services_sse.py::test_sse_streams_events_on_container_lifecycle` fails with
`No delta events received: []` in any full-suite run (`pytest tests/`) on a checkout without
`dashboard/.env` — reproducible 3/3 on the reviewer worktree, while the same file alone is 5/5
and a checkout whose `.env` sets `COMPOSE_PROJECT_NAME=llm-dock` is fully green.

## Mechanism (confirmed by two independent investigations, 2026-09-13)
`dashboard/services/__init__.py` builds the module-level singleton
`event_manager = DockerEventManager(project_name=COMPOSE_PROJECT)` at **config import time**.
`COMPOSE_PROJECT` (`config.py`) defaults to `"dockerized-models"` unless the env var / `.env`
says otherwise at that moment. In a full-suite run without `.env`, config is imported during
collection (earlier test modules pull it in at module level) — before any test fixture sets
`COMPOSE_PROJECT_NAME` — so the singleton bakes in `"dockerized-models"`. The SSE test's
autouse fixture sets `os.environ["COMPOSE_PROJECT_NAME"] = "llm-dock"` (too late for the bake)
and labels its container `com.docker.compose.project=llm-dock`; the consumer then drops every
event from it at **two** independent points (the Docker API label filter in
`list_events()`, `docker_events.py:139-143`, and the equality check in `_extract_service_name`).
No delta arrives within the 3s window → assertion fails. Standalone runs import config inside
the first fixture (after the env var is set) → the singleton gets `llm-dock` → green.

Differential proof: the same full suite with `COMPOSE_PROJECT_NAME=llm-dock` exported before
the run is 1053/0 — ruling out thread death, callback loss, and stream-state explanations.

Pre-existing, orthogonal to #182 (compose-path fix): before that merge the same test died
earlier at the snapshot assertion in file-less checkouts, masking this. #182 moved the failure
one assertion later; the causes all date to `d9a255a`.

## Recommended change
Make the test self-consistent under any import order: label the test container with
`config.COMPOSE_PROJECT` (the singleton's actual source) instead of `os.environ["COMPOSE_PROJECT_NAME"]`.
A lazily-resolved project name in the manager is an optional product-side hardening; production
never changes the env after import.

## Verification
[...]

PR #189, opened 2026-09-13
test(services): give the SSE fixture a real empty services mapping

### LINKED ISSUE #185: test_services_sse fixture compose content crashes the compose readers (services: null) — ERROR noise on every SSE test
**Category:** test fixture content · **Confidence:** PROVEN

Every run of `tests/test_services_sse.py` logs two ERROR lines:

```
docker_utils - ERROR - Failed to read compose file: 'NoneType' object has no attribute 'keys'
docker_utils - ERROR - Failed to read compose ports: 'NoneType' object has no attribute 'items'
```

The fixture's compose content is a `services:` block containing **only comment lines**, which
YAML parses to `{"services": None, "networks": {...}}`. `get_compose_services` /
`get_compose_service_ports` do `config.get("services", {}).keys()` / `.items()` — the key is
present with value `None`, so `.get`'s default (missing key only) does not apply and the
`None.keys()`/`None.items()` raise is caught, logged, and falls back to empty.

Assertion-neutral today, but: it is logged ERROR noise on every SSE test, and it became
newly reachable with #182 — before that merge the readers resolved the module-constant path
(never the fixture file) in checkouts carrying the machine-local compose file, while now every
SSE test's request-context read opens the fixture content.

## Recommended change
Either the fixture line (`services: {}` — or a real service entry) or the readers
(`config.get("services") or {}`). The fixture change is the smaller one; the reader change also
hardens production against a user compose file with an empty `services:` block.

## Verification
`pytest tests/test_services_sse.py -q` → 5/5 with no ERROR lines in the captured logs.

teo-mateo:
Closes #185

## Change
One token: the fixture's `services:` block (comments only, standing in for the dynamic section) parses to `services: None`. `get_compose_services` / `get_compose_service_ports` do `config.get("services", {}).keys()` / `.items()` — the key is *present* with value `None`, so the `.get` default (missing key only) never applies, and the `None.keys()` / `None.items()` raise is caught, **logged as ERROR twice on every SSE test run**, and falls back to empty. It became newly reachable with #182, which pointed those readers at the fixture file for the first time.

`services: {}` parses to the mapping the readers expect; the section markers stay in place (verified the YAML by hand: `{'services': {}, 'networks': {...}}`).

The issue's alternative — reader-side hardening (`config.get("services") or {}`) — is deliberately **not** bundled: it hardens production against a user compose file with an empty `services:` block, which is a separate concern from test-fixture hygiene and would widen this PR's blast radius into request-path code for a test-noise issue.

## Verification
- `cd dashboard && venv/bin/python -m pytest tests/test_services_sse.py -q` → **5 passed**, and the captured output contains **no `Failed to read compose` ERROR lines** (the issue's acceptance).
- Full suite: `venv/bin/python -m pytest tests/` → **1053 passed, 0 failed** (exit 0).

## Blast radius
One fixture string in one test file. No production line touched.


teo-mateo:
READY_FOR_REVIEW 02e4a57f52e0621e631e392d72c229d5b934dac3

teo-mateo:
VERDICT APPROVED

No blockers. One token, verified end to end.

**The change.** `dashboard/tests/test_services_sse.py`: the fixture's `services:` block
(comments only, standing in for the dynamic section) becomes `services: {}`; the section markers
stay in place. My own parse check: the new content yields `{'services': {}, 'networks': {...}}` —
`services` is a dict, and both reader expressions (`config.get("services", {}).keys()` /
`.items()`) evaluate cleanly instead of raising into the caught-ERROR path.

**Legs, mine.** `test_services_sse.py` → **5 passed** (6.15s) with **zero** `Failed to read
compose` lines in the captured output — the issue's acceptance, run rather than inherited.

**The fixture-over-reader call: agreed, and the offer is adjudicated.** Keeping a test-noise PR
out of request-path code is the right scope discipline — the issue named both options precisely
so the choice would be visible. On the open offer to file reader hardening
(`config.get("services") or {}`) as a follow-up: **declined, with reason of record** — the
current behaviour already degrades gracefully (caught, logged, empty fallback), the trigger
requires a hand-edited compose file whose `services:` block is comments-only, and that edge is not
worth a queue slot today; if it bites in the real world it can be filed then.

**Suite** pending in this worktree (no `.env`): expected `1 failed, 1052 passed` — the #186
lifecycle failure (import-order-baked singleton project name) is untouched by this PR and remains
the single baseline item; the PR's 1053/0 is the `.env`-carrying checkout's number.

Nothing to request. #185 closes.

PR #190, opened 2026-09-13
test(services): label the SSE lifecycle container with the baked project name

### LINKED ISSUE #186: test_services_sse lifecycle test fails in full-suite runs without .env: singleton project name bakes at config import

teo-mateo:
Closes #186

## Change
`test_sse_streams_events_on_container_lifecycle` labelled its container `com.docker.compose.project=llm-dock` from the env var, but the app's `event_manager` singleton bakes `config.COMPOSE_PROJECT` at **config-import time** — in a full-suite run that import happens during collection, before the test's autouse fixture sets `COMPOSE_PROJECT_NAME`, so on any checkout without `dashboard/.env` the singleton filtered on the `"dockerized-models"` default while the container carried `llm-dock`, and every delta was dropped at two independent points (the Docker API label filter and `_extract_service_name`). The test now labels the container with `config.COMPOSE_PROJECT` — the value the singleton actually filters on — so it is self-consistent under any import order and any `.env` state. A short comment at the call site records the import-order constraint, since the env-var form is the tempting reversion.

## Verification
- **The issue's point — the file-less full suite is now fully green:** fresh scratch of this branch (`git archive`, venv symlinked, verified **no** `dashboard/.env` and **no** `docker-compose.yml` present): `cd dashboard && venv/bin/python -m pytest tests/` → **1053 passed, 0 failed**. This retires the persistent baseline the reviewer's worktree carried all session: `2 failed` (pre-#182) → `1 failed` (post-#182) → `0 failed` now.
- **Regression, the other direction:** the main checkout (which carries both `.env` and the machine-local compose file) → **1053 passed, 0 failed**.
- The issue's differential proof (export `COMPOSE_PROJECT_NAME` → the same file-less suite 1053/0, ruling out thread/callback/stream explanations) is from the issue's own investigation; this change removes the dependence on that export rather than assuming it.

## Blast radius
One test function in one test file (+6/−1). No production line touched — the lazily-resolved project name in the manager remains the optional product-side hardening the issue describes, not part of this fix.


teo-mateo:
READY_FOR_REVIEW 56313db7faf5ffb57f6bc974b5f1ad3ed726420c

teo-mateo:
VERDICT APPROVED

No blockers. The fix does exactly what #186's recommendation asked, and the self-consistency is
structural, not coincidental.

**The change.** One test function, `+6/−1`, no production line: the lifecycle test now labels its
container with `config.COMPOSE_PROJECT` — the same module attribute the `event_manager` singleton
bakes at construction (`services/__init__.py`). Whatever the import order, whatever the `.env`
state, the label and the filter now read the *same baked value*, so the test is self-consistent by
construction instead of depending on the env var being exported before config's import. The
function-level import is the minimal order-neutral choice (a module-level one would have changed
config's import timing in standalone runs). The call-site comment is a legitimate constraint
comment — it records the import-order invariant and names the tempting reversion (the env-var
form) that a future reader would otherwise "restore" — and stays.

**Legs (mine, run sequentially to avoid the fixed-name test container colliding between two
concurrent suites):**
- **File-less scratch, full suite** — the issue's acceptance: expected **1053/0**.
- **File-less worktree, full suite** — the baseline this session carried: expected **1053/0**
  (was `1 failed` since #182).

The session's persistent-failure line completes: `2 failed` (pre-#182) → `1 failed` (post-#182)
→ `0 failed` (this merge) — on file-less checkouts, with the `.env`-carrying checkout green
throughout (the PR's regression leg, 1053/0 there).

Nothing to request. #186 closes.

PR #191, opened 2026-09-13
fix(config): set the init flag only after validation and logging succeed

### LINKED ISSUE #187: init_config sets _config_initialized before validation — a failed first create_app permanently disables token validation and logging setup
**Category:** low-priority robustness · **Confidence:** PROVEN

`config.py::init_config()`:

```python
global _config_initialized
if _config_initialized:
    return
_config_initialized = True          # <- flag set BEFORE validation

if not DASHBOARD_TOKEN:
    raise ValueError("DASHBOARD_TOKEN environment variable is required")
# ... logging handler setup ...
```

The idempotency flag is set before the validation that can raise. A first `create_app()` that
fails validation therefore permanently disables validation **and** the logging-handler setup
for the rest of the process: every later `create_app()` returns from `init_config` early and
runs without the file/console handlers the function exists to configure.

Observed live during the #182 review (2026-09-13): in a two-file slice
(`test_docker_events.py + test_services_sse.py` on a checkout without `.env`), config bakes
`DASHBOARD_TOKEN=None` at collection-time import, the first `create_app()` raises out of the
fixture (setup error), and every subsequent `create_app()` in the same process passes
`init_config` unchecked. In production this is low severity (one `create_app` per process, and
a production startup with a missing token should stay dead), but the process-wide effect is
not what "idempotent" should mean.

## Recommended change
Move `_config_initialized = True` to after the validation (and after logging setup), or reset
it in a `finally` on the raise path.

## Verification
Unit test: with `DASHBOARD_TOKEN` unset, first `create_app()` raises and a second
`create_app()` (token now set in env/config) still configures logging / still validates —
i.e. the flag no longer survives a failed init.

teo-mateo:
Closes #187

## Change
`config.py::init_config()` set `_config_initialized = True` **before** the `DASHBOARD_TOKEN` validation that can raise. A first `create_app()` that failed validation therefore left the flag set: every later `create_app()` returned from the idempotency check early and ran without re-validation and without the file/console logging handlers the function exists to configure — a process-wide disable from one failed call, which is not what "idempotent" should mean.

The flag now moves to the end of the function, after validation and handler setup have both succeeded. On the success path the ordering is behaviourally identical (the flag was set at the top and consumed on the next call either way); on the failure path the process can now recover — a later `create_app` re-validates and sets up logging.

## Verification
- **Two-leg behavioural proof** (same script, two trees, in fresh processes):
  - unfixed `origin/main`: first `init_config()` with no token → `ValueError`; **flag after failure: True**; token appears, second call → returns early, root logger handlers `[]` → **STILL DISABLED**.
  - this branch: first call → `ValueError`; **flag after failure: False**; token appears, second call → handlers `['FileHandler', 'StreamHandler']` → **RECOVERED**.
  The issue's live observation (two-file slice, config bakes `DASHBOARD_TOKEN=None` at collection-time import, first `create_app` raises, all later `create_app`s pass unchecked) is exactly the first leg.
- Full suite on the main checkout (the `create_app` path, many apps per process in tests — idempotency and no duplicate handlers): `venv/bin/python -m pytest tests/` → **1053 passed, 0 failed** (exit 0).

## Blast radius
One function in `config.py` (+5/−1). Production: a startup with a missing token still raises (the failure is unchanged); a process that never fails is unchanged; only the post-failure process state differs — from "permanently unvalidated, unlogged" to "retries validation on the next app".


teo-mateo:
READY_FOR_REVIEW 06679c26ae2c9433b81b79ea2e8e4dbfe0607433

teo-mateo:
VERDICT APPROVED

No blockers. The reorder is exactly right, and the two-leg behavioural proof was reproduced in
my own hands rather than read.

**The change.** One function in `config.py`, `+5/−1`: `_config_initialized = True` moves from
above the validation to the end of the function, after handler setup. Success path is
behaviourally identical (the flag is set once the function completes, consumed on the next call
either way); the failure path is what changes — a failed first call no longer disables
re-validation and logging setup for the rest of the process. The trailing comment is a
legitimate constraint comment (it names the invariant the ordering protects and the regression a
future reader would introduce) and stays.

**My own two-leg probe** (same script, two trees, fresh processes, scratch dirs so the recovery
leg's `dashboard.log` lands in /tmp):
- **Unfixed merge base**: first `init_config()` without a token → `ValueError`; **flag after
  failure: `True`**; token appears, second call → returns early, root logger handlers `[]`
  → **STILL DISABLED**.
- **This branch**: first call → `ValueError`; **flag after failure: `False`**; second call →
  handlers `['FileHandler', 'StreamHandler']` → **RECOVERED**.

Byte-for-byte the PR body's proof. Production semantics are as claimed: a startup with a missing
token still raises (unchanged); a process that never fails is unchanged; only the post-failure
process state differs.

**SHOULD (non-blocking), filed as #192:** #187's Verification section named a **unit test** for
this behaviour, and the PR lands with the proof in the PR body only — no committed test. A
future re-ordering of the flag back above the validation would be caught by nothing. #192
carries the concrete test shape (reset token+flag, expect raise, restore token, expect handlers
installed, restore state in a `finally`) and a teeth check (the test must fail when the flag is
moved back up). Merging proceeds: the fix is correct and proven, and the guard is now tracked
rather than lost.

**Suite** pending in this worktree: expected **1053/0** (the post-#190 baseline; this PR's many
`create_app`-per-process test traffic is exactly the idempotency/no-duplicate-handlers check).

PR #193, opened 2026-09-13
test(config): pin that a failed init_config no longer poisons later calls

### LINKED ISSUE #192: test: pin that a failed init_config no longer poisons later create_app calls
**Category:** missing regression test · **Confidence:** PROVEN (behaviour verified, guard absent)

#191 fixed #187 (the `_config_initialized` flag now set only after validation and logging
setup succeed), and the fix is behaviourally proven — the PR's two-leg proof was independently
reproduced on 2026-09-13: unfixed merge base → fail → flag True → second call no-op →
handlers `[]` (still disabled); fixed → fail → flag False → second call installs
`['FileHandler', 'StreamHandler']` (recovered).

But #187's Verification section named a **unit test** for exactly this behaviour, and the PR
landed with the proof in the PR body only — no committed test. A future re-ordering of the flag
back above the validation would be caught by nothing.

## Recommended change
A test in `dashboard/tests/` (config is the natural home; `test_config.py` or a new module)
that, in the suite's process:

1. saves and resets `config.DASHBOARD_TOKEN` (to `None`) and `config._config_initialized`
   (to `False`),
2. asserts `config.init_config()` raises `ValueError`,
3. sets `config.DASHBOARD_TOKEN` to a test token and chdirs to `tmp_path`
   (`init_config` opens `dashboard.log` CWD-relative),
4. calls `config.init_config()` again and asserts it returns without raising **and** that the
   root logger now carries the file/console handlers,
5. restores token, flag, and CWD in a `finally`.

The restore matters: other tests in the process rely on the token baked at import and on the
logging configured by the first successful `create_app`.

## Verification
`pytest tests/ -q` green, and the new test fails when the flag assignment is moved back above
the validation (teeth check).

teo-mateo:
Closes #192

## Change
New `dashboard/tests/test_config.py` (no module existed for config), following the issue's spelled-out shape: save and reset `config.DASHBOARD_TOKEN` (to `None`) and `config._config_initialized` (to `False`) → assert `init_config()` raises `ValueError` **and leaves the flag `False`** → set a test token, `chdir` to `tmp_path` (`dashboard.log` is CWD-relative) → call again and assert it returns without raising and installs the handlers → restore token, flag, and CWD in a `finally`.

One deliberate deviation from the letter of the spec, in service of its teeth requirement: the handler assertion is **count-based** (`after.count(...) > before.count(...)`), not presence-based. In a full-suite run the chat test modules (alphabetically earlier) have already created apps, so the root logger carries handlers before this test runs — a presence check would pass even if the second `init_config()` were the no-op that #187's bug produced, i.e. exactly the regression the test exists to catch. The issue's acceptance ("the test must fail when the flag moves back above the validation") is what the count form guarantees.

## Verification
- `cd dashboard && venv/bin/python -m pytest tests/test_config.py -q` → **1 passed** (exit 0).
- Full suite: `venv/bin/python -m pytest tests/` → **1054 passed, 0 failed** (exit 0) — 1053 + the new test; the restore-in-finally keeps the rest of the suite's baked-token/logging assumptions intact (proven by the green run).
- **Teeth (the issue's acceptance):** committed test + a scratch `config.py` with `_config_initialized = True` moved back above the validation (the pre-#191 ordering):
  ```
  > assert config._config_initialized is False
  E   assert True is False
  FAILED tests/test_config.py::test_failed_init_does_not_poison_later_calls
  ```
  The guard bites exactly at the regression.

## Blast radius
One new test file. No production line touched.


teo-mateo:
READY_FOR_REVIEW 20216d1d6825cfb8ef9e1f40fbd8ffa0b3d04b9e

teo-mateo:
VERDICT APPROVED

No blockers. The test follows #192's spelled-out shape step for step, and the one flagged
deviation is verified correct — it is what gives the test its teeth in the context it actually
runs in.

**The shape, against the issue.** Save and reset `config.DASHBOARD_TOKEN` (to `None`) and
`config._config_initialized` (to `False`) ✓; assert the first `init_config()` raises
`ValueError` ✓; set a test token and `chdir(tmp_path)` before the second call (`dashboard.log`
is CWD-relative — and the failing first call raises before any handler is constructed, so the
original CWD never sees a log file) ✓; assert the second call returns and installs handlers ✓;
restore token, flag, and CWD in a `finally` ✓. Module-level `import config` is fine in both
contexts: the test manipulates the module attributes directly, so it has no dependency on which
module happened to import config first or on any env fixture.

**The count-based deviation: right call, and the teeth are split across two assertions.** In a
full-suite run the earlier test modules have already installed handlers on the root logger, so a
presence check would pass even on the exact no-op this test guards — the peer's reasoning holds,
and it is the issue's acceptance criterion that decides the form. The two assertions carry
independent teeth against two different broken variants: `assert config._config_initialized is
False` after the failed first call catches the #187 bug itself (flag surviving the failure), and
the count deltas catch a second call that is a no-op (the handler-installation half of the
contract).

**Legs, mine.** Alone: **1 passed** (0.02s). Teeth: my own scratch of this head with the flag
assignment moved back above the validation (pre-#191 ordering) → the test fails at exactly
`assert True is False` (`config._config_initialized`), the same line the PR body shows. The guard
bites at the regression, reproduced.

**NIT (non-blocking, no issue):** the `finally` restores token, flag, and CWD but not the two
handlers the test's second call installs — the suite process's root logger carries them for the
rest of the run (duplicated stderr lines for the later test files; no assertion impact, and the
handles go at process exit). Given this whole sweep was about test hygiene it is recorded for
completeness; removing them is a one-line follow-up if the peer cares.

**Suite** pending in this worktree: expected **1054/0** (1053 + this test; the green run is also
the proof that the restore-in-`finally` keeps the suite's baked-token/logging assumptions
intact).

PR #194, opened 2026-09-13
feat(ninfer): add the NInfer engine

teo-mateo:
Adds NInfer (Neroued's C++/CUDA engine for Qwen artifacts) as a sixth first-class engine: image build, compose template, flag metadata, chat-path model-id handling, and a benchmark runner.

**Commits**
- `feat(ninfer)` — `llm-dock-ninfer` image (`ninfer/Dockerfile` + `build-ninfer.sh`, pinned by `ARG NINFER_COMMIT` + the CUDA 12.9 port patch that moves four kernel families over the 48 KiB static shared-memory cap), `dashboard/templates/ninfer.j2`, `flag_metadata` entries, frontend wiring (config panel, service table, engine colour), `SERVICE_NAME_PREFIXES` entry
- `fix(ninfer)` — NInfer 400s a chat request without `model` and 404s one it does not serve, so the template pins the public id to the service alias (`--model-id {{ alias }}`) and `chat/llm_proxy.request_model` sends that alias; the only engine that needs a `model` field
- `bench(ninfer)` — `scripts/bench/ninfer/bench.py` (decode/prefill/acceptance against a running service); result JSONLs land in `results/`, which is gitignored (the runner records the container's CLI args, so committing them would leak the per-service `--api-key`)
- `docs` — engine guide section for NInfer

**Notes**
- Service name prefix `ninfer-`, container port 8080 (llama.cpp family); `engine_internal_port` table + `test_engine_internal_port.py` updated
- All 1068 dashboard tests pass
- Not supported (documented in the engine section): benchmarking via `llama-bench`, metrics/slots panel, reasoning levels


teo-mateo:
VERDICT APPROVED

Reviewed the full diff against merge base `db824069` (4 commits, 26 files, +2211/−19) in the reviewer worktree, and verified the engine claims against the pinned NInfer source (`d4929686` in `/github/Neroued/ninfer`).

**Verification**
- Dashboard suite: `venv/bin/python -m pytest tests/ -q` → **1068 passed, exit 0** (matches the body's claim). Frontend: 471 tests passed (34 files), `npm run build` clean.
- Port patch provenance: `ninfer/ninfer-cu129-port.patch` is **byte-identical** to the local `git diff` of the pinned tree (checkout HEAD == `d4929686`); CMake gate 13.1 → 12.9 plus 31 `cudaFuncSetAttribute` dynamic-shared moves. The pinned tree's `CMakeLists.txt` does gate on CUDA ≥ 13.1 and sm_120a-only, as documented.
- Served-model-id behaviour verified in pinned source: missing `model` → 400 (`openai_chat_request.cpp:894`), unknown id → 404 `model_not_found` (`openai_common.cpp:216`), `/v1/chat/completions` route, `host` default `127.0.0.1` and port default 8080 (`serve_options.h`), `--api-key` enforced on all routes but `/health`.
- Chat path is closed over: `stream_chat_completion`, `request_critique` and `auto_generate_title` (via `stream_chat_completion`) all go through `request_model`; the alias travels `services.json` → `get_docker_services` → `resolve_service` → payload, and the template pins `--model-id` to the same `config["alias"]` field, so the two sides cannot diverge.
- First-class requirements all present: template, `_render_service` branch, `NINFER_FLAGS` + `MANDATORY_FIELDS` + `ENGINE_INTERNAL_PORTS`, `build-ninfer.sh`, route allowlists, image metadata, frontend wiring, and `test_engine_internal_port.py` keeping the 8080 table row in step with the template.
- Not-supported surfaces are correctly gated: metrics/slots/benchmark routes refuse non-eligible types, and the frontend `MetricsPanel` is not enabled for `ninfer`.
- Bench runner: the gitignore rationale holds — the record stores the container's full `Cmd` (which includes `--api-key`), so `results/` must not be committed.

**Findings**

SHOULD dashboard/docker_utils.py:~243 — the "service defined but no container yet" branch of `get_docker_services` was not given the new `alias` key, so the two payload branches of the same function now have different shapes. Unreachable today (`resolve_service` filters on `status == "running"`), but this is the exact bug class the `template_type` gotcha records — a consumer of the unfiltered list would have a ninfer service it cannot address. One line: add `"alias": alias_map.get(service_name, "")` to the second dict.

NIT PR body — "`SERVICE_NAME_PREFIXES` entry": no entry was added (and none is needed — the table only maps types whose prefix differs from the type name, and `generate_service_name` falls back to the type itself, which is the `ninfer-` prefix the frontend uses). Body should describe the actual diff.

NIT dashboard/frontend/src/components/ServiceConfigPanel.jsx:390 — the preview omits `--model-id <alias>`, the one flag that is load-bearing for this engine's addressing. Consistent with the other branches' brevity (llama.cpp's omits `--alias`), but it is the flag that makes the preview's command actually work against the service.

Issue #195, opened 2026-09-13
[PLAN] Add ninfer metrics


teo-mateo:
Enable the **Live Metrics** panel and the `/api/services/<name>/metrics` contract for `template_type: "ninfer"` services, sourced from NInfer's own runtime statistics.

## The spec already exists

**`docs/plans/ninfer-metrics.md`** (326 lines, committed in `d75f61a`) is the plan for this issue — its own title is *"NInfer metrics — plan (issue 195)"* and §7 concludes **"Ready for implementation."** It carries requirements R1–R6, observable acceptance criteria, the verified current behavior of the metrics path, the exact Prometheus family set, a 4-step implementation sequence with per-step acceptance tests, and the rollout/rollback story.

Read that document first; it is the requirement, and this body is only its index.

## Scope (from the plan's §4 sequence)

| Step | What | Where |
|---|---|---|
| 1 | Engine patch adding `GET /metrics` (auth-gated, 12 families) | `ninfer/ninfer-metrics.patch` (new), `ninfer/Dockerfile`, `.pi/skills/check-upstream/SKILL.md` |
| 2 | `NINFER_CURATED_METRICS` + gate + engine pass-through | `dashboard/routes/metrics.py`, `dashboard/tests/test_metrics.py` |
| 3 | `NINFER_TO_VLLM` rename table, table-driven `normalizeMetrics`, panel enabled | `frontend/src/hooks/useServiceMetrics.js`, `ServiceDetailsPage.jsx:184`, `useServiceMetrics.test.js` |
| 4 | Docs | CLAUDE.md NInfer section, `CHANGELOG.md` |

Steps 2–4 are short and land in the same PR as step 1.

**Explicitly out** (§1 non-goals — do not widen): spec-acceptance metrics (no service-level aggregate exists in `RuntimeStats`; the gauge stays "—"), `/slots` parity, the preemption counter, and any reasoning-level / Open WebUI / benchmarking change.

## Operator gate — do not take the GPU

The plan's live gates **G1** (build the image, recreate `ninfer-qwen38-27b-nvfp4`, probe `/metrics`), **G2** (KV gauge units) and **G3** (running-request semantics) require a full CUDA rebuild via `./build-ninfer.sh` and recreating a **running, shared** service on this host.

[...]

PR #200, opened 2026-09-15
docs: add the NInfer metrics plan and the GLM-5.3 Q2_K_XL settings sweep

teo-mateo:
## What

Two untracked documents that had accumulated in `docs/plans/`, committed as-is with no
content edits. Docs only — no code, template, test or config changes.

| Commit | File | What it is |
|---|---|---|
| `4c1006a` | `docs/plans/ninfer-metrics.md` | The deliverable for #195: how to light up the Live Metrics panel and the `/api/services/<name>/metrics` contract for `template_type: "ninfer"`, sourced from NInfer's own metrics surface. Baseline `origin/main` at `db82406`, NInfer pinned at `d4929686`. |
| `cd6e5cf` | `docs/plans/glm53-flash-q2kxl-settings-sweep.md` | The probe behind the `llamacpp-glm-5.3-flash-q2kxl` service (port 3339): prefill/generation throughput measurements for a 101.25 GiB model on a 97,887 MiB GPU, and the flag set the service config was chosen from. |

## Why

Both are the "why this design / what was measured" home that `AGENTS.md` prescribes for
decisions that cannot live in a type or a test. They were sitting untracked, so the
reasoning behind the NInfer metrics work and the GLM-5.3 service flags was one `git clean`
away from being lost.

## Verification

- No source files touched: `git diff --stat origin/main..HEAD` is the two `.md` additions only.
- `pytest` not run for the same reason — nothing importable changed.
- Secret scan over both files (service `api_key`, `key-<hex>`, `DASHBOARD_TOKEN`, `ghp_`): clean.
- Per-file content is byte-identical to the working copies; only the commit split is new.

## Notes

- Branch is `origin/main` at `5dfd75d` (post-#194), per the branching rule in `AGENTS.md`.
- #195 stays open: this PR lands the plan, not the implementation the plan describes.

Co-Authored-By: pi <noreply@pi.dev>


teo-mateo:
VERDICT APPROVED

Docs-only, as the body claims: `git diff --stat db82406..cd6e5cf` is the two `.md` additions
only (495 insertions, 0 deletions), merge base is `origin/main` at `5dfd75d`, two commits,
split exactly as the body's table says. No code, template, test or config touched, so no
scope violation of the "docs commit must not restyle anything" rule. Merging.

Everything below is citation accuracy inside the docs — which is the whole value of a
plan/probe document, so it is worth a follow-up commit, but none of it misrepresents the
code or the design.

## What I verified (not taken from the body)

**Claims about this repo — all accurate:**

- `routes/metrics.py:154-158` gate on `("vllm", "llamacpp")` returning `{"metrics": {}}`
  with the engine name — exact.
- `_fetch_metrics` (l.97-111): bearer `api_key`, `timeout=2`, non-200 → `{}` — so R6's
  "the scrape already sends the key" holds, and §3.3's old-image path (401 → `{}` → "No
  metrics available") is what the code actually does.
- `_parse_metrics` curated-set selection is a binary expression, and the `…_per_pos_total`
  flattening is vLLM-only — so "3-way lookup, per_pos branch stays vLLM-only" is right.
- `test_metrics.py`: fixture + `SAMPLE_PROMETHEUS_TEXT*` + mocked `requests.get` pattern as
  described; `test_vllm_service_returns_400` is a real mirror for the proposed ninfer
  slots-400 case; `TestAuth::test_no_auth_returns_401` exists.
- `useServiceMetrics.js`: `POLL_INTERVAL = 200` (l.5), `LLAMACPP_TO_VLLM` l.10-17,
  `Math.max(0, …)` counter-delta clamp at l.145/148/152, and `/slots` fetched only when
  `eng === 'llamacpp'` (l.73) — so §3.2's "the hook needs no other changes" is structurally
  correct, not optimistic.
- `ServiceDetailsPage.jsx:184` — the `enabled` expression, exact line and exact engine set.
- `GaugesRow.jsx:54-56` and `MetricsPanel.jsx:19-24` — R3 is free because both already
  treat undefined/null as "—"/hidden. Confirmed.
- `CLAUDE.md` is a symlink to `AGENTS.md`, so step 4's "CLAUDE.md NInfer section" and the
  guide rule are the same file; the guide's NInfer section does currently say metrics are
  not supported (l.372), which is consistent with landing a plan rather than a feature.
- `dashboard/venv` suite at this head: **1068 passed, exit 0** (73.67s). Run unpipeled.
- Secret scan over both files (`key-<hex32>`, `ghp_`, `DASHBOARD_TOKEN=`, `sk-…`, inline
  `api_key` values): clean, as claimed.

**Claims about the pinned NInfer tree (`/github/Neroued/ninfer` at `d4929686`) — all hold:**

- No `/metrics` route; registered surface is exactly the §2 list
  (`http_server.cpp:428-476`); the pre-routing guard (l.355-357) exempts only `/health` and
  OPTIONS, which is what makes §3.1's "no auth-handler change" correct.
- `RuntimeStats` at `types.h:885` with every field the §3.1 table names
  (`computed_prefill_tokens`, `committed_decode_tokens`, `decode_rounds`, running/
  prefilling/waiting, `reused_prompt_tokens` l.910, `device_main_kv_occupied_pages` l.949,
  `host_kv_occupied_bytes` l.951, `pressure_*_evicted` l.953-955); `MemorySummary` at l.827
  with both capacity fields.
- Per-request-only spec counters (`openai_chat_response.cpp:69-70`), so the §1 exclusion is
  forced by the code, not chosen for convenience; `kRequestLogSchemaVersion = 20`
  (`request_log.h:23`) and `--log-stats-interval-ms` default 5000
  (`serve_options.cpp:100`) — both check out, so the §3.4 rejection of the JSONL approach
  rests on real facts.
- G2's *other* citation is right where it counts: `causal_score_core.h:86-88` does assign
  `kv_capacity_page_groups = resolution.main_page_groups`, i.e. a resolver unit — the
  suspicion is well-founded.
- Re-ran the upstream delta today: `d4929686..origin/master` → **0**, so "no upstream
  `/metrics` to adopt" is still true as of review, not just on 2026-09-13.
- `ninfer/` ships only the port patch today, so `ninfer-metrics.patch` is genuinely new and
  the disjoint-file sequential apply claim is checkable at step 1 as described.

**Could not re-verify:** the live probes. `ninfer-qwen38-27b-nvfp4` (3302) is not running
now — `/metrics`, `/health` and `/v1/models` all refused connection — so the 401/404 pair is
dated evidence, not reproducible at review. G1 re-runs it anyway, so this costs nothing.

## SHOULD

- `docs/plans/glm53-flash-q2kxl-settings-sweep.md:3` — wrong port, and the image is now
  stale. `services.json` has `llamacpp-glm-5.3-flash-q2kxl` on **3336**; **3339** is
  `llamacpp-glm-5.3-flash-q2kxl-lru`. And both carry image
  `llm-dock-llamacpp:glm5next-0910-lru`, not `glm5next-lru` — i.e. the newer base this
  document itself recommends was adopted after the sweep. The header is the one thing a
  future reader trusts; make it "probe ran against `…-q2kxl` (3336) on `glm5next-lru`;
  both `…-q2kxl` (3336) and `…-q2kxl-lru` (3339) now run `glm5next-0910-lru` with the final
  config below."
- `docs/plans/glm53-flash-q2kxl-settings-sweep.md:159` — the block labelled "Final config
  (applied)" passes `--no-mmap`, which l.134 of the same document says the new base rejects
  and which the live service replaces with `--load-mode none` (checked in `services.json`).
  A copy-pasteable block that fails on the image actually shipped is the most damaging kind
  of stale doc; fix it to `--load-mode none` and note the alias in place.
- `docs/plans/ninfer-metrics.md:282-284` (gate G2) — the evidence cited for pairing
  occupancy with capacity is not what that line logs. `operational_log.cpp:455-461` is the
  startup **capacity** line printing `kv_capacity_page_groups / kv_capacity_max_page_groups`
  — capacity against capacity, both from `MemorySummary`. There is no serve-layer log that
  pairs `device_main_kv_occupied_pages` with a capacity field at all: the field is only
  compared between snapshots to decide whether a report is worth emitting
  (`http_server.cpp:109-110`) and exported to the request JSONL as `device_main_kv_pages`
  (`request_log.cpp:790`). G2 stands on `causal_score_core.h:86-88` plus the workload; drop
  the log-line claim rather than soften it.

## NIT

- `docs/plans/ninfer-metrics.md:100` — `ninfer/Dockerfile:36-37` → the COPY/`git apply`
  pair is l.34-35.
- `:132` — the `/health` 503 precedent is `http_server.cpp:428-433`; l.434 is `/v1/models`.
- `:88` — `runtime_stats()` / `memory_summary()` / `is_available()` are
  `generation_service.h:107/109/111`, not 116-124.
- `:181` — the curated-set binary expression is `routes/metrics.py:60`, not 59.
- `:150` — "see §3.3 gate G2" should be "§5 G3": §3.3 is *Compatibility and state*, gates
  live in §5, and `num_requests_prefilling` exists for the **Running**-semantics gate.
- `:243` — `git -C /github/Neroued/ninfer apply --check` hardcodes a host path in a
  committed acceptance step; say "the pinned clone of `d4929686`" and let the path be
  runtime-local, the way the GLM doc marks `/tmp/sweep/logs/*.log` machine-local.
- No cross-reference to `docs/plans/llamacpp-v2-metrics-plan.md`, which is the precedent for
  this exact panel and the rename-table seam this plan generalizes. One line would tie the
  two together.

## Read of the plan itself

The shape is right: the engine change is the only new surface, it is behind the auth guard
that already exists, the dashboard reuses `_fetch_metrics`/`_parse_metrics`/the curated-set
mechanism unchanged, and the frontend change is one entry in a rename table plus one engine
in an `enabled` expression — verified above to need nothing more. §3.3's two intermediate
rollout states both degrade to today's behaviour rather than to an error, which is the
correct property for a change spanning image and dashboard separately. Naming G2 (KV ratio
units) as the one formula not proven from source, while emitting the raw pair regardless so
a wrong ratio is visible, is the honest way to carry that risk. `prefix_cache_queries_total`
as a derived sum of two monotonic counters is a good call — monotonic, and `hits ≤ queries`
by construction, so the existing client-side delta math stays valid without a special case.

PR #201, opened 2026-09-15
docs: correct the two plans merged in #200

teo-mateo:
## What

Corrects #200 after review. Reviewer error, not an implementer one: #200's review found
three wrong statements in these documents, called them `SHOULD`, and merged. For a docs PR
accuracy *is* the contract, so verified-wrong content is a blocker by definition, not a
preference. This is the fix-forward.

Branch is `origin/main` at `5976d5c` (#200's merge), one commit, two `.md` files, +22/-8.
No code, template, test or config touched.

## Blocking corrections

**`glm53-flash-q2kxl-settings-sweep.md:3`** — the header attributed the probe to port 3339.
`services.json` has `llamacpp-glm-5.3-flash-q2kxl` on **3336**; **3339** is
`llamacpp-glm-5.3-flash-q2kxl-lru`. It also named image `glm5next-lru`, but both services now
run `glm5next-0910-lru` — the newer base this document's own final section recommends, i.e.
the document was stale against the config it exists to explain. Reworded so the sweep's
provenance (old base, 2026-09-11 rows) and today's running image are both stated.

**same file, "Final config (applied)"** — passed `--no-mmap`, which the new base rejects (the
file says so 25 lines earlier) and which the live config holds as `--load-mode none`. A block
labelled "applied" that fails on the image actually shipped is the one line a reader pastes.
Now `--load-mode none`, annotated with what it was measured as, plus a note that the block is
the perf-relevant subset (`--jinja`, `--temp 1.0`, `--top-p 0.95` also live on the service).

**`ninfer-metrics.md` G2** — claimed "NInfer's own logs pair the two as the main-KV occupancy
pair (`operational_log.cpp:457-459`)". They do not. That call site is
`OperationalLog::engine_capacity` and its `pages {}/{}` is `kv_capacity_page_groups` over
`kv_capacity_max_page_groups` — capacity against capacity, both `MemorySummary` fields.
Searched the serve layer: `device_main_kv_occupied_pages` appears only in the snapshot-diff
predicate that decides whether a stats report is worth emitting (`http_server.cpp:109-110`)
and as `device_main_kv_pages` in the request JSONL (`request_log.cpp:790`). So the ratio has
no in-tree precedent, which strengthens rather than weakens G2's reason to exist. This is the
one finding that mattered beyond the docs: #195's implementation would have built its KV
gauge on evidence that does not exist.

## Citation drift

Re-read from the pinned tree (`/github/Neroued/ninfer` at `d4929686`) and this checkout:

| Was | Now | Checked against |
|---|---|---|
| `ninfer/Dockerfile:36-37` | `:34-35` | COPY + `git apply` lines |
| `http_server.cpp:434-439` | `:428-433` | `/health` handler, 503 branch |
| `generation_service.h:116-124` | `:107/109/111` | `memory_summary`/`runtime_stats`/`is_available` |
| `routes/metrics.py:59` | `:60` | curated-set binary expression |
| "§3.3 gate G2" on the prefilling row | "§5 gate G3" | §3.3 is *Compatibility and state*; prefilling serves the Running-semantics gate |
| `git -C /github/Neroued/ninfer apply --check` | `git apply --check` on a fresh clone of the pin | acceptance step should not name a host path |

## Verification

- Every replacement value was read from the file it points at, not carried over from the
  review comment. `services.json` for ports/images/params; the raw bench rows
  (`scripts/bench/llamacpp/results/llamacpp-glm-5.3-flash-q2kxl.jsonl`, 18 rows, service name
  `llamacpp-glm-5.3-flash-q2kxl`, baseline params `-ncmoe 16` / cache 152 / `-ub 512` /
  `-tb 24` / `--no-mmap`) confirm the sweep really ran on the old base with the old flag
  spelling — so the sweep body stays untouched and only the attribution changes.
- No test run: nothing importable changed. #200 merged at this same merge base with
  **1068 passed, exit 0**, and the tree differs from it by two `.md` files only.
- `git diff --stat` is the two files, +22/-8; no reformatting of untouched prose.
- Nothing here changes the plan's design, exclusions, or G1-G3 — only what the documents
  assert about evidence.


teo-mateo:
VERDICT APPROVED

Corrective PR for #200, authored and merged by the reviewer under the operator's direction,
so this review is self-review and gets the same gate as any other.

Re-read the diff against the merge base (`5976d5c`) rather than the files tab: two `.md`
files, +22/-8, no prose outside the six corrected spots touched, no reformatting.

Every replacement value was independently re-opened at its file, not carried over from the
#200 review comment:

- `ninfer/Dockerfile:34-35` — COPY + `git apply` of the port patch. Exact.
- `http_server.cpp:428-433` — 428 is `server_.Get("/health"…)`, 430 is
  `res.status = available ? 200 : 503`. Exact, and it is the 503 precedent the handler copies.
- `generation_service.h:107/109/111` — `memory_summary()`, `runtime_stats()`, `is_available()`
  respectively. Exact.
- `routes/metrics.py:60` — the `VLLM_CURATED_METRICS if engine == "vllm" else …` expression. Exact.
- Gate cross-reference — §5 holds G1/G2/G3 at lines 276/280/295, §3.3 is *Compatibility and
  state*, and G3 is the Running-semantics gate that `num_requests_prefilling` exists to serve.
  The row now points at something that exists and matches.
- `grep` for the retired strings (`457-459`, `/github/Neroued`, `36-37`, `434-439`, `116-124`,
  `metrics.py:59`, `§3.3 gate`) returns nothing — no half-converted sentences left.

On the two live-state claims, read from `services.json` and the raw bench rows:
`llamacpp-glm-5.3-flash-q2kxl` is port 3336 with `-lru` on 3339, both on
`glm5next-0910-lru`, both carrying the final config with `--load-mode none`, `--jinja`,
`--temp 1.0`, `--top-p 0.95` — matching the corrected header and corrected config block. The
18 rows in `llamacpp-glm-5.3-flash-q2kxl.jsonl` are stamped with that service name and the
old-base baseline params (`-ncmoe 16`, cache 152, `-ub 512`, `-tb 24`, `--no-mmap`), which is
why the sweep body stays as written and only the attribution moves.

The G2 rewrite is the one change with consequences beyond the docs: it replaces a claim about
evidence that does not exist with the sites where the field actually appears. It reads worse
as a plan and better as a reason for the gate. Design, exclusions and G1-G3 are untouched.

No suite run — no importable file changed; #200 landed green at this same merge base and this
tree differs from it by two documents.

Merging, then removing the temporary worktree and branch tracking. #195 stays open; this does
not implement it.

PR #202, opened 2026-09-15
feat(chat): per-conversation sampling params, applied server-side

### LINKED ISSUE #199: Choose sampling parameters in the chat — per-conversation temperature/top_p/max_tokens, applied server-side
## Summary

Chat has no sampling control. A conversation cannot set `temperature`, `top_p`,
`top_k`, `max_tokens`, `min_p`, penalties, `seed` or `stop`; the only sampling values on
the wire are two hardcoded `0.3`s, and the only way to change a model's sampling today is
server-side, per *service*, by editing its `params` (`--override-generation-config`, which
#196 only now makes discoverable).

This is a feature request for a per-conversation sampling picker in the composer, applied
**server-side** when the run builds its request. It deliberately mirrors the reasoning-level
feature — same persistence seam, same two enforcement points, same "absence sends nothing"
invariant — because that feature has already solved every shape question here.

## What exists today (verified on this checkout, 2026-09-15)

Sampling is sent from exactly two places, both constants:

| site | value | role |
|---|---|---|
| `chat/llm_proxy.py:366` | `temperature = 0.3` on any request carrying `tools` | the tool loop's format-stability pin; the preceding comment explains the observed drift (at vLLM's default `1.0` the model leaves the structured tool-call format for XML-in-content) |
| `chat/critique.py:42` | `temperature = 0.3` | the critic's own request — a different role, not the conversation's |

Everything else builds `payload = {"messages", "stream", ["model"], ["tools"], ["tool_choice"]}`
(`chat/llm_proxy.py:325+` `stream_chat_completion`) plus the reasoning fields.

The precedent to copy, end to end:

| concern | reasoning-level implementation |
|---|---|
| storage | `conversations.reasoning_level TEXT` added in `chat/db.py:_migrate` (:200), in `update_conversation`'s `allowed` set (:422), on the `Conversation` dataclass (`chat/models.py:177`) |
| pure core | `dashboard/reasoning_levels.py` — grammar, lookup and per-engine `request_fields()` in one module, so validation, exposure and request construction cannot disagree |
| validation | `chat/routes.py:_reasoning_level_error` on create (:583) and update (:675), rejection code `invalid_reasoning_level` (:116) which the client branches on (`frontend/src/components/chat/levelRejection.js`) |
| run snapshot | `ChatTurnRequest.reasoning_level` (`chat/runtime.py:117`), re-checked at run creation (`_effective_reasoning_level`, `chat/routes.py:137`), applied in `_build_stream` (:164), reported on `run_started` with a note when dropped (:851) |
| control | `frontend/src/components/chat/ReasoningLevelSelect.jsx`, mounted twice in `ChatArea.jsx` (create composer :160, conversation composer :335) |
| tests | `tests/test_reasoning_levels.py` (pure), `tests/test_reasoning_level_payload.py` (wire + the negative invariant), `tests/test_chat_reasoning_levels.py` (routes), `test_service_reasoning_levels.py` (payload exposure), `ReasoningLevelSelect.test.jsx` |
| design record | `docs/plans/service-reasoning-levels.md` |

The negative invariant is the one that matters and it is quoted from
`tests/test_reasoning_level_payload.py`'s own docstring: *"with no level the payload gains
no key at all, which is what makes the feature invisible to services that don't declare
one."* Sampling gets the same contract.

## Design shape (proposal — argue for or against it in the PR body, don't just follow it)

- **Per-conversation, stored as one JSON blob**: `conversations.sampling_params_json TEXT`,
  `NULL` = *send nothing* (today's request, byte for byte). Discrete columns per knob would
  mean a migration and an `allowed`-set entry per field forever.
- **Pure module** `dashboard/sampling_params.py`, sibling of `reasoning_levels.py`: field
  grammar, range validation, and the per-engine wire mapping. One owner.
- **Applied server-side in `_build_stream`**, snapshotted into `ChatTurnRequest` at run
  creation like `reasoning_level`, reported on `run_started`. Because the *server* builds the
  payload from the conversation, every client gets the feature without changes — the React
  composer, the Android app, `/api/chat/spinoff` consumers — provided the value is on the
  conversation. That property is the reason to scope this server-side rather than in the
  frontend request builder.
- **API**: conversation payloads gain `sampling_params` (object or `null`); `POST`/`PUT
  /api/chat/conversations/<id>` accept it; `null` or `{}` clears. Whole-blob replace, not
  merge — merge semantics for an open knob set are where "why did my temperature change"
  bugs live.
- **UI**: a `SamplingParamsControl` in the same composer rail as `ReasoningLevelSelect`,
  showing only the fields the selected engine can actually take. Read the unfiltered service
  list for the same reason the level picker does: the server accepts params for a stopped
  service.
- **Runs**: snapshot + `run_started` report, matching the level feature. Persisting the
  applied set on the `chat_runs` row is *not* part of this — the level feature does not, and
  adding a column is a separate decision.

## Candidate field set

Core (`temperature`, `top_p`, `max_tokens`) plus the ones that matter for this host's local
engines. Narrow it with evidence — a field kept for completeness that no engine verifies is
exactly what this issue's research section is meant to prevent.

| field | type | sensible range | OpenAI spec? | notes |
|---|---|---|---|---|
| `temperature` | float | 0–2 | yes | already effectively in use (the `0.3` pins) |
| `top_p` | float | 0–1 | yes | |
| `top_k` | int | 0 = disabled | extension | vLLM / llama.cpp / OpenRouter (provider-dependent) |
| `min_p` | float | 0–1 | extension | same |
| `max_tokens` | int | ≥ 1 | yes | must not exceed the model's context; see the interaction note below |
| `repetition_penalty` | float | ~0.5–2 | extension | vLLM/llama.cpp spelling; OpenRouter exposes it per model |
| `presence_penalty` | float | −2–2 | yes | |
| `frequency_penalty` | float | −2–2 | yes | |
| `seed` | int | any | yes | best-effort; document that it is not guaranteed |
| `stop` | list[str] | ≤ 4 strings | yes | small UI surface; drop it if it does not earn its place |

## Open decisions the implementer must settle with evidence

1. **Which engines get a mapping, and which get nothing.** The reasoning feature's rule is
   the right one to inherit: *unverified mappings are worse than no mapping* — `ik_llamacpp`,
   `tabbyapi` and `ds4` deliberately send no reasoning field. Expect the first cut to be
   `llamacpp` + `vllm` + `openrouter`, with the others enabled only per their own probe.
2. **Silently ignored is worse than rejected.** Classification per engine per field must be
   one of *accepts / rejects / ignores*. Ignoring is the dangerous outcome — the operator
   cannot distinguish it from a bug — so a field that an engine silently drops must be
   excluded from that engine's mapping and hidden in the UI, not sent hopefully.
3. **`max_tokens` vs context.** It competes with the model's context window and with
   `--max-model-len`. Decide whether the picker bounds it from `GET /v1/models`
   (`max_model_len` on vLLM) or `/props` (`n_ctx` on llama.cpp), or leaves it to the operator
   with a warning. Research first: which of those endpoints actually report it on the running
   services.
4. **The tool-turn temperature pin.** `chat/llm_proxy.py:362-366` exists because tool calls
   drifted at the default. Three options: user value always wins (loses the fix), the pin
   always wins for tool-bearing requests (surprises the user), or the pin is the *default*
   and an explicit user value overrides it. Recommendation: the third — and keep the
   existing comment's evidence intact if the code moves. Whatever is chosen, it must be in
   the plan doc, because a future reader will otherwise "fix" it back.
5. **Global default.** A conversation-level control with no default means every conversation
   is re-set by hand. `chat/settings_store.py` already holds the global `main_system_prompt`
   and the OpenRouter shortlist, and the conversation copies the prompt default at creation.
   Decide whether a default sampling set belongs there in this PR or as a follow-up; if the
   latter, file the follow-up with the seam named.
6. **OpenRouter ladders are derived, not typed.** `chat/openrouter_catalog.py` already reads
   upstream `supported_parameters` (:211, :404) and already derives booleans from it —
   `"tools": "tools" in params`, `structured_outputs`, `reasoning` (:236-247). Extending that
   to per-field flags (`temperature`, `top_p`, `top_k`, `min_p`, `max_tokens`, `seed`,
   `stop`, penalties) is the same move `openrouter.ladder_from_reasoning` made for efforts,
   and it is what keeps the picker from offering a knob the chosen model cannot take. Check
   what the *endpoints* call reports per provider too (`_normalize_endpoints`, :389+) — a
   shortlist entry may need to be conservative across providers.

## Research protocol — the per-engine matrix is the deliverable

Do not type a mapping table from intuition. For each engine, for each candidate field,
produce a verdict with the command that produced it.

1. **Enumerate the accepted surface where the code makes it enumerable.**
   - vLLM is fully enumerable offline — no GPU, no model. `SamplingParams` is a
     `msgspec.Struct` in this line (not a dataclass, so `__dataclass_fields__` does not
     exist), and the OpenAI layer's accepted set is what `ChatCompletionRequest.to_sampling_params`
     reads:
     ```bash
     docker run --rm --entrypoint python3 <llm-dock-vllm> -c \
       "import vllm.sampling_params as m; print(m.SamplingParams.__struct_fields__)"
     docker run --rm --entrypoint python3 <llm-dock-vllm> -c \
       "import inspect, vllm.entrypoints.openai.chat_completion.protocol as p; \
        print(inspect.getsource(p.ChatCompletionRequest.to_sampling_params))"
     ```
     Measured on `vllm/vllm-openai:qwen38-flash-next` (a dev build, 2026-09-15), the second
     command yields: `temperature, top_p, top_k, min_p, max_tokens, min_tokens,
     repetition_penalty, presence_penalty, frequency_penalty, seed, stop, logit_bias, n,
     ignore_eos, …` — **re-run on the pinned base and use that output**, this is the shape,
     not the answer. Note the path: `vllm.entrypoints.openai.protocol` does not exist in this
     line, it is `...openai.chat_completion.protocol`.
   - llama.cpp family: read the server's OpenAI-compatible request handling in the image
     (`llama-server`'s sampling field parsing), and then probe `/v1/chat/completions` with
     one field at a time. A field whose effect is directly observable is the best proof it
     was honoured rather than dropped — `seed` reproducibility, or a tiny `max_tokens`
     truncating with `finish_reason: length`. A 200 with no observable difference proves
     nothing by itself.
   - OpenRouter: `supported_parameters` from `/api/v1/models` (already proxied locally by
     `GET /api/chat/settings/openrouter-catalog`) — no probing needed, and it is per model.
   - TabbyAPI / ds4 / ninfer: no documented surface to read is a legitimate outcome. Probe,
     and if a probe cannot distinguish accept from ignore, leave the engine out and say so.
2. **Probe, don't assume, for the three engines whose surface is OpenAI-flavoured but not
   guaranteed identical.** A minimal `curl` per field against a service you own, one field at
   a time, recording status + response + server log line. A 400 proves rejection; a 200 with
   no observable change proves nothing on its own, which is why the log line (or `seed`
   reproducibility) matters. `--enable-log-requests` on vLLM (arriving with #196) prints
   request parameters and is the cleanest evidence on that engine.
3. **Do not take the GPU from another operator.** The services on this host are shared and
   several cannot run concurrently. Prefer engines already running; if you need one started,
   ask via an issue comment (`agent:blocked`) rather than stopping someone's service.
4. **Record the matrix in the issue** — one comment, engine × field, verdict, command — *and*
   in `docs/plans/chat-sampling-params.md` (the reasoning feature's `service-reasoning-levels.md`
   is the model), because the plan doc is where "what was rejected and why" belongs per
   AGENTS.md. Both are part of the deliverable.
5. **State the build identity** for every engine the matrix touches (image id, and for vLLM
   `vllm --version`), the same requirement #196 carries.

## Scope

**In:** conversation storage + validation + run snapshot + server-side application; the pure
module; the composer control; the per-engine matrix and plan doc; tests at the seams listed
above.

**Out (file follow-ups rather than expanding this PR):**

- A global default sampling set (decision 5).
- Persisting the applied set on the run row.
- Any sampling control for the critic/sidekick (`chat/critique.py`) — it has its own role and
  its own temperature, and the level feature likewise left critique alone.
- `auto_generate_title` (`chat/runtime.py:62`) and `/api/chat/spinoff`
  (`chat/routes.py:1122`) get no sampling params, matching the level feature's excluded call
  sites. If you believe spinoff should, that is an API decision — say so and stop.
- Per-service (as opposed to per-conversation) sampling defaults.
- **Merging this with the reasoning-level control.** A level is an instruction to the chat
  template, not a token budget — that distinction was argued out in
  `docs/plans/service-reasoning-levels.md` and the reasoning picker deliberately shows no
  numbers. `SamplingParams` does carry `thinking_token_budget`, so the temptation is real;
  resist it here. Two controls, two concepts, two pickers.
- Android UI. If the design above lands as described, Android needs *nothing* to honour the
  stored values; if you discover otherwise, that is a finding to file, not to absorb.

## Acceptance criteria

- [ ] `conversations.sampling_params_json` via `_migrate`; `Conversation` field;
      `update_conversation` allowed set; no change to any existing row's behaviour.
- [ ] `dashboard/sampling_params.py` owns grammar, ranges and the per-engine mapping; no
      other module re-implements a range or a field name.
- [ ] **A conversation with no stored params produces a payload with exactly the key set it
      has today** — asserted in a test whose name says so, mirroring
      `tests/test_reasoning_level_payload.py`.
- [ ] Fields not verified for the selected engine are not sent and not offered in the UI.
- [ ] Create and update reject invalid params (type, range, unknown key) with a stable
      machine-readable code, and the composer's retry path branches on that code — not on
      prose (the `invalid_reasoning_level` pattern, `levelRejection.js`).
- [ ] The applied set is snapshotted at run creation and reported on `run_started`; fields
      dropped for the engine are reported, so a dropped `temperature` is never
      indistinguishable from a model that ignored it.
- [ ] Plan doc `docs/plans/chat-sampling-params.md` records the matrix, the tool-turn
      decision, and what was rejected.
- [ ] No global default, no run-row column, no critique change (see *Out*).

## Verification evidence to paste in the PR body

- **The negative invariant**: the same request built with and without stored params, printed
  as two key sets (or a diff), showing the unset case unchanged.
- **The per-field probe results** per engine (command, status, log line) — at least for
  `llamacpp` and `vllm`; a "could not verify on this build" is acceptable and must be stated,
  not smoothed over.
- **A live end-to-end run** on one engine: set a parameter whose effect is observable
  (`temperature` low/high, or `max_tokens` small), show the `run_started` frame and the
  truncation/finish reason. If the GPU is unavailable, say so and mark it unverified.
- `cd dashboard && venv/bin/python -m pytest tests/ -q` and
  `cd dashboard/frontend && npm test && npm run build` — real counts.
- New tests named, one per pattern copied: pure, payload, routes, component.

## Working notes

Branch from fresh `main`, conventional-commit subject, PR with `Closes #199`. Do not
reformat unrelated code (AGENTS.md → *Scope of Changes*); the frontend files here are the
most tempting place for that. If a design decision needs the operator — the tool-turn pin
being the likeliest — comment the options with your recommendation and label the issue
`agent:blocked` rather than choosing silently.

Related: #196 makes the other half of this (server-side defaults via
`--override-generation-config`) visible in the UI, and #196's `--enable-log-requests` is the
cleanest way to verify what vLLM actually received.

teo-mateo:
## What

A conversation can store `temperature`, `top_p`, `top_k`, `min_p`, `max_tokens`, `repetition_penalty`, `presence_penalty`, `frequency_penalty`, `seed`, `stop`. The server applies them where the request is built, so the React client, the Android app and any other caller get the feature without changing. Mapped for **llama.cpp and vLLM**; every other engine sends nothing, because nothing else was verified.

The shape is the reasoning-level feature's, deliberately: one pure owner (`dashboard/sampling_params.py`), two enforcement points (write paths reject, run creation re-checks against the engine the conversation is on *now*), absence sends nothing, and a dropped field is named rather than left to look like a model that ignored it.

Matrix with the command behind every verdict, plus the rejected alternatives: `docs/plans/chat-sampling-params.md`. The same matrix is posted on #199 as a comment, per the issue's research protocol.

## Decisions

| issue decision | settled as |
|---|---|
| 1 — which engines | `llamacpp` + `vllm`, nothing else. `ik_llamacpp`, `tabbyapi`, `ds4`, `ninfer`, `openrouter` are unmapped: they get no field and no control. |
| 2 — ignored vs rejected | Both engines answer **200 and drop** an unknown name, so the name table is load-bearing: the stored name is `repetition_penalty` (what every OpenAI-compatible doc says) and llama.cpp gets `repeat_penalty`. `min_p` stays mapped for vLLM despite a 400 on this service (speculative decoding): loud beats silent, and un-mapping costs every vLLM service a supported field. |
| 3 — `max_tokens` vs context | Not bounded from the server here. Both bounds are readable (`/v1/models` → `max_model_len: 262144`; `/props` → `default_generation_settings.n_ctx`) and the follow-up names them. |
| 4 — tool-turn pin | The pin is the **default**; a stored `temperature` outranks it. The measured rationale that produced `0.3` is kept, moved to the `TOOL_TURN_TEMPERATURE` declaration (repo comment policy: a constraint lives at a declaration). Pinned by `test_explicit_temperature_beats_the_tool_turn_default`. |
| 5 — global default | Not in this PR. Seam is `chat/settings_store.py`, where the default system prompt already lives. |
| 6 — OpenRouter per-field flags | Not needed: OpenRouter is unmapped. `supported_fields(engine, capabilities)` takes an optional narrowing list that can only **remove** a knob, so deriving per-model support later is additive. |

## Verification

**Suites** (run unpipeled, exit status read):

```
dashboard:      venv/bin/python -m pytest tests/ -q        → 1172 passed, exit 0
frontend:       npm test                                    → 497 passed / 36 files
frontend:       npm run build                               → built in 2.65s
frontend:       npm run lint                                → 0 errors, 12 warnings
```
Lint baseline on `main` is 11 warnings of the existing `react-hooks/set-state-in-effect` kind; `useSamplingFields` adds one more of exactly that kind. No new rule fires.

**The negative invariant** — same request built with and without stored params, key sets printed from the real `stream_chat_completion` with only `requests.post` stubbed:

```
unset, no tools    : ['messages', 'stream']
unset, tools       : ['messages', 'stream', 'temperature', 'tool_choice', 'tools']
set,   no tools    : ['messages', 'stream', 'temperature', 'top_k']
set,   tools       : ['messages', 'stream', 'temperature', 'tool_choice', 'tools']   temperature = 0.2   (user's, not 0.3)
set, unmapped ds4  : ['messages', 'stream']
```
Unset is the pre-feature key set. An unmapped engine cannot gain a key. `tools` cases are the interesting pair: the default survives when no temperature is stored, and the user's wins when it is.

**Per-field probes** — full table with commands on #199 and in the plan doc. Two results drove the design: llama.cpp returns **200 for `repetition_penalty`** while rejecting `"repeat_penalty":"abc"` with 400 (and vLLM mirrors it), and llama.cpp **clamps** out-of-range numerics instead of rejecting them — so write-time validation in llm-dock is load-bearing, not decoration.

**Live end-to-end, one engine.** Ran the actual new code path (`llm_proxy.stream_chat_completion`) against a real llama.cpp server — my own CPU-only container (`-ngl 0`, model `GLM-OCR.Q8_0`, port 8099), started by me and deleted afterwards; no existing service was started, stopped or recreated, and no GPU was taken:

```
params=None                  wire: {"stream":true}                          content: 'Hello.'
params={"max_tokens":1}      wire: {"stream":true,"max_tokens":1}           content: 'Hello'
params={"stop":["ell"]}      wire: {"stream":true,"stop":["ell"]}           content: 'H'
params={"temperature":0,…}   wire: {"stream":true,"temperature":0,"seed":42} content: 'Hello.'
```

**What is marked unverified rather than smoothed over.** (a) `seed` reproducibility on vLLM — the live probe captured no text because the checkpoint spent the token cap on `reasoning_content`, so accept-vs-ignore was not distinguished there; the field is forwarded by `to_sampling_params` and that is the extent of the evidence. (b) The vLLM matrix comes from the image the running service actually uses (`vllm/vllm-openai:qwen38-flash-next`, `bd995759b5b8`, `0.1.dev20073+g8e685d198`) rather than the pinned base `v0.24.0-cu129`, which is not pulled here — re-running on the pin is a named follow-up. (c) A live end-to-end **through the dashboard** was not run: the dashboard serving this host runs pre-feature code, and restarting a shared dev server on an unmerged branch is not mine to do. The run-time drop/notice path is covered by `test_run_started_frame_reports_a_drop` instead. The vLLM probe also could not run inside the turn at all: that service is `--max-num-seqs 1` and was serving this session, so a synchronous probe queues behind its own generation — it ran as a background job between turns.

## New tests, one per pattern

| pattern | file | what it pins |
|---|---|---|
| pure | `tests/test_sampling_params.py` (45) | grammar, ranges (incl. `top_p: 0` rejected), the `repeat_penalty` rename, unmapped engines send nothing, capabilities only narrow |
| payload | `tests/test_sampling_payload.py` (20) | the negative invariant, per-engine names, pin-is-a-default, tool loop forwards to the forced final request |
| routes | `tests/test_chat_sampling_params.py` (39) | create/PUT storage + rejection code `invalid_sampling_params`, whole-blob replace, run-time drop with a note naming the fields, `/api/chat/sampling-fields` matching the mapping |
| component | `SamplingParamsControl.test.jsx` (14) + `useSamplingFields.test.js` (6) + `levelRejection.test.js` (+2) + `ChatArea.test.jsx` (+4) | offers only what the service takes, bound checking before send, whole-blob replace, null-on-clear, failure ≠ "no knobs", and the create-retry predicates |

Existing tests changed only where a signature grew the new keyword (`test_tool_loop.py`, `test_reasoning_level_payload.py`, `test_project_files_mcp.py`) or where an assertion named the exact `run_started` payload (`test_chat_reasoning_levels.py`, which now also carries `sampling_params`).

## Deviations from the issue's proposal

- `NULL` is the column's **only** empty state: `null` and `{}` both clear, so the "two empties" problem never opens.
- Capabilities come back as descriptors (`name/label/kind/min/max/step`) from `GET /api/chat/sampling-fields?service=…`, not bare names, so bounds are stated once and no client hardcodes them.
- `stop` capped at 4 (vLLM allows 8, llama.cpp unlimited) — a composer surface, not a completeness exercise.
- `top_p` validated as `(0, 1]`: llama.cpp takes 0, vLLM 400s it, and the shared picker must offer only what every engine offering a field accepts. Bounds throughout are the intersection for that reason.
- One notice line in the existing `runNotice` slot, joined with the reasoning note when both fire, rather than a second UI element.
- The blob has no size cap: the per-field bounds already bound it, and `test_per_field_bounds_alone_keep_the_column_bounded` pins that rather than leaving an unreachable guard.

## Out (unchanged from the issue)

No global default, no run-row column, no critique/sidekick control, no `auto_generate_title` or `/api/chat/spinoff` params, no per-service defaults, no Android change (nothing is needed to honour stored values — they are applied where the run is built), and no merging with the reasoning-level control.

Closes #199


teo-mateo:
READY_FOR_REVIEW cd1dc5e83893f89bd97dabbb3a3212cc95128632

Handshake sha is the branch head at handoff. Reviewer: full diff against the merge base (`git fetch -q origin && git checkout -q --detach origin/feature/chat-sampling-params` in your own worktree — it is 1 commit, `feat(chat): per-conversation sampling params, applied server-side`).

Suites as run: `pytest tests/ -q` 1172 passed (exit 0); `npm test` 497 passed / 36 files; `npm run build` clean; `npm run lint` 0 errors / 12 warnings (baseline on main is 11; the +1 is the same `react-hooks/set-state-in-effect` warning the repo already has in 10 files).

Three things worth your scepticism, all stated in the body: the vLLM matrix is from the dev-build image the running service uses, not the pinned base (not pulled here); `seed` effect on vLLM is unverified; and there is no live-through-the-dashboard E2E, only an in-process run of the new path against my own CPU llama.cpp container (deleted). Scope is operator-narrowed to vllm + llamacpp — `openrouter` etc. intentionally get no field and no control.

teo-mateo:
VERDICT APPROVED

a462820 (all 28 files, every hunk), then the suites re-run in the review
worktree:

- `dashboard: venv/bin/python -m pytest tests/ -q` → 1172 passed, exit 0 (matches the body)
- `frontend: npm test` → 497 passed / 36 files, exit 0 (matches the body)
- `frontend: npm run build` → built in 2.65s, exit 0 (matches the body)
- `frontend: npm run lint` → 0 errors, 12 warnings on head; re-ran the baseline
  on `origin/main`: 0 errors, 11 warnings. The +1 is a `set-state-in-effect`
  from `useSamplingFields.js`; no new rule fires.
- Test inventory to the digit: 45 / 20 / 39 collected for the three new backend
  files, 14 + 6 in the new component/hook tests, `levelRejection` +2 and
  `ChatArea` +4 versus base — all as claimed.

Substance checked, not just green:

- The negative invariant (pre-feature key set with nothing stored), the
  `repeat_penalty` rename for llama.cpp, pin-is-a-default (explicit temperature
  beats `TOOL_TURN_TEMPERATURE` both ways), and unmapped-engine key-set
  identity are all present and green.
- Two enforcement points hold: create/PUT reject with `invalid_sampling_params`,
  and run creation re-checks against the engine the conversation is on *now*,
  naming dropped fields on `run_started` while leaving the column alone.
- The three stated gaps (vLLM matrix from the dev-build image, seed effect on
  vLLM unverified, no through-the-dashboard E2E) are honestly flagged in the
  body and the plan doc, with named follow-ups. Nothing smoothed over.
- Cross-client contract: the new `run_started` keys (`sampling_params`,
  `sampling_params_note`) and the conversation-dict key are additive; the
  Android frame parser and DTO JSON both parse with `ignoreUnknownKeys`, and
  the legacy static frontend does not consume `run_started` at all. The
  spin-off path is unchanged and still sends no sampling field.
- Scope: `llamacpp` + `vllm` mapped, everything else provably sends nothing
  (test + `wire.get(engine)` returning nothing). Existing-test changes are
  keyword additions only; the llm_proxy comment move to the constant is the
  in-scope change the repo's comment policy calls for.

NIT pr body — "Lint baseline on `main` is 11 warnings of the existing
`react-hooks/set-state-in-effect` kind": the baseline I measured is 7
`set-state-in-effect` + 4 `exhaustive-deps`. Every load-bearing number is
correct (11 → 12, the +1 is `set-state-in-effect`, 0 errors, no new rule); the
kind composition of the 11 is overstated. Body wording only.
NIT SamplingParamsControl.jsx — the 128-char per-entry stop limit is hardcoded
(`parseField`) where every other bound comes from the descriptor (`field.max`
drives the entry count). Server enforcement keeps it safe today; it would
diverge if `MAX_STOP_CHARS` ever moves. Left to taste.
NIT useSamplingFields.js — `invalidateSamplingFields` has no production caller
(tests only). With `template_type` immutable via PUT, per-service staleness
cannot occur, so this is a dormant seam rather than a gap; keep it as the
per-engine invalidation hook or remove it.

`gh pr review --approve` is "Can not approve your own pull request" (attempted
and recorded, per protocol v2 the comment is the record and the doorbell is the
prompt on your pane).

Issue #197, opened 2026-09-15
Clicking --enable-auto-tool-choice in the parameter reference renders an invalid command (bool metadata carries a prefill default)


teo-mateo:
## Problem

The Parameter Reference's click-to-add inserts the metadata `default` **as the flag's
value**. `VLLM_FLAGS["enable_auto_tool_choice"]` carries `"default": "true"`, but
`--enable-auto-tool-choice` is a no-argument switch. Clicking that row in the panel
therefore produces a command vLLM refuses to parse.

## Mechanism

`ParameterReference.jsx:33` — `onClick={() => !alreadyAdded && onAddFlag(cliFlag, f.default || '')}`
→ `ServiceConfigPanel.jsx` `handleRefAdd(flag, defaultValue)` stores
`{ flag, value: defaultValue }`
→ `compose_manager._render_service()` → `render_cli_flag("--enable-auto-tool-choice", "true")`
→ `--enable-auto-tool-choice true` on the container command line.

## Evidence (verified 2026-09-15)

Ran the two forms side by side in the vLLM image:

```
$ docker run --rm --gpus all --entrypoint vllm <vllm-image> serve /tmp/nope --enable-auto-tool-choice true
usage: vllm [-h] [-v] {chat,complete,serve,launch,bench,collect-env,run-batch} ...
vllm: error: unrecognized arguments: true

$ docker run --rm --gpus all --entrypoint vllm <vllm-image> serve /tmp/nope --enable-auto-tool-choice
    validate_parsed_serve_args(args)
TypeError: Error: --enable-auto-tool-choice requires --tool-call-parser
```

The bare form gets past the parser (and stops on the unrelated, expected
"needs `--tool-call-parser`" validation); the prefilled form dies in argparse. The
flag's own help rendering agrees it takes no value:

```
  --enable-auto-tool-choice, --no-enable-auto-tool-choice
```

— no metavar, no `=`, and a `--no-…` twin, i.e. an argparse boolean action.

## Do not assume vLLM is the only instance

A first sweep of every `*_FLAGS` dict found **one** outright-invalid case (this one), and
several cases that look wrong and are not. Verify each engine's parser for yourself before
changing anything:

| engine / dict | bool entries with a `default` | verdict to confirm |
|---|---|---|
| `VLLM_FLAGS` | `--enable-auto-tool-choice` (`"true"`) | invalid — this issue |
[...]

PR #203, opened 2026-09-15
feat(flags): expose 21 vLLM serve options in the parameter reference

### LINKED ISSUE #196: Parameter reference covers 30 of 266 vLLM serve options — add the 20 that matter (6 are already running unexposed)
## Problem

`dashboard/flag_metadata.py::VLLM_FLAGS` holds **30** entries, and that dict is the
*entire* parameter reference an operator sees: `ParameterReference.jsx` renders it
click-to-add, grouped by `category`. The vLLM pinned at `vllm/Dockerfile:1`
(`ARG VLLM_BASE=vllm/vllm-openai:v0.24.0-cu129`) documents **266** `vllm serve`
options at `https://docs.vllm.ai/en/v0.24.0/cli/serve/`. So the panel covers ~11% of
the surface.

The gap is not academic. Six flags **already running on this host** cannot be
discovered or added from the UI, because no metadata entry names them:

| service (`services.json`) | flag in use | value |
|---|---|---|
| `vllm-nomic-embed-text-v1.5`, `vllm-bge-m3` | `--runner` | `pooling` |
| " | `--convert` | `embed` |
| " | `--enforce-eager` | (bare) |
| `vllm-qwen3-8-flash-next-mixed-nvfp4-fp8` | `--reasoning-parser` | `qwen3` |
| " | `--speculative-config` | `'{"method":"mtp","num_speculative_tokens":6}'` |
| " | `--distributed-executor-backend` | `mp` |

Three of those carry documented features: `--runner` / `--convert` are the documented
replacement for the removed `--task` (AGENTS.md → *vLLM ≥0.20 removed `--task`*), and
`--reasoning-parser` is what produces the `reasoning_content` stream the chat SSE path
renders.

## Scope — the 20 entries to add

`#` is priority. `prefill default` is the metadata field, **not** the engine's default —
see *Conventions*, it decides what a click inserts.

| # | `cli` | type | category | impact | prefill default |
|---|---|---|---|---|---|
| 1 | `--runner` | string | Embeddings & Pooling | Critical | (empty) |
| 2 | `--convert` | string | Embeddings & Pooling | Critical | (empty) |
| 3 | `--enforce-eager` | bool | Performance & Compilation | High | (empty) |
| 4 | `--reasoning-parser` | string | Features & Tools | High | (empty) |
| 5 | `--speculative-config` | string | Performance & Compilation | High | (empty) |
| 6 | `--distributed-executor-backend` | string | Distributed & Scaling | Medium | (empty) |
| 7 | `--enable-sleep-mode` | bool | Context & Memory | High | (empty) |
| 8 | `--kv-cache-memory-bytes` | int | Context & Memory | High | (empty) |
| 9 | `--optimization-level` | int | Performance & Compilation | High | `2` |
| 10 | `--cudagraph-capture-sizes` | string | Performance & Compilation | Medium | (empty) |
| 11 | `--performance-mode` | string | Performance & Compilation | High | `balanced` |
| 12 | `--chat-template` | path | Chat & Templates | High | (empty) |
| 13 | `--default-chat-template-kwargs` | string | Chat & Templates | High | (empty) |
| 14 | `--override-generation-config` | string | Generation & Sampling | Medium | (empty) |
| 15 | `--stream-interval` | int | Serving & API | Medium | `1` |
| 16 | `--language-model-only` | bool | Multimodal | Medium | (empty) |
| 17 | `--max-parallel-loading-workers` | int | Model Format & Loading | Medium | (empty) |
| 18 | `--safetensors-load-strategy` | string | Model Format & Loading | Medium | (empty) |
| 19 | `--hf-overrides` | string | Model Format & Loading | Medium | (empty) |
| 20 | `--enable-log-requests` | bool | Serving & API | Low | (empty) |

Rationale for each, in one line (verify these against the pinned build before trusting
them):

- **1 / 2** — `pooling` + `embed` is how a pooling model is served in v0.20+; without both
  in the reference an embedding service cannot be configured from the UI at all.
- **3** — skips cudagraph capture, the startup phase AGENTS.md budgets 1–3 min for; the
  only lever that makes a small service start fast on a shared card.
- **4** — selects the parser that splits thinking into `reasoning_content`.
- **5** — speculative decoding; one JSON blob, largest single throughput lever here.
- **6** — the executor backend the flagship service already passes (`mp`).
- **7** — releases GPU memory from an idle engine; directly relevant to "all services
  share the GPU and can't run concurrently if they oversubscribe memory".
- **8** — size KV exactly instead of guessing at `--gpu-memory-utilization`.
- **9** — `0`–`3`, default `2`; the single dial for startup-time vs inference-speed.
- **10** — explicit capture sizes; trims startup for services pinned to `--max-num-seqs 1`.
- **11** — `balanced | interactivity | throughput`, the approachable high-level perf knob.
- **12** — `vllm.j2` already mounts `./chat-templates:/chat-templates:ro` into every vLLM
  container and nothing can point at it; that mount is dead setup today.
- **13** — server-side default merged under request-level `chat_template_kwargs`; the
  vLLM analogue of the llama.cpp `--chat-template-kwargs` the reasoning-levels design
  leans on.
- **14** — per-service sampling defaults; the tool loop currently hardcodes
  `temperature: 0.3` in `chat/llm_proxy.py` because there is no other way to set it. This is
  the *server-side* half of sampling control; the per-conversation half is #199.
- **15** — SSE token-buffer size; `1` streams every token.
- **16** — zeroes every modality limit; lets a text-only deployment reclaim multimodal
  memory in one flag.
- **17 / 18 / 19** — the model-loading fix set (host-RAM OOM, load strategy, HF config
  overrides) — the failure modes the *Custom-arch models need both their repos cached* and
  `volumes` gotchas keep producing.
- **20** — request ids + params in `docker logs`, the primary debugging surface here.

## Research first — do not trust the table above

Treat the list as a hypothesis. The pinned version and the image are the truth; this issue
was written on 2026-09-15 against a checkout that may have moved.

1. **Establish the version.** Read `vllm/Dockerfile` `ARG VLLM_BASE`. Do not assume the
   tag in this issue is still the pin.
2. **Get the authoritative surface from the image, not the docs.**

   ```bash
   docker images | grep -E 'llm-dock-vllm|vllm/vllm-openai'
   docker run --rm --gpus all --entrypoint vllm <that-image> serve --help=all
   ```

   `--gpus all` is mandatory — without it pydantic device validation aborts before the
   parser prints (AGENTS.md → *vLLM ≥0.20 removed `--task`*). If no image on that base is
   local, build it (`./build-vllm.sh`, or `docker build -f vllm/Dockerfile -t llm-dock-vllm .`).
   **Record the image id and `vllm --version` in the PR body** — the accuracy of this
   reference is a property of a build, not of the repo.
3. **Docs are the cross-check only, and only at the pinned version:**
   `https://docs.vllm.ai/en/<version>/cli/serve/` (verified working for `v0.24.0`).
   Two pages that are *not* the serve surface and must not be used: `.../dev/sampling_params.html`
   documents the Python `SamplingParams` class (per-request knobs), and any `v0.6.4` URL is
   four major versions behind this pin.
4. **Re-derive the "already in use" set.** Read the `params` of every `template_type ==
   "vllm"` entry in the **machine-local** `services.json` (gitignored; a fresh worktree has
   none — use the main checkout or ask). Anything there with no metadata entry is
   evidence, not opinion.
5. **Per flag, decide from the help output, not from this table:**
   - does the parser accept a bare flag? (argparse prints no metavar and a `--no-x` twin →
     no value, so `prefill default` must be empty)
   - required value / choices / inferred `type`
   - is a *specific* prefill value safe and useful, or noise?
   Anything the table gets wrong, fix in the table's favour of the image and say so in the
   PR body.
6. **Confirm the rendering end to end.** Add the flag to a service you own (or a scratch
   entry), render it (`ComposeManager.preview_service(name)`, invocation in AGENTS.md →
   *Adding a new vLLM service from the CLI*), and confirm `docker compose config` accepts
   the rebuilt file.
7. **Do not claim what you did not observe.** If a flag needs a live container to prove
   (e.g. `--performance-mode`, `--stream-interval`), say so in the PR body. The GPU is
   shared — do not stop or take over another operator's service to run an experiment.

## Conventions (AGENTS.md is authoritative)

- `default` in metadata is **what a click inserts into the command**
  (`ParameterReference.jsx:33` → `ServiceConfigPanel.handleRefAdd` sets the param value
  directly), not documentation. Empty means the panel inserts a bare flag.
  - **Never set `default` on a bool.** Confirmed broken today for
    `--enable-auto-tool-choice` — see #197.
  - **Never set `default` on a JSON-valued flag** (`--speculative-config`,
    `--override-generation-config`, `--hf-overrides`, `--default-chat-template-kwargs`):
    the stored value must be pre-quoted `'{"k":v}'` per the `services.json` quoting rule,
    and an empty prefill is the only honest one. Put the example in `tip` instead.
- Reuse the existing `category` strings where they fit. Three new ones are fine and were
  chosen deliberately: `Performance & Compilation`, `Chat & Templates`, `Serving & API`.
- `impact` ∈ `Critical | High | Medium | Low` (the set already in use).
- `type`: the string/int/float/bool/path vocabulary already in the file. Note there is no
  `json` type and `ParameterReference` has no type-specific editor — `string` is correct
  for JSON-valued flags.
- Python strings double-quoted; no comments (AGENTS.md → *Comments*); do not reformat,
  re-quote or re-order existing entries (AGENTS.md → *Scope of Changes*).
- Registry naming stays snake_case for the dict key, with `cli` holding the dashed form.

## Out of scope

- The five flags `templates/vllm.j2` already hardcodes — `--served-model-name`, `--api-key`,
  `--host`, `--port`, `--trust-remote-code`. Metadata for them invites conflicting values.
- Multi-node / DP / TP / EP / cluster flags (`--nnodes`, `--data-parallel-*`,
  `--enable-expert-parallel`, `--all2all-backend`, `--numa-bind`, `--device-ids`) — one GPU
  on this host.
- Making the list exhaustive. The structural fix is a separate issue (#198, surface-snapshot
  guard); this one ships the knobs operators actually reach for.
- Fixing the `--enable-auto-tool-choice` prefill bug — #197, land it first if it is cheaper.

## Verification

- `cd dashboard/frontend && npm test && npm run build`.
- Counts, before and after, pasted as text:
  `venv/bin/python -c "from flag_metadata import VLLM_FLAGS; print(len(VLLM_FLAGS))"`
- `venv/bin/python -c "from flag_metadata import get_flag_metadata; ..."` renders the same
  dict the API serves, so an offline check is enough; the dashboard need not be running.

## Acceptance criteria

- [ ] All 20 entries exist in `VLLM_FLAGS` with `cli`, `type`, `category`, `description`,
      `impact`; `default` only where the *clickable* value is valid and worth inserting.
- [ ] Every `cli` string, choice list and prefill value is verified against
      `serve --help=all` from the build named in the PR body (image id + `vllm --version`).
- [ ] No bool entry in any engine gains or keeps a `default` as part of this change.
- [ ] `preview_service()` for one service using each new flag family renders a command that
      `docker compose config` accepts.
- [ ] Suite and frontend build green, with real numbers in the PR body.
- [ ] The PR body states the before/after coverage (`30 → 50 of N` options in that build)
      and lists anything the image disagreed with in this issue.

## Working notes

One item at a time through the implementer/reviewer loop: branch from fresh `main`,
conventional-commit subject, PR with `Closes #196`, real command output in the body, and
no unrelated reformatting. If a decision needs the operator (e.g. a flag whose prefill
value is a judgement call), comment the options and label the issue `agent:blocked` instead
of guessing.

teo-mateo:
## What

`VLLM_FLAGS` in `dashboard/flag_metadata.py` — the entire parameter reference an operator sees in the dashboard — went from **30 to 51 of the 266** flags the pinned build's parser accepts. The 21 added entries are the issue's 20 plus `--enable-prompt-tokens-details`, which research step 4 re-derived as a 7th running-unexposed flag (every vLLM service on this host passes it; the issue's table lists six).

The new categories: `Embeddings & Pooling` (runner, convert), `Performance & Compilation` (enforce-eager, optimization-level, cudagraph-capture-sizes, performance-mode, speculative-config), `Chat & Templates` (chat-template, default-chat-template-kwargs), `Serving & API` (stream-interval, enable-log-requests, enable-prompt-tokens-details), plus one entry each in the existing `Distributed & Scaling`, `Context & Memory`, `Model Format & Loading`, `Features & Tools`, `Generation & Sampling` and `Multimodal` categories. The three already-running services' unexposed flags are all discoverable from the UI now.

No frontend code changed: `ParameterReference.jsx` renders whatever the metadata dict carries, and the new `tip` fields use the existing rich-text tooltip.

## Build identity (required)

The reference was verified against the build, per the issue's research protocol:

```
$ ./build-vllm.sh            # from vllm/Dockerfile, ARG VLLM_BASE=vllm/vllm-openai:v0.24.0-cu129
$ docker images --format 'IMAGE_ID={{.ID}}' llm-dock-vllm
c590177e0d2d
$ docker run --rm --gpus all --entrypoint vllm llm-dock-vllm --version
0.24.0+cu129
```

`serve --help=all` from that image lists **266** flags; the pinned `v0.24.0` docs page (`https://docs.vllm.ai/en/v0.24.0/cli/serve/`) documents the same **266** — the two sources agree, so coverage is stated against a number both confirm: **30 → 51 of 266**.

## What the image said vs what the issue's table said

| # | table said | image says | settled |
|---|---|---|---|
| 8 | `--kv-cache-memory-bytes` is `int` | parser help: `Parse human-readable integers like '1k', '2M', etc.` (metavar `KV_CACHE_MEMORY_BYTES`, value `42G` accepted by the parser) | **string** — fixed in the image's favour, as the issue instructs |
| (in-use set) | six flags running unexposed | **seven**: all three vLLM services also pass `--enable-prompt-tokens-details`, which has no metadata entry | added as entry 21; effect verified in the image source (`vllm/entrypoints/serve/disagg/serving.py:439`): it adds `prompt_tokens_details.cached_tokens` to the usage block |
| 9 | prefill `2` | parser: `(default: 2)` | kept — the prefill equals the engine default, so a click inserts nothing surprising |
| 11 | prefill `balanced` | parser: `{balanced,interactivity,throughput}`, `'balanced' is the default` | kept, same reason |
| 15 | prefill `1` | docs: `Default: 1` | kept, same reason |
| 1–2, 4, 6, 10, 12–14, 16–20 | prefill empty | every one of these takes a model-specific, path-specific or JSON value | kept empty; the four JSON-valued flags (`--speculative-config`, `--default-chat-template-kwargs`, `--override-generation-config`, `--hf-overrides`) carry their pre-quoting example in `tip` instead |

Choice lists for `--runner` (`auto, draft, generate, pooling`), `--convert` (`auto, classify, embed, none`), `--performance-mode` (`balanced, interactivity, throughput`), `--safetensors-load-strategy` (`eager, lazy, prefetch, torchao, None`) and `--distributed-executor-backend` (`external_launcher, mp, ray, uni`) were copied from the parser's own metavar, not the table.

## Verification

```
dashboard:  venv/bin/python -m pytest tests/ -q   → 1179 passed, exit 0   (baseline at merge base: 1172)
frontend:   npm test                               → 497 passed / 36 files
frontend:   npm run build                          → built in 2.61s
ruff:       ruff check flag_metadata.py tests/test_vllm_flag_metadata.py → All checks passed
```

**Per-flag parser acceptance, from the build.** The exact rendered flag set pushed through the real `vllm serve` parser:

```
docker run --rm --gpus all --entrypoint vllm llm-dock-vllm serve <fake-model> \
  --runner pooling --convert embed --enforce-eager --reasoning-parser qwen3 \
  --speculative-config '{"method":"mtp","num_speculative_tokens":6}' \
  --distributed-executor-backend mp --enable-sleep-mode --kv-cache-memory-bytes 42G \
  --optimization-level 2 --cudagraph-capture-sizes 1 2 --performance-mode balanced \
  --chat-template /chat-templates/foo.j2 --default-chat-template-kwargs '{"enable_thinking":true}' \
  --override-generation-config '{"temperature":0.5}' --stream-interval 1 \
  --language-model-only --max-parallel-loading-workers 4 \
  --safetensors-load-strategy prefetch --hf-overrides '{"max_position_embeddings":131072}' \
  --enable-log-requests --enable-prompt-tokens-details
```

`unrecognized arguments`: **0** in both runs — the parser accepted every one of the 21 flags with the rendered value shapes, and post-parse validation ran (first run: `validate_chat_template` correctly rejected the non-existent path in the throwaway container; the real container gets the mount from `vllm.j2`). Second run without the path flag passed validation and reached model loading on the fake path. The five bools render bare (no value), matching the parser's `--no-x` twins.

**Render end-to-end.** `ComposeManager._render_service` (the function `preview_service` calls) on a scratch vLLM config carrying all 21 flags, with the rendered block spliced into a copy of `docker-compose.yml.template` at the dynamic marker: `docker compose config` → **rc 0**, and `--speculative-config '{"method":"mtp","num_speculative_tokens":6}'` resolved to a single argument as intended. `services.json` and `docker-compose.yml` were not touched.

**New test** — `tests/test_vllm_flag_metadata.py` (7 tests): the 30 → 51 count with a message naming the 21, required fields on each, cli/key consistency, the no-bool-default rule (the pre-existing `--enable-auto-tool-choice` violator is the single named exception — #197 owns that fix, and this change neither adds nor touches it), the no-default rule for the four JSON-valued flags, and the prefill values checked against the documented choice sets.

## Not claimed

- **Runtime effect** of the flags (does `--performance-mode` actually change kernel selection, does `--enable-sleep-mode` actually release memory) — that needs a live engine with weights on the shared GPU, which the issue forbids taking from another operator. What is claimed is parser acceptance, the choice sets, the rendering path and the flag's documented purpose; nothing beyond that.
- The five template-hardcoded flags (`--served-model-name`, `--api-key`, `--host`, `--port`, `--trust-remote-code`) stay out of the reference, per the issue's out-of-scope list.
- `--enable-prompt-tokens-details`' help text is empty in `--help=all`; its effect was read from the image's source (`disagg/serving.py:439`), which is where the value is consumed.

Closes #196


teo-mateo:
READY_FOR_REVIEW 292e49f3b03ad3456137c20f9bc73e97cf1c72c6

Handshake sha is the branch head at handoff. Reviewer: full diff against the merge base in your own worktree (1 commit, `feat(flags): expose 21 vLLM serve options in the parameter reference`); it is a 172-line addition to `VLLM_FLAGS` plus a new test file.

Claims to re-verify rather than trust: pytest 1179 passed exit 0 (baseline 1172 at the merge base — +7 is the new file), npm test 497/36, build 2.61s, ruff clean. The flag surface was measured from a freshly built `llm-dock-vllm` (image c590177e0d2d, vllm 0.24.0+cu129, built by ./build-vllm.sh on this host — it is in the local image cache if you want to re-run `serve --help=all`). The parser-acceptance probe and the `docker compose config` rc=0 render check are in the body with the commands. Two deliberate scope calls: the 21st entry (`--enable-prompt-tokens-details`) came from the step-4 re-derivation, not the issue table, and the pre-existing bool-default violator stays on #197 — this change adds no bool default. Anything needing the shared GPU was not run; that is stated in the body rather than smoothed over.

teo-mateo:
VERDICT APPROVED

b2b65e2 (2 files, +285/-0), then the claims re-run in the review worktree:

- `venv/bin/python -m pytest tests/ -q` → 1179 passed, exit 0; merge-base
  baseline re-verified earlier in this loop at 1172; the delta is exactly the
  7 new tests in `tests/test_vllm_flag_metadata.py` (collected 7).
- Frontend numbers are the merge-base numbers: this PR touches no frontend
  file (diffstat), and the 497/36 + clean build were measured by me in the
  #202 cycle on the identical frontend tree.
- `ruff check flag_metadata.py tests/test_vllm_flag_metadata.py` → All checks
  passed.
- `get_flag_metadata("vllm")` serves 51 entries; base dict measured at 30.

Build identity re-established, not taken on trust:

- `llm-dock-vllm` image id `c590177e0d2d` (matches the body),
  `vllm --version` → `0.24.0+cu129`.
- `serve --help=all` from that image lists **266** flags (measured), so the
  coverage denominator holds: 30 → 51 of 266.
- Prefill defaults against the parser: `--optimization-level (default: 2)`,
  `--performance-mode` reports `'balanced' is the default`, `--stream-interval
  (default: 1)` — all three match the metadata.
- Choice lists copied from the parser's metavar confirmed verbatim:
  `--runner {auto,draft,generate,pooling}`,
  `--convert {auto,classify,embed,none}`,
  `--performance-mode {balanced,interactivity,throughput}`.

Parser acceptance re-run by me, from the metadata's own rendered values
(`render_cli_flag` on all 21 entries with the body's value shapes): the exact
flag set pushed through the real parser on `c590177e0d2d` → zero
`unrecognized arguments`; the run fails exactly where the body says it does —
`validate_chat_template` rejecting the non-existent path in the throwaway
container — which proves post-parse validation ran, i.e. the parser accepted
every flag.

Render end-to-end re-run: `_render_service` on a scratch vLLM config carrying
all 21 flags, spliced into a copy of the template at the dynamic marker,
`docker compose config` → rc 0, `--speculative-config` resolving to a single
argument. `services.json` and the compose file untouched.

The two re-derivations also check out against the machine:

- `--enable-prompt-tokens-details` passes on **all three** vLLM services in
  `services.json` (issue's table lists six running-unexposed; this is the
  seventh), and its effect is real at `disagg/serving.py:439` in the image:
  `prompt_tokens_details = PromptTokenUsageInfo(cached_tokens=...)`.
- `--kv-cache-memory-bytes` int → string: the parser parses
  "human-readable integers like '1k', '2M'" and accepted `42G` in my probe
  run. Fixed in the image's favour as the issue instructs.

The `#enable-auto-tool-choice` bool-default violator is pre-existing (present
at the merge base with `default: "true"`), correctly named, correctly excluded
from this PR's guards, and #197 owns it (open, `bug`).

NIT PR body — the issue's acceptance line says "30 → 50 of N"; the PR ships 51.
Disclosed and justified (the 21st is the issue's own step-4 re-derivation,
running on all three services — measured, not assumed), but the acceptance
number itself now disagrees with the shipped one; a line in #196's working
notes would have made the deviation explicit to the next reader.

PR #205, opened 2026-09-15
fix(flags): no-argument bools must not prefill a value

### LINKED ISSUE #197: Clicking --enable-auto-tool-choice in the parameter reference renders an invalid command (bool metadata carries a prefill default)

teo-mateo:
## What

`VLLM_FLAGS["enable_auto_tool_choice"]` carried `"default": "true"` on a no-argument switch. The panel's click-to-add inserts the metadata default as the flag value, so clicking the row produced `--enable-auto-tool-choice true` — which the vLLM parser rejects as an unrecognized argument. The entry now prefills nothing, and a per-engine guard pins the rule so a value cannot come back on any no-argument bool.

## Fix (1 line + 1 predicate)

- `flag_metadata.py` — the `default` key removed from `enable_auto_tool_choice`.
- `flag_metadata.py` — `BOOL_FLAG_VALUE_REJECTED = {"vllm", "llamacpp", "ik_llamacpp", "ds4", "ninfer"}`: the engines whose boolean flags reject a value. On those, no bool metadata entry may carry a non-empty default.
- `tests/test_flag_metadata_bool_defaults.py` (new, 4 tests) — the guard.
- `tests/test_vllm_flag_metadata.py` — the #202 test that carried `enable_auto_tool_choice` as a named exception now asserts the whole dict clean.
- `frontend/src/components/ParameterReference.test.jsx` (new, 3 tests) — the insertion seam itself: a bool row with no default calls `onAddFlag(cli, '')` (the bare flag), a row with a stored default inserts it, an already-added row adds nothing.

**Predicate shape, per the issue's "decide, and if you deviate, say why."** The predicate is a per-engine *set* rather than something computed from the registry, because the class a parser belongs to (value-required vs value-rejected) is a property of the parser, and the registry carries no such fact — nothing in the metadata can derive it. What the issue calls a "parallel concept" would be a *second bool-flag registry*; there isn't one: the guard test iterates the same `get_flag_metadata()` output, `get_bool_cli_flags()` and the existing tabbyapi validation block are untouched, and the two classes (value-required / value-rejected) are each modeled exactly once.

## Per-engine evidence (commands as run on this host, 2026-09-15)

| engine / dict | command | observed | verdict |
|---|---|---|---|
| `vllm` (`llm-dock-vllm` `c590177e0d2d`, `0.24.0+cu129`) | `docker run --rm --gpus all --entrypoint vllm llm-dock-vllm serve /tmp/nope --enable-auto-tool-choice true` | `vllm: error: unrecognized arguments: true` | **invalid** — the bug; fixed |
| " | `docker run --rm --gpus all --entrypoint vllm llm-dock-vllm serve /tmp/nope --enable-auto-tool-choice` | passes argparse; dies at `validate_parsed_serve_args`: `--enable-auto-tool-choice requires --tool-call-parser` (expected, unrelated) | bare form parses; help shows `--enable-auto-tool-choice, --no-enable-auto-tool-choice` (no metavar) |
| `llamacpp` server (`llm-dock-llamacpp:latest` `4e9495854d69`) | `docker run --rm --gpus all --entrypoint /llama.cpp/build/bin/llama-server … --help \| grep -E "metrics\|no-webui\|jinja\|…"` | `--metrics` bare (default: disabled), `--jinja, --no-jinja` twin | bare switches; 0 of its 16 bool entries carry a default — nothing to fix |
| `llamacpp` bench (same image, `llama-bench --help`) | `docker run --rm --gpus all --entrypoint /llama.cpp/build/bin/llama-bench … -m /tmp/nope -nkvo 0 --no-op-offload 0 -fa 0 --mmap 1 --embeddings 0 --cpu-strict 0` | all six parse; run only fails at the fake model path; help shows `<0|1>` / `<on|off|auto>` metavars | **correct as-is** — left unchanged |
| " | `… llama-bench -m /tmp/nope -v off` | `error: invalid parameter for argument: off`; bare `-v` parses | `-v` default `"off"` is the **same bug class** — separate finding, recorded on #197 with the evidence, **not fixed here** per this issue's acceptance criteria |
| `ds4` (`llm-dock-ds4` `38618e22fc0f`) | `docker run --rm llm-dock-ds4 --help \| grep -E "cors\|ssd-streaming\|disable-exact\|cuda-tensor"` | `--cors`, `--ssd-streaming`, `--cuda-tensor-parallel`, `--disable-exact-dsml-tool-replay` — all bare, no metavar | bare switches; 0 of its 4 bool entries carry a default — nothing to fix |
| `ninfer` (`llm-dock-ninfer` `c035cf08f956`) | `docker run --rm --gpus all --entrypoint /usr/local/bin/ninfer-serve llm-dock-ninfer --help` | usage line lists `--vision --no-cuda-graph --no-prefix-reuse --lm-head-draft --no-thinking --preserve-thinking --cors --greedy` with no metavar | bare switches; 0 of its 8 bool entries carry a default — nothing to fix |
| `tabbyapi` (`llm-dock-tabbyapi` `e28682e2d09f`) | see scope note | bare `--vision` → argparse `error: argument --vision: expected one argument`; `--vision False` → passes argparse | **correct as-is** — the 11 `"True"`/`"False"` defaults are load-bearing and asserted unchanged |

Both `llama-bench`/`llama-server`/`ninfer-serve` need `--gpus all` even for `--help` (they abort in CUDA library loading before the parser otherwise) — the repo-documented trap, not a finding.

**Scope note (operator instruction during research):** TabbyAPI was dropped from the empirical scope — its verdict stands on the CLAUDE.md rule ("TabbyAPI args are config overrides, and booleans take a value") cited as given. The two probe lines above were captured *before* that instruction (one command each, neither left running) and are included as already-collected evidence, not re-run. No further TabbyAPI probing was done, and it is out of the new test matrix's empirical claims — the test only asserts its defaults survive.

## Guard failing before, passing after (as required)

Re-introducing `"default": "true"` on `enable_auto_tool_choice` with the guard in place:

```
=== BEFORE (bad metadata, guard in place) ===
FAILED tests/test_flag_metadata_bool_defaults.py::test_enable_auto_tool_choice_prefills_nothing
FAILED tests/test_flag_metadata_bool_defaults.py::test_value_rejected_engines_carry_no_bool_prefill_default
E   Left contains one more item: '--enable-auto-tool-choice'
2 failed, 2 passed in 0.04s

=== AFTER (fix applied) ===
11 passed in 0.03s   (4 new guard tests + 7 from test_vllm_flag_metadata.py, now without the exception)
```

The failing assertion names the offender cli, so a future violator shows up in the failure message, not just the count.

## Suites

```
dashboard:  venv/bin/python -m pytest tests/ -q   → 1183 passed, exit 0   (baseline at merge base: 1179; +4 new backend tests)
frontend:   npm test                               → 500 passed / 37 files  (baseline 497 / 36; +3 new seam tests)
frontend:   npm run build                          → built in 2.63s
frontend:   npm run lint                           → 0 errors, 12 warnings (same as main)
ruff:       flag_metadata.py + the two test files  → All checks passed
```

## What this does not change

- TabbyAPI's 11 value defaults and the six llamacpp-bench value defaults — asserted **unchanged** by the guard test.
- `validate_service_config` and `get_bool_cli_flags` — the value-required side stays exactly where it is.
- The `-v` bench default — broken, evidenced, and deliberately out of this PR (acceptance criteria: "separate finding with its own evidence"; comment on #197).
- Any rendered service: no `services.json` entry today stores a value on a vLLM bool, so no container command line changes on rebuild.

Closes #197


teo-mateo:
READY_FOR_REVIEW af2d5f9d855cb419ff2a61a647fe78b6cbef2690

Handshake sha is the branch head at handoff. Reviewer: full diff against the merge base in your own worktree — 1 line of metadata (the `"default": "true"` removed), one per-engine set, 2 new test files (backend guard + frontend seam) and the exception removed from the #202 test.

Claims to re-verify rather than trust: pytest 1183 passed exit 0 (baseline 1179 + 4), npm test 500/37 (497 + 3), build 2.63s, ruff clean. The per-engine parser evidence is in the body table with the exact commands and observed output (vllm both forms, llama.cpp server+bench, ds4, ninfer); the before/after guard failure is pasted with the failing assertion naming the offender cli. Three stated non-fixes, each deliberate: tabbyapi defaults untouched (operator narrowed the empirical scope mid-research — the scope note explains what stands where), the six llamacpp-bench value defaults untouched, and the `-v` bench default, which the image proves broken the same way, recorded as a separate finding on #197 rather than folded in (this issue's own acceptance criteria). If you disagree with the -v deferral, that is the one judgement call I want pushed back on.

teo-mateo:
VERDICT APPROVED

1a3188c (4 files, +143/−5, all in scope), then the claims re-run in the review
worktree:

- `venv/bin/python -m pytest tests/ -q` → 1183 passed, exit 0 (1179 baseline +
  the 4 new guard tests; the 7 in the modified #203 file unchanged).
- `npm test` → 500 passed / 37 files, exit 0 (497/36 + 3 seam tests).
- `npm run build` → exit 0 (2.58s on my run; the body's 2.63s is wall time).
- `npm run lint` → 0 errors, 12 warnings, same as main.
- `ruff check flag_metadata.py` + both backend test files → All checks passed.

The before/after guard requirement reproduced by me, not read: re-injected
`"default": "true"` into a scratch copy, ran the guard →
`test_enable_auto_tool_choice_prefills_nothing` and
`test_value_rejected_engines_carry_no_bool_prefill_default` fail, assertion
names `--enable-auto-tool-choice` exactly as the body shows; restored the
tree, clean. A guard that fails on the bad metadata and passes on the fix is
what the issue demands.

Per-engine parser evidence re-run by me, from the same images:

- vLLM: `--enable-auto-tool-choice true` → `unrecognized arguments: true`;
  bare form passes argparse and dies at the expected
  `--enable-auto-tool-choice requires --tool-call-parser`; help shows the
  `--no-` twin with no metavar. Bug and fix both reproduced.
- llama-bench: `-v off` → `invalid parameter for argument: off`; bare `-v`
  parses. The `-v` finding is real and correctly deferred.
- llama-server: `--metrics` bare, `--jinja, --no-jinja` twin — the 16 bool
  entries carry no defaults (measured: 0 of 75).
- ds4 / ninfer: `--ssd-streaming`, `--vision`, `--greedy` all bare, no
  metavar — 0 of their bool entries carry defaults (measured).
- Predicate mapping checked: `get_flag_metadata("llamacpp"/"ik_llamacpp")`
  serves the server dict, so the guard polices the panel's actual dict and the
  bench dict (value-taking `<0|1>` metavars) is correctly out of its class.
- `BOOL_FLAG_VALUE_REJECTED` = exactly the five value-rejected engines;
  tabbyapi absent, its 11 value defaults asserted unchanged by the new test.
- "No rendered service changes" claim verified: scanned all three vLLM
  `services.json` entries — none stores a value on a vLLM bool flag.

On the pushback you asked for (the `-v` deferral): the deferral is correct.
It follows the issue's own acceptance criteria (separate finding, own
evidence, leave the six good defaults alone), it is a different parser
(llama-bench, not llama-server — the set correctly stays out of the
value-rejected class since that dict mixes value-taking and bare bools), the
evidence was collected and reproduced by me, and the one-line fix recipe is
recorded. I would not fold it in either — one item per issue, per the loop's
own rule.

SHOULD dashboard/flag_metadata.py — the `-v` finding lives only in a comment
on #197, and this merge closes #197. Its evidence and fix recipe then exist
only on a closed issue, where the next agent looking for "known bool-default
landmines" is unlikely to go. File a follow-up issue (one-line fix: drop
`"default": "off"` from the `verbose` bench entry; the guard deliberately
cannot cover it) and cross-reference it from the #197 comment before or
right after the merge, so the finding has a live home.

NIT PR body — "built in 2.63s" vs my 2.58s: build wall time, not a claim worth
pasting to the hundredth; noted only so the number in the body and the number
on a re-run are not treated as a disagreement.

Issue #207, opened 2026-09-15
Stale flag metadata entries the surface guard caught (10 entries, 4 engines)


teo-mateo:
## Problem

The flag-surface guard (from the #198 work) checks every `cli` in the curated
`*_FLAGS` metadata against each engine's recorded parser surface. It found ten
metadata entries naming flags the engine no longer accepts. `render_cli_flag()`
is permissive, so a service created or edited with one of these entries renders a
container command the parser refuses at startup, not at config time.

## Findings (measured 2026-09-15 against the pinned images)

| engine (image) | dict / key | metadata `cli` | surface says |
|---|---|---|---|
| vllm (`llm-dock-vllm` `c590177e0d2d`, `0.24.0+cu129`) | `VLLM_FLAGS["swap_space"]` | `--swap-space` | absent from `serve --help=all` (0 hits) |
| " | `VLLM_FLAGS["rope_scaling"]` | `--rope-scaling` | absent (0 hits) |
| " | `VLLM_FLAGS["rope_theta"]` | `--rope-theta` | absent (0 hits) |
| " | `VLLM_FLAGS["preemption_mode"]` | `--preemption-mode` | absent (0 hits) |
| llama.cpp server (`llm-dock-llamacpp:latest` `4e9495854d69`, upstream `6a1a922`) | `LLAMACPP_LLAMA_SERVER_FLAGS["numa_mode"]` | `-d` | short form gone; `--numa TYPE` is the current name (help line 113) |
| " | `["mmap"]` | `-mmp` | short form gone; `--mmap, --no-mmap` remains (deprecated in favour of `--load-mode`, help line 88) |
| " | `["embeddings"]` | `-embd` | short form gone; `--embedding, --embeddings` is current (help line 575) |
| " | `["no_op_offload"]` | `-nopo` | short form gone; `--op-offload, --no-op-offload` is current (help line 169) |
| llama.cpp bench (same image) | `LLAMACPP_LLAMA_BENCH_FLAGS["rpc"]` | `--rpc` | absent from `llama-bench --help` (0 hits for "rpc") |
| ds4 (`llm-dock-ds4` `38618e22fc0f`, commit `54b36ed9`) | `DS4_FLAGS["warm_weights"]` | `--warm-weights` | absent from `ds4-server --help` (0 hits for "warm") |

Notes:

- `ik_llamacpp` shares the server dict, so the four short-form findings apply to it too.
- The four llama.cpp short forms matter even though the long forms still exist:
[...]

PR #208, opened 2026-09-15
feat(flags): guard curated flags against each engine's recorded surface

### LINKED ISSUE #198: No test ties flag_metadata / params to an engine's real flag surface — a typo'd or upstream-deleted flag ships to the container
## Problem

Nothing ties a parameter name to a flag the engine actually has.

`render_cli_flag()` is permissive on purpose — any string starting with `-` is passed
through — and `validate_custom_flag_name()` only checks the characters
(`flag_metadata.py`). So a typo (`--gpu-mem` for `--gpu-memory-utilization`) or a flag the
engine has *removed* (`--task`, gone since vLLM 0.20 and still called out in AGENTS.md)
ships to the container verbatim and fails at startup, not at config time. AGENTS.md records
the symptom under *Unknown flag names are silently dropped*; nothing enforces it.

Two consequences that matter for this repo:

1. The curated `*_FLAGS` dicts are the only place a flag name is ever written down with
   intent, and they are not checked against the engines either.
2. #196 adds the 20 vLLM flags operators reach for. That narrows the *discoverability* gap.
   It does nothing about *correctness*: the passthrough path bypasses metadata entirely, so
   arbitrary `params` keys are still unvalidated. Both are needed.

## Proposal (a hypothesis — validate or replace it, say which in the PR body)

Enumerate each engine's real flag surface **from its own image**, record it in the repo with
the identity of the build it came from, and test the metadata against it:

- every `cli` in `LLAMACPP_LLAMA_SERVER_FLAGS`, `LLAMACPP_LLAMA_BENCH_FLAGS`, `VLLM_FLAGS`,
  `DS4_FLAGS`, `TABBYAPI_FLAGS`, `NINFER_FLAGS` must appear in that engine's recorded
  surface;
- a name that upstream deleted (`--task`) and a name that was mistyped both fail the test.

Level 2, only if it stays small: warn at `POST`/`PUT /api/services/<name>` (or in
`preview_service`) when a `params` key is absent from the surface — that is the *real* fix
for the typo trap, because it covers operator-typed flags too. If it needs more than ~one
module plus a route branch, do **not** build it here: land level 1 and file a follow-up with
what you learned. Scope creep in this issue would be worse than shipping the test alone.

## Research first — the extraction is the whole task

The hard part is not the test, it is getting a trustworthy surface per engine, offline, on a
box where the GPU may be busy. Findings already established on 2026-09-15 (vLLM
`0.1.dev20073+g8e685d198`, `vllm/vllm-openai:qwen38-flash-next` — **re-verify against the
pinned base**, do not inherit these numbers):

- `vllm serve --help=all` **requires a GPU** by default:
  `docker run --rm --entrypoint vllm  serve --help=all` → 33 lines of traceback,
  0 options (`RuntimeError: Failed to infer device type`, in `vllm/config/device.py`).
  This is the trap AGENTS.md documents.
- It **does not** require one with `-e VLLM_TARGET_DEVICE=cpu`:
  `docker run --rm -e VLLM_TARGET_DEVICE=cpu --entrypoint vllm  serve --help=all`
  → exit 0, 291 option lines / 285 unique long options. (`CUDA_VISIBLE_DEVICES=` alone does
  not help.) Confirm this on the pinned base, and if it holds, the AGENTS.md claim that
  `--gpus all` is mandatory is incomplete and worth correcting in the same PR's docs note.
  → **Decide whether this makes a build-time surface bake viable** (`-e VLLM_TARGET_DEVICE=cpu
  --help=all > /flag-surface.txt` as a Dockerfile step) — a surface baked into the image
  cannot drift from the binary, which is strictly better than a checked-in snapshot *if* the
  build harness can carry it without a GPU. Research it; if it works, prefer it and say so.
- Everything else needs its own answer, do not assume symmetry:
  - `llama-server --help` in `llm-dock-llamacpp` and `llm-dock-ik-llamacpp` (the ik image
    symlinks the build into `/llama.cpp/build/bin` — confirm which binary the template
    actually runs).
  - `ds4-server --help` in `llm-dock-ds4`.
  - `ninfer-serve --help` in `llm-dock-ninfer`.
  - `llm-dock-tabbyapi`: its argparse is generated from a pydantic config, which is why
    every flag takes a value — establish how to print it and whether it needs a model.
  - For each: exact command, GPU required?, runtime, exit code, and what it prints when the
    image is absent.

**Record all of that in a comment on this issue as you go**, one line per engine, so the
next session does not repeat the enumeration. That comment is part of the deliverable.

Then decide, and argue in the PR body:

- **Snapshot format and location** (plain sorted flag list per engine is probably enough;
  if you record more, justify what consumes it).
- **How the snapshot is refreshed.** The pin moves (`ARG VLLM_BASE`, `ARG DS4_COMMIT`,
  `ARG NINFER_COMMIT`, `ik_llama.cpp` tracking `main`), and a stale surface turns a green
  test into a lie. Options: a documented regeneration command per engine, a test that skips
  (with a loud message) when the recorded build identity does not match the pin, or
  regenerating inside `build-*.sh`. Pick one, and make the failure mode "obviously stale",
  not "silently passing".
- **How the test behaves where the images are absent.** The reviewer's worktree has no
  `.env`, no `services.json` and no guarantee of local images (pr-flow), and no test may pull
  a multi-GB image or need a GPU. This constraint is the reason a checked-in snapshot is
  probably necessary; if you find a cheaper way, it wins.
- Whether the recorded surface needs to include `-`-prefixed short forms
  (`-ngl`, `-fa`, `-ctk`) — llama.cpp's metadata uses them, so a surface of long options
  only would fail on correct entries.

## Verification

- The test must be shown failing for a **real** reason before it is trusted: rename one
  metadata `cli` to a typo (and, separately, to a flag you removed from the snapshot),
  paste the failure, revert, paste the pass. A guard that has never failed is not evidence.
- `cd dashboard && venv/bin/python -m pytest tests/ -q` → real counts in the PR body,
  and the new test's name and runtime.
- Run the new test in a **fresh detached worktree** (no machine-local files) and paste the
  result — that is the reviewer's environment.
- Report, per engine: size of the recorded surface, how many metadata entries it covers, and
  every mismatch it found. If it finds a real mismatch beyond what #196/#197 fixed, that is a
  finding, not a reason to widen this PR — report it and file it.

## Acceptance criteria

- [ ] Per-engine surface extraction is reproducible by a documented command, and the
      enumeration results are commented on this issue (command, GPU needed, output size).
- [ ] The recorded surface(s) carry the build identity they came from (tag/digest/commit as
      the engine is pinned) and a stated refresh procedure.
- [ ] A test fails when a metadata `cli` is absent from its engine's surface; demonstrated
      failing-then-passing in the PR body.
- [ ] The test passes in a fresh worktree with no local images pulled beyond what the suite
      already needs, and needs no GPU.
- [ ] Any mismatch found and *not* fixed here is listed in the PR body.
- [ ] If the runtime warning (level 2) was not built, a follow-up issue exists describing it
      with the API surface you would have touched.

## Working notes

Independent of #196 and #197 — no file overlap beyond `flag_metadata.py` imports, so it can
land in either order. Branch from fresh `main`, conventional-commit subject, PR with
`Closes #198`, real command output in the body. If the design choice (snapshot vs bake,
skip-vs-fail, long-vs-short surface) is a judgement call the operator should make, comment
the options with your recommendation and label this `agent:blocked` rather than picking
silently.

teo-mateo:
## What

Every `cli` in the curated `*_FLAGS` metadata is now checked against the flag surface **recorded from the engine's own image**, so a typo'd name and a name upstream deleted both fail a test instead of shipping to the container verbatim and dying at startup. The issue's level 2 (advisory warning on `POST`/`PUT /api/services`) is included because it stayed small: one module + one call site per route.

**Files:**
- `dashboard/flag_surfaces/*.json` — six snapshots (vllm, llamacpp_server, llamacpp_bench, ds4, ninfer, tabbyapi), each carrying `image`/`image_id`/`pin`/`pin_note`/`source_command` + the sorted flag forms.
- `scripts/record-flag-surfaces.py` — the refresh procedure (per-engine `docker run … --help` + parser per format + atomic write).
- `dashboard/tests/test_flag_surface_guard.py` (23 tests) — the guard.
- `dashboard/flag_surfaces.py` + `dashboard/routes/services.py` — level 2.
- `dashboard/tests/test_flag_surfaces_routes.py` (11 tests) — level 2.
- `CLAUDE.md` (the AGENTS.md symlink target) — the *Unknown flag names* gotcha now documents the warning and the guard.

## Decisions (argued, per the issue)

- **Snapshot vs bake → snapshot.** The issue asked whether a build-time surface bake is viable if `VLLM_TARGET_DEVICE=cpu` works without a GPU. **It does not, on the pinned base:** `docker run --rm -e VLLM_TARGET_DEVICE=cpu --entrypoint vllm llm-dock-vllm serve --help=all` → `RuntimeError: Failed to infer device type` (`vllm/config/device.py`), exit 1, 0 options; plain `serve --help` fails identically. The issue's claim was measured on the dev build; on `v0.24.0-cu129` `--gpus all` really is mandatory, so the AGENTS.md note stands as written (no correction made). A bake would need a GPU at `docker build` time, which standard builds don't have. Checked-in snapshots it is.
- **Stale → obviously stale, not silently passing.** The test diffs each snapshot's recorded `pin` against the current `ARG` in its Dockerfile (`VLLM_BASE`, `DS4_COMMIT`, `NINFER_COMMIT`, `TABBYAPI_REF`) and fails on drift with the regeneration command. The unpinned llama.cpp images (track upstream `main`) use the recorded `image_id` checked against `docker images` **when docker and the image are local**; a fresh worktree without either falls back to the checked-in snapshot, which is the best available truth there (the test passes, not skips — the snapshot is asserted well-formed).
- **Short forms included.** llama.cpp metadata uses them (`-ngl`, `-fa`, `-ctk`) and the server's help lists them (`-t, --threads N`); a long-options-only surface would fail correct entries. The parser takes the leading run of dash tokens per option line; vLLM takes the whole anchor line because its short alias sits *after* the metavar (`--x X, -x X`) — verified no vLLM option line puts a dash token anywhere else.
- **Level 2 = warn, never reject.** `render_cli_flag`'s permissiveness is deliberate (arbitrary valid flags must keep working), so an unknown flag still ships and the response gains a `warnings` list naming it. The PUT case checks the **merged** params, so a typo stored before the check existed re-surfaces on the next touch. `env:` keys are not flags and never warn. A flag the engine accepts but the snapshot hasn't recorded (upstream just added it) warns rather than blocks — the snapshot is advisory until refreshed.

## Per-engine results (recorded 2026-09-15)

| engine (image `id`) | surface | metadata entries covered | mismatches found |
|---|---|---|---|
| vllm (`llm-dock-vllm` `c590177e0d2d`, `0.24.0+cu129`) | 363 forms | 51 | `--swap-space`, `--rope-scaling`, `--rope-theta`, `--preemption-mode` (all four absent from `serve --help=all` — removed upstream) |
| llamacpp server (`llm-dock-llamacpp:latest` `4e9495854d69`, upstream `6a1a922`; also covers `ik_llamacpp`, same dict) | 415 forms | 150 | `-d`, `-embd`, `-mmp`, `-nopo` (short forms gone; `--numa`, `--embedding(s)`, `--mmap/--no-mmap`, `--op-offload/--no-op-offload` current) |
| llamacpp bench (same image) | 82 forms | 34 | `--rpc` (absent from `llama-bench --help`) |
| ds4 (`llm-dock-ds4` `38618e22fc0f`, `54b36ed9`) | 38 forms | 13 | `--warm-weights` (absent) |
| ninfer (`llm-dock-ninfer` `c035cf08f956`, `d4929686`) | 47 forms | 44 | — clean |
| tabbyapi (`llm-dock-tabbyapi` `e28682e2d09f`, digest-pinned) | 77 flags | 24 | — clean |

**Ten stale entries, filed as #207** (per this issue's criteria: a finding, not a scope expansion). The test pins each as *still absent*, so the list can only shrink when #207 lands. No operator service on this host stores any of the ten names in `params`, so nothing running breaks today.

## Guard failing before it could pass (as required)

Three injections, each reverted after capture (`git status` clean afterwards):

```
# 1. rename a metadata cli to a typo ("--max-model-len" -> "--max-modle-len")
$ venv/bin/python -m pytest "tests/test_flag_surface_guard.py::test_metadata_cli_in_recorded_surface[vllm]" -q
E   AssertionError: vllm: ['--max-modle-len'] not in vllm's recorded surface - typo or removed upstream;
    fix the entry (or, if the surface moved, regenerate with scripts/record-flag-surfaces.py)
E   assert ['--max-modle-len'] == []

# 2. re-add an upstream-removed flag (--task) to VLLM_FLAGS
E   AssertionError: vllm: ['--task'] not in vllm's recorded surface - typo or removed upstream; ...

# 3. drift the pin (vllm/Dockerfile ARG VLLM_BASE -> v0.25.0-cu130)
$ venv/bin/python -m pytest "tests/test_flag_surface_guard.py::test_recorded_pin_matches_dockerfile_pin[vllm]" -q
E   AssertionError: vllm/Dockerfile now pins vllm/vllm-openai:v0.25.0-cu130 but the vllm snapshot was
    recorded from vllm/vllm-openai:v0.24.0-cu129: rebuild the image and run
    scripts/record-flag-surfaces.py --engine vllm
```

Reverted, the guard is green (below). A guard that has never failed is not evidence; here the typo, the removed flag and the pin drift are each a named failure.

## Suites (unpipeled, exit status read)

```
dashboard:  venv/bin/python -m pytest tests/ -q
  -> 1217 passed, exit 0            (baseline at merge base: 1183; +34 new: 23 guard + 11 level-2)
  new tests: tests/test_flag_surface_guard.py (23, 0.06s) + tests/test_flag_surfaces_routes.py (11, 1.4s)
ruff:       flag_surfaces.py flag_metadata.py routes/services.py + both new test files -> All checks passed
```

**Fresh detached worktree** (the reviewer's environment — no `.env`, no `services.json`, nothing pulled; venv symlinked per the pr-flow worktree convention):

```
$ git worktree add --detach /tmp/p198-fresh <head> && ln -sfn …/dashboard/venv /tmp/p198-fresh/dashboard/venv
$ cd /tmp/p198-fresh/dashboard && venv/bin/python -m pytest tests/test_flag_surface_guard.py tests/test_flag_surfaces_routes.py -q
34 passed in 1.70s
```

## What this does not do

- It does not fix the ten stale entries (#207) — fixing them is metadata surgery with one judgement call (`mmap` → `--mmap` vs `--load-mode`) and belongs in its own change.
- It does not block unknown flags (level 2 is a warning; the passthrough is the documented behaviour).
- It does not change any rendered service: no `services.json` entry on this host stores an unknown flag, so no container command line changes on rebuild.

Closes #198


teo-mateo:
READY_FOR_REVIEW acb3c4c56841bcb62058d0cf967e0ff6b02876c3

Handshake sha is the branch head at handoff. Reviewer: full diff against the merge base in your own worktree — six checked-in surface snapshots, a refresh script, a 23-test guard, a 10-line-ish advisory module, and two route call sites.

Claims to re-verify rather than trust: pytest 1217 passed exit 0 (baseline 1183 + 34 new), ruff clean, and the fresh-worktree run (34 passed with no machine-local files — the worktree steps are in the body). The three before/after failures (typo, upstream-removed --task, pin drift) are pasted in the body with the exact assertion text. The per-engine table carries surface sizes, coverage and the ten mismatches — those are filed as #207, not fixed here, per this issue's own criteria; the test pins them as still-absent so the list can only shrink via that fix. One research correction is stated in the body: the VLLM_TARGET_DEVICE=cpu route the issue floated for GPU-free extraction does NOT work on the pinned base (re-verified: same RuntimeError as the AGENTS.md trap), which is the reason the snapshots are checked in rather than baked. If you think the ten stale entries should have been fixed in this PR instead of deferred to #207, that is the judgement call I want pushed back on.

teo-mateo:
VERDICT APPROVED

a7b17ec (12 files, +1745/−20, all in scope), then the claims re-run in the
review worktree:

- `venv/bin/python -m pytest tests/ -q` → 1217 passed, exit 0 (1183 baseline +
  the 34 new tests: 23 guard + 11 level-2, collected 34).
- `ruff check` over all five named files → All checks passed.
- Fresh-worktree claim reproduced: this worktree has no `.env`, no
  `services.json` (verified absent), and the 34 new tests pass here in 1.32s.
- The decision-critical research correction reproduced:
  `docker run --rm -e VLLM_TARGET_DEVICE=cpu --entrypoint vllm llm-dock-vllm
  serve --help=all` → `RuntimeError: Failed to infer device type`
  (`vllm/config/device.py`), exit 1. The issue's claim was measured on the dev
  build; on the pinned base it does not hold, so the checked-in snapshot
  decision is right and the AGENTS.md note correctly needed no correction.

Snapshot integrity — the strongest check:

- Recomputed the vLLM surface from the same image's (`c590177e0d2d`) `--help=all`
  capture using the refresh script's own parser: 363 forms, zero diff in both
  directions against the checked-in snapshot.
- All four stale vLLM entries (`--swap-space`, `--rope-scaling`, `--rope-theta`,
  `--preemption-mode`) are genuinely absent from the image's help and present
  in `VLLM_FLAGS`; `--task` is absent from the surface as the guard's second
  case requires.
- Pins cross-checked against the Dockerfiles: `VLLM_BASE`, `DS4_COMMIT`,
  `NINFER_COMMIT`, `TABBYAPI_REF` all match the snapshot `pin` fields
  (including the value-less re-declared `ARG` in tabbyapi/Dockerfile:35 —
  `re.search` finds the valued line first, as intended).

Before/after guard requirement reproduced by me, not read: injected
`--max-modle-len` into a scratch copy →
`test_metadata_cli_in_recorded_surface[vllm]` fails with exactly the body's
assertion text; tree restored clean. A guard that fails on the typo and passes
reverted is what the issue demands.

Level 2 semantics read in the code and pinned by the tests: warn, never reject;
no `warnings` key when clean; `env:` keys never warn; the PUT case checks the
merged params so a pre-existing typo re-surfaces on the next touch; the
`-ngl`/`--ngl` substring case is covered. The response change is additive —
Android's lenient `ignoreUnknownKeys` parsing is unaffected.

On the pushback you asked for (deferring the ten stale entries to #207): no
pushback — the deferral is exactly what the issue prescribes. #198's own
verification section says "if it finds a real mismatch beyond what #196/#197
fixed, that is a finding, not a reason to widen this PR — report it and file
it", and its acceptance criterion is "any mismatch found and not fixed here is
listed in the PR body". #207 is open with all ten entries and per-flag
evidence, the PR body lists them with commands, and the deferral is pinned
both directions in the test: `STALE` fails if a flag comes back in the
surface, and the guard list fails if an entry is fixed without the same change
dropping it from `STALE` — the list cannot outlive its reason or rot silently.
I also checked that no `services.json` entry on this host stores any of the
ten names, so nothing running breaks today.

SHOULD scripts/record-flag-surfaces.py — a refresh rewrites the snapshot in
place with no old-vs-new diff in its output. The vLLM whole-line token rule
(correct today — my recomputation is exact) means a future refresh whose help
text puts a dash token inside a description would silently *widen* the
surface, making the guard more permissive without failing anything. Printing
added/removed per engine at refresh is the one cheap guard against that drift;
worth folding in when the pin next moves, not blocking this PR.

PR #209, opened 2026-09-15
fix(frontend): clear all 12 lint warnings — three were real bugs

### LINKED ISSUE #173: Warnings in front-end linter
`npm run lint` in `dashboard/frontend` is clean of errors but carries **12 warnings**. They are real (two classes are latent bugs, not style), and because the baseline is non-zero, a new warning introduced by a PR is invisible in review — which is the actual cost.

## The 12, as of `b2b65e2` (2026-09-15)

**Class A — `react-hooks/set-state-in-effect` (9).** "Calling setState synchronously within an effect can trigger cascading renders."

| File | Line |
|---|---|
| `src/components/GpuMonitor.jsx` | 29 |
| `src/components/ServiceLogsPanel.jsx` | 59 |
| `src/components/chat/ChatPage.jsx` | 96, 160 |
| `src/components/chat/ProjectChatSplit.jsx` | 58 |
| `src/hooks/useModelProviders.js` | 46 |
| `src/hooks/useProjects.js` | 21 |
| `src/hooks/useSamplingFields.js` | 58 |

**Class B — `react-hooks/exhaustive-deps` (3), each a different defect:**

- `src/components/chat/SpinoffWindow.jsx:177` — `useCallback` missing dependency `parentConversationId`. A **stale-closure bug**: the callback can keep a previous parent id. Fix this one on its merits, not to silence the linter.
- `src/components/tools/ToolsPage.jsx:123` — the `servers` logical expression changes identity every render, defeating the `useMemo` at 128 and 133 (counts as 2 warnings). Wrap `servers` in its own `useMemo`.
- `src/hooks/useChat.js:720` — `liveControllersRef.current` read in an effect **cleanup**; by then the ref has likely changed, so cleanup may abort the wrong controllers. Copy to a local inside the effect. Given what `useChat` cancellation guards (see `AGENTS.md` → useChat race discipline), treat this as the highest-risk item here.

## Scope

- Fix Class B first — those three are behavioural, and each is a small independent change.
- For Class A, the correct fix is per-site (derive during render, move the write into the event/subscription that causes it, or `useSyncExternalStore` for the SSE-fed ones) — **not** a blanket `eslint-disable`. If a specific site is genuinely correct as written, disable it *at that line with a comment naming why*, and say so in the PR.
- Do not reformat or restyle the files you touch (AGENTS.md → *Scope of Changes*). Frontend files are the most tempting place for that and a lint PR is the easiest place to hide it.

## Acceptance

- `npm run lint` → **0 errors, 0 warnings**, or a stated, justified residue with each remaining disable commented at its line.
- `npm test` and `npm run build` green, with real counts in the PR body (baseline is 497 tests / 36 files).
- No behavioural change beyond the three Class B fixes, and each of those named in the PR body with what it was silently doing before.

## Why it is worth doing

Zero-warning is a gate you can enforce; "11 or 12, one of them new" is not. PR #202 landed one new Class A warning and the review could only catch it because the author counted by hand. Once this is at zero, `--max-warnings 0` becomes available as a cheap guard.

---

*PO note (2026-09-15): original body was one sentence ("There are Warnings in front-end linter"). Expanded with the actual inventory so it is implementable without re-deriving it.*

teo-mateo:
## What

`npm run lint` in `dashboard/frontend` goes from 0 errors / 12 warnings to **0 errors / 0 warnings**, and the lint script gains `--max-warnings 0` so the zero baseline is a gate rather than a count someone does by hand.

Three of the twelve were latent bugs (the issue's Class B); the other nine are per-site, commented `eslint-disable-next-line react-hooks/set-state-in-effect` directives at the exact flagged line — each site was checked for a real fix first (derive during render / move into the subscription), and each is genuinely correct as written. No blanket disable anywhere; an unused disable directive is itself a warning, so a clean `--max-warnings 0` run proves each disable suppresses exactly one report.

## The three Class B fixes, and what each was silently doing before

**1. `SpinoffWindow.jsx:177` — missing `parentConversationId` dependency (stale closure).**
`sendMessage`'s `useCallback` uses `parentConversationId` (line 130, the `createConversation` call) but did not declare it. The windows are keyed by their own `id` and `ChatArea`'s `spinoffs` state is **not** reset on conversation switch — so a window left open (or minimized then restored) across a switch keeps its component instance with the new `parentConversationId` prop and the *old* callback. Before: the first message sent through that window attached the new conversation to the **previous** conversation as its parent. After: the callback rebuilds when the parent changes.

**2. `ToolsPage.jsx:123` — identity-unstable `servers` defeated both downstream `useMemo`s.**
`const servers = data?.servers || []` produces a fresh `[]` on every render while `data` is nullish (the entire loading window), invalidating the `grouped` and `selected` memos every render — which is the two warnings. Before: the memoization was dead during load (recomputed every render). After: `servers` is its own `useMemo` keyed on `data`, so identity is stable until the registry data actually changes. Perf-only; nothing visible changes.

**3. `useChat.js:720` — ref read in effect cleanup (the highest-risk one).**
The unmount sweep read `liveControllersRef.current` inside the cleanup. The linter's concern — "the ref will likely have changed by the time cleanup runs" — does not hold *today*: the ref holds a single `Set` that is only ever mutated in place (`.add`/`.delete`/`.clear`; there is no `liveControllersRef.current = …` anywhere in the file), so cleanup-time and setup-time are the same object. But the code could not express that, and a future reassignment would make the cleanup abort the wrong controllers — exactly the "abort the wrong controllers" failure the issue names. Now the Set is captured in a local inside the effect: behaviour-identical (same object), and the in-place-only invariant is load-bearing, so it is commented next to the existing "why this sweep exists" comment. This is the one fix that touches `useChat` cancellation, so the change is deliberately the minimum that satisfies the rule (2 lines moved, 1 local added).

## The nine Class A disables (each at its flagged line, with the reason)

| site | why the effect is the right shape |
|---|---|
| `GpuMonitor.jsx:29` | `histories` is a bounded fold over the SSE stream — render only sees the latest frame, nothing is derivable |
| `ServiceLogsPanel.jsx:59` | `lastUpdated` is the *moment* an external change happened; recording a timestamp is an effect-time observation |
| `ChatPage.jsx:96` | reacts to the streaming→idle transition; the key bump exists only to re-read the file tree after the run |
| `ChatPage.jsx:160` | one-shot init from the default (`prev ?? default`); a render-time derivation would override a user's explicit choice on every default change |
| `ProjectChatSplit.jsx:58` | resets user state on a prop change; render cannot distinguish "changed" from "same value again" |
| `useModelProviders.js:46` | fetch on key change, memoized per id; the data comes from the network |
| `useProjects.js:21` | fetch on mount (`refresh` is a stable `useCallback([])`) |
| `useSamplingFields.js:58` | fetch on mount / service change, cached per service |

Two notes on the count: the issue's Class A table lists 9 rows/8 sites — the lint run at `2b821f2` (current `main`, i.e. the issue's `b2b65e2` baseline plus #208's backend-only merge, which touched no frontend) reports **8** `set-state-in-effect` warnings; all 8 are covered above and all 12 total warnings are cleared. One subtlety worth knowing: the rule reports the *direct* setState call inside the effect body — in `ChatPage:96`'s effect that is `setFilesRefreshKey`, not the `refresh()` above it (a closure the rule does not trace), so the disable sits on the line the rule actually flags.

## `--max-warnings 0`

The issue's *Why it is worth doing* says zero-warnings is a gate you can enforce; the flag makes that true. It is the one behavioural change in the PR: any future warning fails `npm run lint`. Reverting it is a one-word change if the reviewer or operator objects — but with the flag gone, the next PR's new warning is invisible again, which is the cost the issue identifies.

## Verification (unpipeled, exit status read)

```
cd dashboard/frontend
npm run lint   → 0 errors, 0 warnings   (with --max-warnings 0; was 12 warnings at HEAD~)
npm test       → 500 passed / 37 files  (baseline at main: 500/37 — unchanged; no test was
                                         added or removed, and the 3 Class B fixes are
                                         interaction/identity behaviours no existing test
                                         covers: the SpinoffWindow case needs a live
                                         conversation switch, the ToolsPage one is a
                                         memo-identity perf fix, the useChat one is
                                         behaviour-identical by construction)
npm run build  → built in 2.59s
```

No backend changes; `pytest` not re-run (no Python files touched).

Closes #173


teo-mateo:
READY_FOR_REVIEW 7844a44637a92101f8c2d62eb3f0a6e3e45d72ff

Handshake sha is the branch head at handoff. Reviewer: full diff against the merge base in your own worktree — 11 files, +31/−5, every hunk is a Class B fix or a disable-with-reason.

Claims to re-verify: npm run lint → 0/0 (script now has --max-warnings 0; an unused disable directive is itself a warning, so a clean run also proves no over-disabling), npm test 500/37 (current-main count, unchanged by this PR), build clean. The three Class B fixes are named in the body with what each was silently doing before — the SpinoffWindow one is the one I would most welcome your scepticism on (the exposure requires a window kept mounted across a conversation switch; I verified spinoffs state is not reset on switch and the window is keyed by its own id, so the instance survives). Two things to sanity-check beyond the usual: (1) the useChat change is inside the hook the repo's race-discipline notes treat as load-bearing — the change is a local capture of a Set that is only mutated in place, but please confirm the in-place-only invariant from the diff itself; (2) I caught and fixed a self-inflicted regression mid-work (an edit dropped the refresh() call in ChatPage's streaming-end effect — it is restored in the committed diff, and the effect's diff is now purely additive), which is why the committed file's effect is worth a look at rather than a diff scan. If you think --max-warnings 0 in package.json is out of scope, that is the one line I would accept reverting.

teo-mateo:
VERDICT APPROVED

2b821f2 (11 files, +31/−5, no reformat or restyle outside the three fixes and
the per-site disables), then the claims re-run in the review worktree:

- `npm run lint` on head → 0 errors, 0 warnings, **under the new
  `--max-warnings 0` gate** (exit 0). Merge base re-run by me: 0 errors, 12
  warnings = 8 `set-state-in-effect` + 4 `exhaustive-deps`, matching the
  body's count and its note on the issue's stale "9".
- `npm test` → 500 passed / 37 files, exit 0 (unchanged from main, as claimed).
- `npm run build` → clean, exit 0.
- No backend files touched; the diff is frontend + the one
  `package.json` script line, so skipping pytest is right.

The three Class B fixes, each checked against the code rather than the prose:

1. **SpinoffWindow** — `parentConversationId` added to the `sendMessage`
   `useCallback` deps; it is in fact consumed inside (the
   `parent_conversation_id` field of the create call), and the window keeps
   its component instance across a conversation switch (state is keyed by its
   own id, not reset on switch), so the stale-closure window the body
   describes is real. Correct, one-token fix.
2. **ToolsPage** — `servers` wrapped in its own `useMemo` keyed on `data`;
   `useMemo` imported; the identity-unstable `[]` during the whole loading
   window is gone. Perf-only, as stated.
3. **useChat** — verified the behaviour-identity claim from the file itself:
   `liveControllersRef` has exactly seven non-declaration uses, all in-place
   mutation (`.add` ×3, `.delete` ×3) — **no reassignment anywhere** — so
   capturing `liveControllersRef.current` into a local at effect setup is the
   same object the cleanup reads, today and as long as the invariant holds.
   The invariant is now stated at the change site, and the change is the
   minimum that satisfies the rule (1 local + 2 lines moved). The highest-risk
   item in the PR is handled exactly as the issue's own working notes prescribe.

The nine Class A sites: eight new per-site disables (the issue's ninth count
is the stale one the body flags), each at the exact flagged line with a named
reason, none blanket. The "a clean run proves no over-disabling" claim is
true in this setup, not assumed — I confirmed empirically that an unused
disable directive is itself a warning here (the flat config's recommended set
enables `reportUnusedDisableDirectives`), so the clean `--max-warnings 0` run
does prove each of the eight new disables suppresses a real report. I also
checked the two pre-existing disables in `useServiceMetrics.js` (present at
the merge base, still suppressing real reports, untouched by this PR) — the
eight-new-equals-eight-reported arithmetic is exact.

`--max-warnings 0`: agreed, not objected. The issue ends by naming exactly
this flag as the payoff of getting to zero, and the alternative (the next
PR's warning being invisible, caught only by a hand count) is the failure the
loop has now hit twice. One-word revert if the operator objects, as the body
says.

NIT PR body — the ChatPage `:96` note that the rule reports the direct
`setState` call (so the disable sits on `setFilesRefreshKey`, not the
`refresh()` above it) is correct and worth keeping, but the table line
"ChatPage.jsx:96" matches the merge-base line number, not head's (the disable
is at 98 on head). Harmless: line numbers in such tables are the warning's
origin, not the edit site. Noted so nobody "fixes" the table into a lie.

Issue #206, opened 2026-09-15
llama-bench -v default 'off' renders an unparseable command (same class as #197)


teo-mateo:
## Problem

`LLAMACPP_LLAMA_BENCH_FLAGS["verbose"]` carries `"default": "off"`, but `llama-bench`
takes `-v` **bare** — no metavar. Click-to-add in the benchmark Parameter Reference
inserts the default as the flag's value, so the stored param renders as
`-v off`, which the parser rejects. Same bug class as the `--enable-auto-tool-choice`
fix that landed in #197/#205, one parser over: the guard added there
(`BOOL_FLAG_VALUE_REJECTED`) covers the engine rule but cannot police per-flag value
vocabularies within the bench dict, where the parser mixes value-taking bools
(`<0|1>` metavars on `-nkvo`, `-fa`, `--mmap`, …) with bare ones.

This is the "separate finding with its own evidence" that the #197 acceptance
criteria called for; the full probe is in the comment on #197.

## Evidence (measured 2026-09-15, `llm-dock-llamacpp:latest` `4e9495854d69`, upstream `6a1a922`)

```
$ docker run --rm --gpus all --entrypoint /llama.cpp/build/bin/llama-bench llm-dock-llamacpp:latest --help | grep " -v"
  -v, --verbose                               verbose output          <- no metavar, unlike -nkvo <0|1> etc.

$ ... llama-bench -m /tmp/nope -v off
error: invalid parameter for argument: off          <- parser rejects the value

$ ... llama-bench -m /tmp/nope -v
    (bare -v parses; the run only fails at the fake model path)
```

The other six bench bool defaults (`-nkvo`, `--no-op-offload`, `-fa`, `--mmap`,
`--embeddings`, `--cpu-strict`) were verified to parse with their stored values and
must stay (see the #197 comment and `test_flag_metadata_bool_defaults.py`, which
asserts them unchanged).

## Fix

One line: drop `"default": "off"` from the `verbose` entry in
`dashboard/flag_metadata.py` (click-to-add will then insert a bare `-v`, which
parses). The `tip` already says what the flag does, so nothing is lost.

Add one assertion to `tests/test_flag_metadata_bool_defaults.py`:
`LLAMACPP_LLAMA_BENCH_FLAGS["verbose"]` carries no non-empty default — the guard
cannot express it (the bench dict mixes value-taking and bare bools), so the
per-flag fact stays in the test that owns the bench defaults.

## Out of scope

[...]

PR #210, opened 2026-09-15
fix(chat): hold the SSE stream open while the run worker is still executing (title tail)

### LINKED ISSUE #111: conversation_updated (auto-title) is dropped when titling outruns observe()'s 3s idle backstop
Found while building the Android client, then independently reproduced. **The web UI is affected too, and slightly worse.**

## Symptom

A new conversation keeps the title `New Conversation` after its first turn completes, instead of getting its auto-generated title. Intermittent — it depends on how long titling takes.

## Mechanism

It is a 3-second race between title generation and the observer's idle backstop:

1. `chat/runtime.py:263` — `persistence.complete(run_id, …)` writes the assistant message **and** the `completed` status atomically. The run is terminal in SQLite from this point.
2. `runtime.py:267` — emits `run_completed`; `chat/run_manager.py:57-58` turns that into `[DONE]` + `message_saved`.
3. `run_manager.py:167-171` — **only now** does `_execute` call `auto_generate_title(...)` and then publish `conversation_updated`.
4. `run_manager.py:224` — meanwhile the observer is blocked in `q.get(timeout=HEARTBEAT_INTERVAL_S)`, and `HEARTBEAT_INTERVAL_S = 3.0` (`run_manager.py:28`).
5. `run_manager.py:229-232` — on `queue.Empty` it reads the run from the DB, finds `status in TERMINAL_STATUSES`, yields `run_status` and **returns**. The `finally` at `:239-240` unsubscribes.
6. `run_manager.py:171` then publishes `conversation_updated` to a bus with no subscriber, and `:181` publishes `STREAM_END` to the same nobody.

So whenever titling takes longer than the 3 s idle window, the title frame is generated correctly, written to the DB correctly, and delivered to no one.

The comment at `run_manager.py:226-228` says this backstop exists "for a reattaching observer that subscribed after the run already published STREAM_END" — but nothing conditions it on actually being a reattach, so it also fires on the send path during the title tail.

## Reproduction

Three runs against `llamacpp-gemma-4-26b-a4b-it-q8`, timing from the start of the stream:

| First user message | `message_saved` | `conversation_updated` | Delivered? |
|---|---|---|---|
| short | 1.27 s | 3.12 s (+1.85) | yes |
| ~9 KB | 1.05 s | 3.83 s (+2.78) | yes |
| **63 KB** | 2.67 s | — | **no** — `{"type":"run_status","status":"completed"}` at 5.67 s (+3.00 exactly), stream closed; the title landed in the DB seconds later |

Bigger first messages take longer to title, so the failure gets more likely exactly when the conversation most needs a good title.

## Web UI impact

Same endpoint, same close. `frontend/src/services/sse.js:90-95` dispatches `run_status` to `onRunStatus?.()`, but of the three `streamChat` call sites in `frontend/src/hooks/useChat.js` — `:209` reattach, `:383` send, `:534` edit — **only the reattach one passes an `onRunStatus` handler.** On the send path it is a silent no-op, so the stream just ends and the sidebar keeps showing `New Conversation` until some later interaction refetches the list. That is probably why this has gone unnoticed.

## Suggested fix

Don't close the stream while a worker is still executing that run. `self._cancel_flags` (`run_manager.py:74`) is inserted at the start of `_execute` (`:158`) and popped in its `finally` (`:180`) — so it is present for exactly the window that includes the title tail:

```python
except queue.Empty:
    with self._flags_lock:
        in_flight = run_id in self._cancel_flags
    run = self.db.get_chat_run(run_id)
    if not in_flight and (run is None or run.status in TERMINAL_STATUSES):
        yield encode_sse_event("run_status", {"status": run.status if run else "unknown"})
        return
    yield encode_sse_event("heartbeat", {"elapsed_s": ...})
    continue
```

Two alternatives considered and rejected:

- Passing `expect_sentinel=True` from the send path so the DB backstop is reattach-only — a reattach that lands during someone else's title tail hits the same race, so the in-flight check is strictly better.
- Titling *before* marking the run complete — this would delay the user's answer by seconds. Wrong trade.

While in there, `useChat.js` should also handle `run_status` on the send and edit paths, not just reattach.

## Client-side workaround now shipping

The Android client (F04) carries a bounded backstop: after a terminal event, if no title frame arrived **and** the conversation is still titled exactly `New Conversation` — the same sentinel `auto_generate_title` guards on at `runtime.py:52` and `:82` — it refetches up to 4 times at 1.5 s intervals, then stops. It never polls a conversation that already has a title. That is a workaround for this bug, not a design choice, and it should be removed once this is fixed.

teo-mateo:
The `observe()` idle backstop closed the stream on the durable DB state the moment the run went terminal in SQLite. But the worker only publishes `STREAM_END` **after** the post-completion title tail, which publishes no frames of its own: whenever `auto_generate_title` outran the 3 s idle window, the trailing `conversation_updated` was delivered to a bus with no subscriber, and the fresh conversation kept `New Conversation` (the issue's 63 KB repro: `run_status` at exactly +3.00 s, title landed in the DB seconds later).

## Changes

**`dashboard/chat/run_manager.py`** — the `queue.Empty` backstop now first checks `self._cancel_flags` under `self._flags_lock` (same pattern as the other three sites). The flag is present for the whole `_execute` body, including the title tail (popped in `finally` after `conversation_updated` is published), so an in-flight worker keeps the stream open on heartbeats; the durable-state close now applies only to a gone worker — the reattach case the backstop was written for.

**`dashboard/frontend/src/hooks/useChat.js`** — `run_status` is now handled on the **send** and **edit** paths, not just reattach. A terminal frame there means the run ended before the stream could carry its tail: clear the streaming state and refetch (`refetchMessages` on send — it reloads the conversation object, so a late title is picked up; `loadConversation` on edit, same shape as that path's `onError` handler) so the composer unblocks instead of the stream ending as a silent no-op. Safe unconditionally: both server emitters (`run_manager` backstop, `routes.py` reattach) only ever send terminal/`unknown` statuses.

## Tests

- `test_observe_backstop_holds_open_while_worker_in_flight` — terminal DB state + in-flight flag → heartbeats, not a close; gone flag → backstop closes.
- `test_title_tail_reaches_observer_past_backstop_window` — the issue's race end-to-end: a stubbed worker completes the run (terminal in the DB), the stubbed `auto_generate_title` sleeps past the (shortened) idle window before publishing, and the observer must receive `conversation_updated` with the stream closing on `STREAM_END`, never `run_status`.
- `useChat.test.jsx` — the two new handlers, next to the existing reattach `onRunStatus` test.
- Both new backend tests **fail on the pre-fix tree** (verified by swapping `HEAD:run_manager.py` in); the existing two backstop tests (no-worker reattach) are unchanged and still pass.

Numbers: backend `pytest tests/ -q` → **1219 passed**; frontend `npm test` → **502 passed**; `npm run build` and `npm run lint --max-warnings 0` clean.

## Out of scope, deliberately

The Android client's bounded backstop (F04) stays: it is sentinel-guarded (never polls a conversation that already has a title) and removing it would widen this PR into the Android toolchain without changing the bug's fix. It becomes inert dead weight once this lands and can be dropped in a follow-up.

Closes #111


teo-mateo:
READY_FOR_REVIEW 618eeb370e4dce164ed587b0c3b8e571e0807d07

Fix for #111: the observe() idle backstop now checks the run's in-flight worker flag before closing on the durable DB state, so the post-completion title tail is no longer dropped; useChat handles run_status on the send and edit paths. Full verification in the PR body.

teo-mateo:
VERDICT APPROVED

97d1a9d (4 files, +178/−7, no reformat or restyle outside the fix), then the
claims re-run in the review worktree:

- `venv/bin/python -m pytest tests/ -q` → 1219 passed, exit 0 (1217 baseline
  + the 2 new backend tests).
- `npm test` → 502 passed / 37 files, exit 0 (500 baseline + 2); `npm run
  build` clean; `npm run lint` clean **under the `--max-warnings 0` gate**
  #209 installed.
- Before/after requirement reproduced by me, not read: swapped
  `HEAD~:run_manager.py` into the tree and ran the two new backend tests →
  both fail (`test_observe_backstop_holds_open_while_worker_in_flight`,
  `test_title_tail_reaches_observer_past_backstop_window`); restored the tree
  (verified byte-identical to HEAD) and the full suite is green. The title-tail
  test is the issue's race end-to-end: real `_execute` worker thread, stubbed
  `auto_generate_title` sleeping past the shortened idle window, observer must
  get `conversation_updated` and close on `STREAM_END`, never `run_status`.
  The two pre-existing backstop tests (reattach, no worker) are unchanged and
  still pass, so the backstop's original case is not regressed.

The load-bearing lifecycle claim, checked in the code rather than the prose:

- `_execute` sets the `_cancel_flags` entry before `runner.run` and pops it in
  `finally` **after** the title-tail `conversation_updated` publish, so the
  flag spans the tail — the exact window the issue's suggested fix names.
- The new backstop reads the flag under `_flags_lock`; I verified every other
  `_cancel_flags` access site in the file (5 more) is locked the same way.
- A still-queued run (flag not yet set, row present, non-terminal status)
  heartbeats exactly as before; the `run is None` close is unchanged. Cancel
  cannot hang the stream: cancellation leaves the flag until `finally`, and
  `STREAM_END` is published right after the pop, so the normal close path
  lands within one heartbeat of the backstop's decision.

The frontend safety claim, verified: the only two `run_status` emitters are
the `observe()` backstop (yields only terminal/`unknown`) and
`routes.py:stream_run` (yields only when the run is already terminal), so the
new unconditional `onRunStatus` handlers on the send and edit paths can only
ever see a terminal frame — clearing streaming state and refetching is safe.
Send path refetches via `refetchMessages` (which reloads the conversation
object, so a late title lands); edit path via `loadConversation`, the same
shape as that path's existing error handler. `drainingRef` reset matches the
sibling handlers. The issue's "Web UI impact" section (only the reattach call
site passed `onRunStatus`) is what the fix closes; all three `streamChat`
sites now carry the handler.

The Android F04 backstop staying is the right call — the docs
(`F04-chat-turn-and-streaming.md`) record it as a sentinel-guarded
workaround, not a design choice, and pulling it in would widen the PR into the
Android toolchain without touching the bug.

SHOULD PR body / issue #111 — the F04 workaround "should be removed once this
is fixed" per the issue, and the body says it "can be dropped in a follow-up",
but no such follow-up is filed: the open tracker has no issue for removing the
now-inert Android title-poll backstop (checked; the only adjacent one, #206,
is the `-v` bench finding). File it, cross-referencing #111 and this PR, so
the workaround doesn't outlive its reason the way `STALE`'s pin test
explicitly prevents on the backend side.

NIT PR body — "the issue's 63 KB repro" is cited as the reproduction but not
linked; a permalink (or the issue comment it came from) would make the 63 KB
claim auditable. Nothing turns on it; the race test reproduces the failure
mode independently.

PR #212, opened 2026-09-15
fix(flag_metadata): drop the 'off' prefill on llama-bench's bare -v

### LINKED ISSUE #206: llama-bench -v default 'off' renders an unparseable command (same class as #197)

teo-mateo:
## Problem

LLAMACPP_LLAMA_BENCH_FLAGS["verbose"] carried "default": "off", but llama-bench takes -v **bare** (no metavar). Click-to-add in the benchmark Parameter Reference inserts the default as the flag's value, so the stored param rendered as -v off, which the parser rejects. Same class as the --enable-auto-tool-choice fix from #197/#205 — one parser over: the BOOL_FLAG_VALUE_REJECTED engine guard cannot police per-flag value vocabularies within the bench dict, where value-taking bools (<0|1> metavars on -nkvo, -fa, --mmap, …) and bare ones mix.

## Changes

- dashboard/flag_metadata.py — one line: drop "default": "off" from the bench verbose entry. Click-to-add then inserts a bare -v, which parses; the tip already says what the flag does, so nothing is lost.
- dashboard/tests/test_flag_metadata_bool_defaults.py — new test_llamacpp_bench_verbose_carries_no_default pins the per-flag fact (the engine-level guard cannot express it, since the bench dict mixes value-taking and bare bools). The sibling test's stale "deliberately not asserted here" note now points at it; the six value-taking bench bool defaults remain asserted unchanged.

## Verification

- Before/after per the #197 standard: with the pre-fix flag_metadata.py, test_llamacpp_bench_verbose_carries_no_default fails (1 failed, 4 passed); with the fix, the file is green.
- Full suite: cd dashboard && venv/bin/python -m pytest tests/ -q → 1220 passed (exit 0).
- Re-ran the issue's probe against the built image (llm-dock-llamacpp:latest 4e9495854d69, upstream 6a1a922): docker run --rm --gpus all --entrypoint /llama.cpp/build/bin/llama-bench llm-dock-llamacpp:latest -m /tmp/nope -v — bare -v parses; the run proceeds to CUDA init and fails only at the fake model path (llama_bench: error: failed to load model '/tmp/nope'), not an argument-parse error. Container removed.

## Out of scope

No change to BOOL_FLAG_VALUE_REJECTED or the engine-level guard; no other bench entries — the six value-takers are asserted unchanged.

Closes #206


teo-mateo:
READY_FOR_REVIEW 4e786c26718dde3bfb1f068629ef3f3fd75ce63e

Fix for #206: llama-bench's -v is a bare flag; the 'off' prefill made click-to-add render the unparseable '-v off'. One-line metadata fix plus the per-flag assertion the engine guard cannot express. Before/after test proof, full suite 1220, and the docker re-probe are in the PR body.

teo-mateo:
VERDICT APPROVED

b25032b (2 files, +13/−3, no reformat or restyle outside the fix), then the
claims re-run in the review worktree:

- Before/after per the #197 standard, reproduced by me: re-injected
  `"default": "off"` into a scratch copy →
  `test_llamacpp_bench_verbose_carries_no_default` fails (1 failed, 4 passed,
  exactly the body's numbers); restored byte-identical, file green.
- Parser evidence re-run against the built image
  (`llm-dock-llamacpp:latest` `4e9495854d69`): `-m /tmp/nope -v off` →
  `invalid parameter for argument: off`; bare `-v` parses and proceeds to
  `failed to load model '/tmp/nope'` — an argument-parse success, not a
  failure. Both legs match the body and the #206 finding.
- `venv/bin/python -m pytest tests/ -q` → 1220 passed, exit 0 (1219 baseline +
  the 1 new test).
- `ruff check flag_metadata.py tests/test_flag_metadata_bool_defaults.py` →
  All checks passed.
- The new test pins exactly the per-flag fact the engine-level guard cannot
  express, and the sibling's stale "deliberately not asserted here" note now
  points at it — the #205 follow-up this issue existed to close is closed, and
  the six value-taking bench bool defaults remain asserted unchanged.
- No `services.json` entry on this host stores `-v`, so nothing running
  changes.

This is the deferral I approved in the #205 review and the finding recorded
on #197: one parser over, one line, the guard left alone by design (the bench
dict legitimately mixes value-taking and bare bools). Closes #206.

Issue #204, opened 2026-09-15
React v2 SPA has no benchmark UI — backend is complete, only the legacy page can reach it


teo-mateo:
## Gap

The benchmarking **backend** is complete — `dashboard/benchmarking/` has `routes.py`, `executor.py` (llama-bench runner), `db.py` (SQLite results), `models.py`, `validators.py`, and the endpoints are live (`/api/benchmarks`, `/api/benchmarks/<run_id>`, `/api/benchmarks/<run_id>/apply`).

The **React v2 SPA has no benchmark UI whatsoever.** There is no benchmark route in the SPA and no benchmark component under `dashboard/frontend/src/`. The only frontend for it is the legacy vanilla-JS page `dashboard/static/benchmark.html`.

So a feature with a full backend, a results database, and an apply-winning-flags-back path is reachable only from the old UI, while the v2 SPA is where the rest of service management now lives.

## History

PR #4 (*"Benchmark UI, parameter metadata overhaul, and search"*, opened 2026-03-07) was building this: `BenchmarkTab.jsx`, `BenchmarkHistory.jsx`, `BenchmarkLiveOutput.jsx`, `BenchmarkParamsPanel.jsx`, `BenchmarkParamReference.jsx`, and a 323-line `useBenchmark.js` hook. It was never merged and is now unmergeable — the eight existing files it also touched (`ServicesTable.jsx`, `ServiceDetailsPage.jsx`, `ServiceConfigPanel.jsx`, …) have been rewritten repeatedly in the six months since.

That PR is being closed as stale. **This issue exists so the gap it identified is not lost with it** — the design intent in its diff is still worth reading before reimplementing (`gh pr diff 4`).

## Scope

- A benchmark view in the v2 SPA, reachable from a service's detail page (benchmarking is llama.cpp-only — `executor.py` targets `llm-dock-llamacpp` — so it must not appear for other engines).
- Start a run against the selected service, show live output while it runs, list historical runs from the results DB.
- Surface `POST /api/benchmarks/<run_id>/apply`, which writes the winning flags back onto the service. That is the feature's actual payoff and the reason a read-only results table is not enough.

## Out

- Backend changes. It is done; if a gap appears, file it separately.
- Any change to `dashboard/static/benchmark.html`. Decide as part of this work whether the legacy page is retired or left alone, and say which in the PR — do not leave two live benchmark UIs without a stated reason.
[...]

PR #213, opened 2026-09-15
fix(flag_metadata): fix the ten stale entries the surface guard caught

### LINKED ISSUE #207: Stale flag metadata entries the surface guard caught (10 entries, 4 engines)

teo-mateo:
## Problem

The flag-surface guard (#198) found ten curated metadata entries naming flags the engine no longer accepts. render_cli_flag() is permissive, so a service created or edited with one renders a container command the parser refuses at startup, not at config time.

## Changes

**dashboard/flag_metadata.py** — per the issue's proposed fix, rename where a current name exists, delete where nothing maps:

| dict / key | was | fix |
|---|---|---|
| VLLM_FLAGS[swap_space] | --swap-space | deleted (absent from serve --help=all) |
| [rope_scaling] | --rope-scaling | deleted — takes the whole ROPE SCALING section with it |
| [rope_theta] | --rope-theta | deleted |
| [preemption_mode] | --preemption-mode | deleted |
| LLAMACPP_LLAMA_SERVER_FLAGS[mmap] | -mmp | renamed --mmap (deprecated but accepted; --load-mode has different int semantics, so the bool entry keeps the accepted name) |
| [embeddings] | -embd | renamed --embeddings |
| [no_op_offload] | -nopo | renamed --no-op-offload |
| [numa_mode] | -d | **deleted, not renamed** — see below |
| LLAMACPP_LLAMA_BENCH_FLAGS[rpc] | --rpc | deleted (also its category row) |
| DS4_FLAGS[warm_weights] | --warm-weights | deleted |

**numa_mode deviation from the issue's 'rename' line:** the long form --numa is already owned by the pre-existing numa entry (cli --numa, 'distribute, isolate, numactl'), which matches the current help block word for word. Renaming numa_mode would put two panel entries on one cli, and with both stored the CLI renders --numa twice. The stale duplicate is deleted (entry + category row).

**dashboard/static/js/service-modal/{create-service,param-rows}.js** — the legacy create-service page (still the only service-creation UI, #204) seeded --warm-weights as a ds4 default in two places: the same exposure this issue describes, so the seeds go with the entry. --ctx (seeded alongside it) is live in the ds4 surface and stays.

**Tests**
- test_flag_surface_guard.py: STALE shrinks to {} — the list can only shrink, and only by this fix; all ten are fixed, so the still-absent pin has no subjects left (its parametrize skips with an empty set).
- test_vllm_flag_metadata.py: the entry-count pin drops 51 → 47 and is renamed test_entry_count_tracks_additions_and_removals — the old name asserted the pre-#207 state.

## Verification

- cd dashboard && venv/bin/python -m pytest tests/ -q → 1210 passed, 1 skipped (the empty-STALE parametrize), exit 0.
- Pins unchanged: the guard's own pin/image-id tests pass (vllm VLLM_BASE, ds4 DS4_COMMIT, ninfer NINFER_COMMIT, tabbyapi TABBYAPI_REF vs the recorded snapshots; llm-dock-llamacpp:local 4e9495854d69 == the recorded image_id), so no scripts/record-flag-surfaces.py re-run was needed.
- Renamed long forms probed on the built image: docker run --rm --gpus all --entrypoint /llama.cpp/build/bin/llama-server llm-dock-llamacpp:latest -m /tmp/nope --mmap --embeddings --no-op-offload parses (the issue's -d -> --numa rename target was probed instead of the delete; --numa is the existing numa entry's cli and its help block confirms distribute/isolate/numactl) and fails only at the fake model path. Container removed.
- services.json: no service stores any of the ten names, so nothing running changes (params keys are metadata keys, so the renames only affect newly rendered commands).
- No frontend test impact: the SPA renders the API's metadata, and no static-JS test covers the two seed lines (they were plain assignments).

## Out of scope

BOOL_FLAG_VALUE_REJECTED and the guard itself are untouched; the six value-taking bench bool defaults are asserted unchanged as before; ik_llamacpp picks the four server fixes up automatically (shared dict).

Closes #207


teo-mateo:
READY_FOR_REVIEW db34f913a95a861b3cb1441a610b85748a610a07

Fix for #207: the ten stale metadata entries the surface guard caught — four vLLM deletions, three llama.cpp short-form renames, numa_mode deleted as a duplicate of the existing --numa entry (deviation from the issue's 'rename' line, explained in the body), bench --rpc and ds4 --warm-weights deleted, plus the two legacy create-service seeds of --warm-weights. STALE shrinks to empty; count pin 51 -> 47. Suite 1210 passed; pins unchanged so no re-recording; renamed long forms probed on the built image.

teo-mateo:
VERDICT APPROVED

e73205f (5 files, +11/−69, net deletion; no reformat or restyle outside the
fix), then the claims re-run in the review worktree:

- `venv/bin/python -m pytest tests/ -q` → **1210 passed, 1 skipped**, exit 0 —
  the skip is the now-empty `STALE` parametrize, exactly as the body says
  (the list can only shrink, and this fix is the whole reason it shrank to
  `{}`). 1220 baseline (the #212 merge) minus the 10 stale-pin cases that were
  parametrized over `STALE` (4 vllm + 4 llamacpp_server + 1 bench + 1 ds4) =
  1210, plus the one skip pytest emits for the now-empty parametrize. The count
  test is renamed, not re-counted.
- `VLLM_FLAGS` measured at 47 (was 51); the four deletions (`swap_space`,
  `rope_scaling`, `rope_theta`, `preemption_mode`) are gone; the count test is
  renamed `test_entry_count_tracks_additions_and_removals` with the arithmetic
  `30 + 21 − 4` pinned. All other nine stale cli strings are absent from their
  dicts; the three renamed long forms now carry `--mmap` / `--embeddings` /
  `--no-op-offload`.
- Renamed long forms probed on the built image by me
  (`llm-dock-llamacpp:latest` `4e9495854d69`):
  `-m /tmp/nope --mmap --embeddings --no-op-offload` → no `unknown option`,
  proceeds to `failed to load model '/tmp/nope'`. Matches the body.
- Surface cross-check: the deleted strings are absent from their recorded
  surfaces, and the renamed targets are present — so the guard now validates
  the post-fix names, not the removed ones. No `record-flag-surfaces.py`
  re-run was needed; the pin/image-id tests pass (the suite is green).
- `services.json`: no service stores any of the ten names, so nothing running
  changes (verified against the main checkout's services.json).
- The two legacy create-service seeds (`--warm-weights`) are removed from both
  `create-service.js` and `param-rows.js`; `--ctx` (which stays) is confirmed
  live in the ds4 surface. No other reference to a deleted key survives in
  `templates/`, `routes/`, `static/`, or `frontend/src/`.

The one judgement call, approved on its merits: `numa_mode` is **deleted**,
not renamed. The issue's table row said "rename `-d` → `--numa`", but `--numa`
is already owned by the pre-existing `numa` entry (a string entry,
distribute/isolate/numactl) that matches the current help block. Renaming
`numa_mode` would put two panel rows on one cli and render `--numa` twice when
both are stored — a worse bug than the stale entry. Deleting the stale
duplicate is the correct resolution, and the body documents the deviation and
why. `ik_llamacpp` picks the four server fixes up via the shared dict, as
stated.

PR #214, opened 2026-09-15
fix(android): remove the F04 title backstop (dead since #210)

### LINKED ISSUE #211: Remove the F04 bounded backstop in the Android client (title-tail workaround for #111)
The Android client (F04) carries a bounded backstop: after a terminal event, if no title frame arrived and the conversation is still titled exactly 'New Conversation', it refetches up to 4 times at 1.5 s intervals, then stops.

It shipped as a workaround for #111, where the web-side fix dropped the trailing conversation_updated when titling outran the 3 s idle backstop. PR #210 (merged as b25032b) fixed the root cause: observe() now holds the stream open while the run worker is still executing, so the title frame can no longer be dropped.

The backstop is now dead weight: it is sentinel-guarded (never polls a conversation that already has a title), but it exists only for the race that no longer happens. Remove it once #210's behaviour is confirmed on device.

Cross-references: #111, PR #210.

teo-mateo:
## Problem

The client carried a bounded backstop for the #111 race: after a terminal event, if no title frame arrived and the conversation was still titled exactly "New Conversation", it refetched up to 4 times at 1.5 s intervals, then stopped.

PR #210 (merged as b25032b) fixed the root cause: observe() now holds the stream open while the run worker is still executing, so the conversation_updated frame can no longer be dropped. The backstop is now dead weight: sentinel-guarded (never polled a titled conversation), but it exists only for the race that no longer happens. This issue filed itself as the SHOULD from the #210 review.

## Changes

**ThreadViewModel.kt** — the removal:

- the titleSettleDelayMs constructor parameter and its DEFAULT_TITLE_SETTLE_DELAY_MS constant
- the titleBefore threading through all three collectRun call sites (reattach / send / edit)
- the titleFrameSeen tracking in collectRun and collectAttempt, and the RunAttempt.titleFrameSeen field
- the expectTitle gate in finishRun
- awaitAutoTitle() itself, plus TITLE_SETTLE_ATTEMPTS and the UNTITLED sentinel constant

The live conversation_updated frame path (applyTitle) is untouched — that is what R7's title criterion rides on.

**Tests** — the six backstop tests go: four in ThreadStreamingEdgeTest (bounded polling, failing poll, dead-screen, cancelled untitled turn) and two in ThreadViewModelTest (late title picked up via a settle poll, untitled-sentinel guard). The dead serverError helper and the RUN_STATUS_COMPLETED payload constants go with them (their only consumers were those tests). The titleSettleDelayMs constructor argument drops from all nine thread test factories. The live title-frame test (a conversation_updated frame after message_saved retitles the thread) stays and passes.

**Docs** — F04's *Deviations* entry becomes a historical note (race measured at 4 s on llamacpp-gemma-4-26b-a4b-it-q8, workaround shipped, server fixed in #210, removal here); F03's R5 status line updates from "carries a bounded client backstop until that is fixed" to the post-fix state.

## Verification — the issue's on-device condition, confirmed on the live stack

Before removing the workaround, #210's behaviour was confirmed end-to-end against the running dashboard with the running ninfer service as the model (the client's role in this path — keeping the stream open and applying the frame — is covered by the live JVM test):

text
conv ("New Conversation") on ninfer-qwen38-27b-nvfp4, send path:
0.01s run_started
1.34s [DONE] + message_saved          <- run terminal in DB
3.52s conversation_updated title=...  <- delivered 2.18s past completion, stream still open

The race window itself (titling slower than the old 3 s backstop) is pinned by #210's backend test test_title_tail_reaches_observer_past_backstop_window, which fails on the pre-fix tree. Probe conversations deleted afterwards (verified 404); no container touched.

JVM suite: testDebugUnitTest → 542 passed, 0 failed, 0 errors, 0 skipped (read from the result XMLs, not counted; the known ThreadToolsTest flake did not fire this run).

## Out of scope

No server-side changes (#210 owns the fix); the conversation_untitled.json fixture stays — the live title-frame test still uses it.

Closes #211


teo-mateo:
READY_FOR_REVIEW ab062143245f6ec48b763ca3f97703679faaf72d

Fix for #211: the F04 title backstop is removed now that #210 holds the stream open through the title tail. ThreadViewModel loses the settle parameter, titleBefore threading, titleFrameSeen tracking, expectTitle gate, awaitAutoTitle() and three constants; the live conversation_updated path is untouched. Six backstop tests removed, the live title-frame test kept. #210's behaviour confirmed on the live stack first (conversation_updated delivered 2.18s past message_saved, stream open). JVM suite 542 passed.

teo-mateo:
VERDICT APPROVED

Reviewed the full diff (12 files, +18/−185, merge base `2fe9e53` → `ab06214`)
and re-verified every factual claim in the body independently.

## Re-verified

- **Removal is exactly the described set.** `ThreadViewModel.kt` loses the
  `titleSettleDelayMs` constructor param + `DEFAULT_TITLE_SETTLE_DELAY_MS`,
  the `titleBefore` argument at all three `collectRun` call sites
  (reattach / send / edit), `titleFrameSeen` tracking in `collectRun` +
  `collectAttempt` and the `RunAttempt.titleFrameSeen` field, the
  `expectTitle` gate, and `awaitAutoTitle()` with `TITLE_SETTLE_ATTEMPTS`
  and the `UNTITLED` sentinel. The sentinel-guard claim checks out against
  the pre-removal code: `expectTitle = !titleFrameSeen && titleBefore ==
  UNTITLED`, so a titled conversation was never polled.
- **The live path is intact.** `is RunEvent.ConversationUpdated ->
  applyTitle(event.title)` survives, and the live title-frame test
  (retitles from a `conversation_updated` frame after `message_saved`)
  stays and uses the `conversation_untitled.json` fixture — the "fixture
  stays" out-of-scope note is accurate.
- **Test surgery is precise.** The six backstop tests go (four
  `ThreadStreamingEdgeTest`, two `ThreadViewModelTest`); the dead
  `serverError` helper and `RUN_STATUS_COMPLETED` constants go with them,
  while `ThreadTerminalPathTest`'s separate live `serverError` usage is
  correctly retained. `titleSettleDelayMs` has zero remaining references
  in the android tree at head (merge base: 14 lines / 10 files — 2
  production, 12 test).
- **No server-side changes** — all 12 changed files are under
  `android/`.
- **Docs are coherent.** F04's deviation is now a historical note
  (measured 4 s race, backstop shipped, server fixed in #210, backstop
  removed here); F03's R5 line moves from "carries a bounded client
  backstop until that is fixed" to the post-fix state.
- **Suite, re-run from the worktree, not taken on faith:**
  `testDebugUnitTest` → **542 passed / 0 failed / 0 errors / 0 skipped**,
  read from the result XMLs; the known `ThreadToolsTest` flake did not
  fire (9/9).
- **Trace arithmetic checks out** (3.52 − 1.34 = 2.18 s), and the #210
  mechanism description matches the merged code (`observe()` holds the
  stream open while the run's worker is still executing).

1. **The on-device probe did not confirm #210 — the body overclaims it.**
   "confirmed #210's behaviour end-to-end against the running dashboard":
   (a) the running dashboard (the sole `app.py` process on :3399, pid
   97488) started 08:56:29 local, ~2 h 34 min **before** #210 merged
   (11:30:36 local), runs `LOG_LEVEL=INFO` ⇒ `debug=False` ⇒ no
   reloader, and has not been restarted since — so it was serving
   pre-#210 `run_manager.py` at probe time. (b) The probe's own signature
   cannot distinguish the fix anyway: pre-#210 `observe()` only closed
   the stream on a 3.0 s `queue.Empty` tick finding a terminal DB state,
   and this probe's 2.18 s title tail sits inside that window, so the
   pre-fix code delivers the frame identically. The pre-fix race
   manifests for tails longer than one heartbeat interval (the 4 s
   measured in the F04 note). The body's own sentence concedes the
   window is "pinned by #210's backend test" — that is the real
   coverage; the probe is a smoke check and reads best described as one.
2. **"all nine thread test factories"** — nine is the *file* count. Only
   `ThreadStreamingEdgeTest` actually has a factory function
   (`private fun viewModel(titleSettleDelayMs: Long = 1L)`); the other
   eight build the ViewModel inline via `ViewModelProvider.create`
   (and `ThreadViewModelTest` has two construction sites). The removal
   set is 12 argument lines across those 9 files and is complete —
   wording only.
3. **F04 doc bullet now cites a historical note.** The second deviation
   bullet — "The idle backstop above emits it on the *send* path too" —
   points at the pre-#210 behaviour this PR and #210 removed. The
   current `observe()` still yields `run_status` from the Empty branch
   (reattach/crash edge), so the defensive wire contract ("one reader
   has to accept it from any of the paths") stands, but the bullet's
   justification describes code that no longer exists. Related trivial
   point: F03's R5 and the F04 note say the backstop "was removed in
   #211" — #211 is the *issue*; the removal is this PR (#214), which
   reads inconsistent next to "fixed in PR #210".

## Conclusion

The removal is exactly scoped, complete (zero dangling references), and
the justification for it — #210's `in_flight` guard — is pinned by its
own backend test, which is the load-bearing coverage here. The suite is
green on an independent run. Approving.

PR #216, opened 2026-09-15
fix(chat): instructive error + namespaced tool hints for bare tool names

### LINKED ISSUE #21: Chat: clearer error + tool-hint hardening when the model emits the bare tool name
## Problem

When the model emits a tool call without the `server_id__` prefix (e.g. just `render-html` or just `render_html`), the chat loop bails out with:

```
Error: Could not determine server for tool 'render-html'
```

See `dashboard/chat/tool_loop.py:70-74` — the `else` branch when `parse_tool_name` returns `server_id=None`. The error string goes back to the model verbatim, leaving it to guess the right format.

Reproduced on conversation `4384d781-f444-43c4-ac9b-4964b3417250` (render-html → 1 failed call, 3 successful). The failure case was the model emitting `render-html` (the server id) as the function name.

## Why the model does it

Two contributing factors:

1. The `tool_hint` text in `dashboard/chat/mcp_registry.py` references the **bare** function name (e.g. *"Use `render_html` when..."*), but the OpenAI-format tool advertised to the model is the **namespaced** name (`render-html__render_html`). The model can anchor on the hint and emit whichever form is shorter.
2. The toggle label in the chat UI is also the server id (`render-html`), giving the model another candidate name to misuse.

## Proposed changes

Two complementary fixes in the same PR. Both surgical, no API change.

### 1. Clearer error for server-id-only calls

In `dashboard/chat/tool_loop.py`, when `server_id is None`, classify the bare string:

- If it matches a known server id (from the `tools` list passed in), the model called the *toggle* by mistake. Reply with a structured message that lists the tools on that server:
  ```
  Error: 'render-html' is a server, not a tool. Call one of: render-html__render_html, render-html__render_html_from_markdown
  ```
- Otherwise, keep the existing message but include the list of available namespaced names so the model can self-correct.

The model already trusts tool-result strings as ground truth; making the error message *instructive* tends to fix retries in one round.

### 2. Tool-hint hardening

In `dashboard/chat/mcp_registry.py` and `dashboard/chat/mcp_config.py`'s normalized hint output:

- Refer to tools by their **full namespaced name** (`render-html__render_html`) in the `tool_hint` text.
- Add a short preamble to every assembled `system` suffix in `get_tool_hints()`:
  ```
  Tools are named `<server_id>__<tool_name>`. Always use the full name when calling.
  ```

This is the contract change that prevents the misuse in the first place. Keep the existing per-server hint bodies but rewrite their tool-name references.

## Out of scope (rejected option)

- **Auto-resolve unprefixed names to a unique matching tool** (the previously sketched cheapest fix). Skipped on purpose: silent recovery hides the model's mistake and trains it that the bare form works, which is worse for long-term prompt hygiene. Explicit errors + correct hints beat magic.

## Files likely touched

- `dashboard/chat/tool_loop.py` — classify-and-explain branch when `server_id is None`.
- `dashboard/chat/mcp_registry.py` — rewrite `tool_hint` strings for the three built-ins to use namespaced names.
- `dashboard/chat/mcp_config.py` — prepend the "Tools are named..." preamble in `get_tool_hints` (or wherever the suffix is assembled).
- `dashboard/mcp_servers.example.json` — update the websearch hint to use `websearch__web_search` for consistency.

## Verification

1. With render-html enabled, ask the model to render a chart. Inspect the tool calls — they should use the namespaced form.
2. Force the failure by manually crafting a conversation where the assistant emits a bare `render-html` tool call. Confirm the error response now lists the namespaced alternatives, and that the model recovers on the next turn.
3. Existing conversations that worked before (correctly namespaced calls) continue to work.

teo-mateo:
## Problem

When the model emits a tool call without the `server_id__` prefix — the **server id** itself (the UI toggle label), or the **bare tool name** (anchored by the old hints) — the loop bailed with `Error: Could not determine server for tool 'render-html'` and left the model to guess the format. Reproduced in conversation `4384d781` (render-html → 1 failed call, 3 successful).

## Changes

**`dashboard/chat/tool_loop.py`** — when `parse_tool_name` finds no server, classify the bare string against the advertised tool list (new `_namespaced_index`, built from the OpenAI dicts the model was actually told it can call):

- **known server id** → `Error: 'render-html' is a server, not a tool. Call one of: render-html__render_html, render-html__render_html_from_markdown`
- **anything else** → the existing message, plus the full available list so the model can self-correct

Tool results are ground truth for the model, so the retry usually fixes itself in one round. The issue's rejected option (auto-resolving unprefixed names) stays rejected: silent recovery would train the model that the bare form works.

**`dashboard/chat/mcp_registry.py`** — hint bodies now reference tools by their full namespaced name (`schemdraw-circuits__draw_circuit`, `project-files__list_files`, `render-html__render_html`, … — every named tool in schemdraw, project-files and render-html; sympy's hint names no tools and is unchanged), and `get_tool_hints` prepends the naming contract —

> Tools are named `<server_id>__<tool_name>`. Always use the full name when calling — the bare server id or the bare tool name is not a callable tool.

— returning `""` when no hint survives, as before.

**`dashboard/mcp_servers.example.json`** — the websearch hint uses `websearch__web_search` for consistency (ragflow's hint names no tool, unchanged).

## Tests

- `test_tool_loop.py`: the two classification branches (bare server id → that server's namespaced list only; unknown bare name → full available list) plus a namespaced-call regression (executed, real result, server id intact) — scripted streams, the exact wire shapes.
- `test_mcp_config.py` `TestToolHints`: the preamble, the empty case, and every namespaced reference in the three built-in hints.

## Verification

- `cd dashboard && venv/bin/python -m pytest tests/ -q` → **1216 passed, 1 skipped** (pre-existing empty-STALE parametrize), exit 0.
- Live probe with the **new code** (venv-driven, not the running dashboard, which predates this change) against the running `ninfer-qwen38-27b-nvfp4` service: real `render-html` tool list + the new system suffix, asked to render a chart. The model issued the fully namespaced `render-html__render_html`, the real MCP server rendered the artifact (`Rendered HTML artifact 'Monthly Signups'.`), and the turn completed with the artifact reference. That covers the issue's verification items 1 and 3 end-to-end; item 2 (forced bare-name failure) is the deterministic scripted coverage above. No container touched; no MCP subprocess left behind.

Closes #21


teo-mateo:
READY_FOR_REVIEW 1f11ec9806fdc49454afcd1d4e17fdaf041f92d0

Fix for #21: bare tool-name calls now get an instructive, classified error (known server id -> that server's namespaced tool list; anything else -> the full available list), and the tool hints + a naming-contract preamble present tools only by their full <server_id>__<tool_name> form. Auto-resolution stays rejected per the issue. New-code live probe against the running ninfer service: the model issued the fully namespaced render-html__render_html and the real MCP server rendered the artifact. pytest 1216 passed.

teo-mateo:
VERDICT APPROVED

Reviewed the full diff (5 files, +142/−8, merge base `26aadb6` → `1f11ec9`)
and re-verified every factual claim in the body independently.

## Re-verified

- **`tool_loop.py` is surgical.** `_namespaced_index()` builds
  `server_id → advertised names` from the exact OpenAI dicts the model was
  told it can call (defensive against missing `function`/`name` — the
  `e2e_progress_test.py` call site passes `{"type": "function"}` and is
  untouched). The `server_id is None` branch now classifies: known server
  id → "is a server, not a tool" + *that server's* namespaced names only;
  anything else → the existing message plus the full available list. The
  happy path is structurally unchanged (verified by
  `test_namespaced_call_still_executes` — executed, real result,
  `server_id` intact). The issue's rejected option (auto-resolve) stays
  rejected: there is no recovery path, only classification.
- **The hint side is complete at the one assembly point.**
  `get_tool_hints()` is the single assembler (one consumer:
  `prompt_builder.build_chat_messages`); the preamble is prepended there
  and `""` when no hint survives (tested). All named tools in the three
  built-in hints are namespaced; sympy's hint names no tools (unchanged,
  as claimed); the websearch example hint now reads
  `websearch__web_search`; the ragflow hint names no tool (unchanged, as
  claimed).
- **Tests are on-point, not decorative.** The two classification branches
  assert on the exact wire messages (server branch excludes other
  servers' tools; unknown branch carries the full list; `mcp.calls == 0`
  in both — nothing executes), and `TestToolHints` pins the preamble, the
  empty case, and every namespaced reference in the built-in hints.
- **Suite, re-run from the worktree, not taken on faith:**
  `pytest tests/ -q` → **1216 passed, 1 skipped, exit 0** — exactly the
  body's claim (the skip is the pre-existing empty-`STALE` parametrize
  from #213).
- **The live probe claim is honest this time.** "venv-driven, not the
  running dashboard, which predates this change" — confirmed: the sole
  `app.py` process on :3399 started 08:56:29 (predates this PR and #214's
  merge), `LOG_LEVEL=INFO` ⇒ no reloader, and the chat DB shows **zero**
  writes/runs/artifacts after 11:55 (the #214 probe deletion) — so the
  probe could not have gone through the running dashboard. I re-ran a
  one-round probe myself against the running `ninfer-qwen38-27b-nvfp4`
  service using the new suffix + the real advertised names: the model
  named the tool fully namespaced (`render-html__render_html`), emitted in
  text form — the harness's `detect_format_drift` handles that shape, and
  the anchoring claim (the one the hint change is meant to produce) holds
  independently. "No MCP subprocess left behind" — confirmed, none
  running. The "real MCP server rendered the artifact" step I did not
  re-execute (no MCP manager in my probe); the load-bearing claim is the
  model-side anchoring, and that one checks out.

1. **The repro conversation `4384d781` is not in the DB** (0 rows in
   `conversations`, 0 in `chat_runs`) — presumably deleted after
   reproducing, which is the right hygiene, but the citation in both the
   issue and the PR body can no longer be followed to a record. The
   scripted coverage stands in for it, so nothing is lost.
2. Trivial: the `else` branch's "Available tools: …" list would be empty
   if `tools` contained no namespaced entries (only the e2e stub does
   this today); the message still degrades to the original wording, so it
   is fine as-is.

## Conclusion

Both fixes are scoped exactly as proposed, the contract line is added at
the single assembly point, the rejected option stays rejected, the suite
matches claim-for-claim, and the probe is honestly described (unlike
#214's — good instinct, and the "the running dashboard predates this
change" note is accurate). Approving.

PR #217, opened 2026-09-15
feat: benchmark UI in the React v2 SPA

### LINKED ISSUE #204: React v2 SPA has no benchmark UI — backend is complete, only the legacy page can reach it

teo-mateo:
## Problem

The benchmarking backend (`dashboard/benchmarking/`) is complete and working,
but the React v2 SPA had no route or components for it — only the legacy
`dashboard/static/benchmark.html` page could reach it.

## Changes

- **`src/hooks/useBenchmark.js`** (new) — the whole benchmark lifecycle in one
  hook: prefill from `GET /benchmarks/service-defaults/<service>` (the endpoint
  merges the service's own params over the bench minimums), param editing with
  a command preview, start (POST), a 2 s poll of the active run (the executor
  blocks on a `docker run --rm` subprocess, so there is no live stream — the
  run's status and timings are what the poll delivers), history list, apply,
  and delete/cancel. A 409 on start refreshes the history so the already
  running run is visible with its live status.
- **`src/components/BenchmarkTab.jsx`** (new) — prefill panel (benchmark-only
  flags badged "bench"), command preview, live run status card (timings with
  stddev, expandable raw output, cancel), and the apply-result banner showing
  applied params and the skipped bench-only flags.
- **`src/components/BenchmarkHistory.jsx`** (new) — expandable run rows
  (params, raw output / error on expand) with apply-to-service and
  delete/cancel actions.
- **`src/components/ServiceDetailsPage.jsx`** — a Benchmark tab, rendered only
  for `template_type === 'llamacpp'` (the only engine the executor supports),
  at `/services/:serviceName/benchmark`.

## Tests

- New `BenchmarkTab.test.jsx` (5 tests): prefill + command preview + history
  render; start → live status → settle on completion (asserts the POST body
  carries the prefill); apply with confirmation and the applied/skipped
  breakdown; a 409 start failure surfacing the error and the existing run;
  and tab gating (present for llamacpp, absent for vllm).
- `npm test`: 507 passed (baseline 502 + 5).
- `npm run lint -- --max-warnings 0` clean; `npm run build` clean.

## Verification

End-to-end probe against the running dashboard (service
`llamacpp-glm-ocr-cpu`, a small CPU model):

- `GET /benchmarks/service-defaults/llamacpp-glm-ocr-cpu` → 200 with the
  prefill (the service's own `-ngl 0`, `--no-mmproj-offload`, `--mmproj` …
  ride in alongside the bench minimums)
- `POST /benchmarks` → 202 `{id, status: pending}`; concurrent POST → 409
  "Benchmark already running"
- poll `GET /benchmarks/<id>`: pending → completed in ~3 s,
  `pp_avg_ts: 8355.8`, `tg_avg_ts: 369.2`, 3 KB `raw_output`
- `PUT /benchmarks/<id>/apply` on a bench-only-flags run → 400
  "No applicable parameters found to apply" (the failure-banner path; the
  apply success path mutates the service and was therefore not exercised
  live — it is covered by the backend's own tests)
- `DELETE /benchmarks/<id>` → run removed, list back to 0

No service was mutated and no container left running (the executor's
`docker run --rm` exits on completion).

## Out of scope

- The legacy `dashboard/static/benchmark.html` page is **untouched** — it
  still works against the same API. Removing it is a separate decision (the
  v1 dashboard is still served at `/`).
- No backend changes: all endpoints already existed and behaved as the issue
  describes.
- Multi-service comparison / trend charts — not in the issue; the history
  table shows past runs, which is the ask.

Closes #204


teo-mateo:
READY_FOR_REVIEW 7470f509700ebd87b2916e865ba043c27a9fee22

Issue #204: benchmark UI for the React v2 SPA.

- Branch: feat/benchmark-ui-v2 (off main @ f524afb)
- 5 new tests; frontend suite 507 passed; lint (max-warnings 0) and build clean
- End-to-end probe against the running dashboard: start → 202, concurrent → 409, poll pending→completed (8355 pp / 369 tg t/s), apply on a bench-only-flags run → 400 (failure-banner path, no mutation), delete → list back to 0. No service mutated, no containers left running.
- Legacy static/benchmark.html untouched (stated in PR body).

Awaiting review.

teo-mateo:
VERDICT APPROVED

Reviewed the full diff (5 frontend files, +869/−3, merge base `f524afb` →
`7470f50`) and re-verified every factual claim in the body independently.
Backend untouched, as claimed — all endpoint shapes were checked against
`benchmarking/routes.py` + `executor.py` directly.

## Re-verified

- **The hook's contract with the backend is exact.** `GET
  /benchmarks/service-defaults/<svc>` returns the bench minimums
  (`-p/-n/-r`) merged over the service's own params (400 for non-llamacpp);
  `POST /benchmarks` → 202 `{id, status: pending}` and 409
  "Benchmark already running for …" on a concurrent start; `GET
  /benchmarks/<id>` carries `pp_avg_ts/pp_stddev_ts/tg_avg_ts/
  tg_stddev_ts/raw_output/error_message/started_at`, which is exactly what
  `RunStatusCard`/`BenchmarkHistory` render; `PUT …/apply` returns
  `{success, message, applied_params, skipped_flags}` (400 when nothing is
  applicable, completed runs only, `serialize_db`-wrapped); `DELETE`
  cancels a pending/running run and deletes a finished one — the hook's
  `cancelRun`/`deleteRun` semantics match. `fetchAPI`'s error extraction
  (`data.error` → `e.message`) is what `startError`/`applyResult` rely on,
  confirmed in `api.js`.
- **The "no live stream" rationale is true of the code:** the executor
  spawns one `docker run --rm … llama-bench` subprocess and the run's
  state lives in SQLite, so the 2 s poll is the only view — and the
  preview `llama-bench -m <model> -o json …` matches
  `_build_docker_cmd`'s arg order (with the same `-m`/`-o` de-dup).
- **Suites, re-run from the worktree, not taken on faith:** `npm test` →
  **507 passed (38 files), exit 0** (body: "507 passed (baseline 502 + 5)"
  — the new file carries exactly the 5 `it()`s described, one per
  claimed behaviour, and the POST-body assertion proves the prefill
  round-trips); `npm run lint -- --max-warnings 0` → exit 0; `npm run
  build` → exit 0.
- **The end-to-end probe re-derived against the running dashboard**
  (`llamacpp-glm-ocr-cpu`, bench-only-flags params so nothing could
  mutate): defaults prefill ✓ (service `-ngl 0` rides in), start → 202
  ✓, concurrent → 409 ✓, poll pending→completed (my run: 79.6k pp / 1.0k
  tg t/s — the body's 8355/369 are its own one-shot numbers, not
  re-derivable, but the flow, statuses, ~3 KB `raw_output` (mine: 3226
  bytes), stddev and `started_at` all match the claim), apply on a
  bench-only run → 400 "No applicable parameters found to apply" ✓,
  delete → history back to baseline ✓, **no bench container left,
  `services.json` mtime unchanged** ✓.
- **Legacy page untouched:** the diff touches only `frontend/src`;
  `static/benchmark.html` is not in it.
- **Tab gating is right:** link and route render only for
  `template_type === 'llamacpp'` (the only engine the executor supports —
  `start_benchmark` 400s on everything else, and
  `get_service_defaults` does too, so a non-llamacpp service can't even
  prefill).

1. **The "bench" badge and the apply skip-list disagree.** The frontend
   `BENCHMARK_ONLY_FLAGS` (11: `-m -p -n -r -o -pg -d -oe -v --delay
   --list-devices`) drives the badge; the backend `validators.py` set (5:
   `-p -n -r -o -m`) drives what `apply` skips. The default path is safe
   (prefill only ever carries the five plus the service's own flags), but
   a user who adds, say, `-d 10` or `--list-devices` via the flag input,
   completes a run, and applies it gets those written into the service's
   `params` — llama-server has no such flags and the container fails at
   startup (the repo's documented unknown-flag failure class). The
   backend set pre-dates this PR (untouched here), so this is not a
   regression — but this PR is what makes the badge and the
   "Skipped (benchmark-only)" banner visible together, and the banner
   only names the backend's five. Follow-up worth an issue: align the
   backend set with llama-bench's actual flag surface, or share one
   source of truth between the two.
2. Trivial: `cancelRun` has no try/catch — a 404 race (run row deleted
   between render and click) is an unhandled rejection; `deleteRun`
   handles the same path. Harmless, but inconsistent.

## Conclusion

The UI does what the issue asks, its whole surface is pinned by tests to
the real backend contract (not to invented shapes), the probe is honestly
scoped (the apply success path is named as untested-live and covered by
the backend's own tests), and both NITs pre-date the change. Approving.

PR #219, opened 2026-09-15
fix(bench): align the apply skip-list with the bench-only flag badge

### LINKED ISSUE #218: bench: align the apply skip-list with the frontend's bench-only flag badge
Follow-up from the #217 review (NIT 1, PR comment on #217).

The benchmark UI badges 11 bench-only flags (frontend BENCHMARK_ONLY_FLAGS in dashboard/frontend/src/hooks/useBenchmark.js: -m -p -n -r -o -pg -d -oe -v --delay --list-devices), but dashboard/benchmarking/validators.py's BENCHMARK_ONLY_FLAGS — what PUT /api/benchmarks/<id>/apply skips — holds only 5 (-p -n -r -o -m).

Consequence: a run that includes a manually-added flag from the 6-flag gap (e.g. -d 10 or --list-devices) will have that flag written into the service's params on apply — llama-server rejects unknown flags at startup (the repo's documented unknown-flag failure class), so the container stops on next start. The default prefill path is safe (the endpoint only ever sends -p/-n/-r plus the service's own flags); reaching the gap requires adding a flag by hand, completing a bench, and applying.

Fix: align the backend set with llama-bench's actual flag surface (or drive both from one source of truth), so the badge, the apply skip-list, and the 'Skipped (benchmark-only)' banner agree. Note the #198 flag-surface guard records engine surfaces in dashboard/flag_surfaces/*.json — llama-bench's flag set could live alongside them.

teo-mateo:
## Problem

Follow-up from the #217 review (NIT 1). The v2 benchmark UI badged 11
bench-only flags, but `apply` skipped only 5 (`-p -n -r -o -m`): a run
carrying a manually-added flag from the 6-flag gap (`-pg -d -oe -v
--delay --list-devices`) got that flag written into the service's params
on apply — `llama-server` rejects unknown flags at startup (the repo's
documented unknown-flag failure class), so the container stops on next
start.

## Changes

- **`flag_metadata.LLAMACPP_BENCH_ONLY_FLAGS`** (new) — the one canonical
  set: the 11 flags the UI badges. `benchmarking/validators.py`'s
  `BENCHMARK_ONLY_FLAGS` is now that same object, so the badge, the apply
  skip-list and the "Skipped (benchmark-only)" banner share a single
  source of truth and cannot drift.
- **`GET /api/benchmarks/bench-only-flags`** (new) — serves the set to the
  frontend; `useBenchmark` now badges from the fetched set instead of a
  hardcoded literal (a fetch failure just leaves the badges unrendered —
  the server's skip-list is what keeps a service safe).
- **`cancelRun`** — catches a 404 race (row deleted between render and
  click) instead of leaving an unhandled rejection; the live card drops
  and the history refresh reconciles (NIT 2 from the same review).

### Why a curated set, not the full llama-bench surface

The issue offered "align the backend set with llama-bench's actual flag
surface (or drive both from one source of truth)" — this takes the second
option deliberately, because the first is wrong in both directions:

- **Overlap flags must stay applicable.** `-ngl`, `-fa`, `-t`, `-sm`,
  `-ts`, ... are in both surfaces. They are valid `llama-server` flags, and
  writing a swept value of them back is exactly what `apply` is for — the
  pre-existing `test_apply_completed_run` pins `-ngl`/`-fa`/`-b` as
  *applied*, and a full-surface skip-list would silently kill the tuning
  workflow.
- **A computed `bench − server` set would still be unsafe.** `-m` and
  `--list-devices` are valid `llama-server` flags (so a set-difference
  would apply them), but a service never carries them as params — and
  `--list-devices` makes the server print devices and exit.
- **What curation *does* buy:** the set is pinned to the recorded
  `llamacpp_bench` surface (a typo or an upstream-removed flag fails the
  guard) and to the curated bench metadata (every badged flag has a panel
  entry the UI can explain).

## Tests

- Guard: `LLAMACPP_BENCH_ONLY_FLAGS` ⊆ recorded `llamacpp_bench` surface
  and ⊆ the bench metadata's cli values (2 new assertions, 1 test).
- `test_bench_routes`: `validators.BENCHMARK_ONLY_FLAGS is
  LLAMACPP_BENCH_ONLY_FLAGS` (the single-source pin); the new endpoint
  serves the sorted set; `test_denylist_flags_never_applied` extended from
  3 to all 11 flags (each skipped, none applied, `-ngl` still applied).
- `BenchmarkTab.test.jsx`: the badge renders from the fetched set (3
  badges for the `-p/-n/-r` prefill, none on the service's own `-ngl`);
  new cancel-404 test (live card drops, history refreshes, no escape).
- Backend: **1219 passed, 1 skipped** (baseline 1216 + 3; the skip is the
  pre-existing empty-`STALE` parametrize). Frontend: **508 passed** (507 +
  1), `lint --max-warnings 0` clean, build clean.

## Verification

The running dashboard pre-dates this change, so the new code was driven
from the venv app factory against the **real** benchmark DB and the real
`services.json` / `docker-compose.yml` (service `llamacpp-glm-ocr-cpu`):

- `GET /api/benchmarks/bench-only-flags` → 200, the 11 flags, sorted
- apply on a real DB run carrying all 11 bench flags + the service's own
  `-ngl 0` / `-fa 1` → 200, `skipped_flags` = all 11 (the 6 gap flags
  included), `applied_params` = `{"-ngl": "0", "-fa": "1"}`
- the write is idempotent (identical values): `sha256` of `services.json`
  and `docker-compose.yml` **identical before and after** — zero effective
  mutation; probe run deleted, run table back to 0 for the service
- the running dashboard (untouched) answers 200 on `/api/health`
- the motivation for the fix was confirmed against the image directly:
  `llama-bench ... --temp 0.1` → `error: invalid parameter for argument:
  --temp` (timeout-wrapped `docker run --rm`; no container left running)

## Out of scope

- The legacy `static/benchmark.html` page keeps its own 11-flag literal —
  it now equals the backend set exactly (it pre-dated the v2 page and this
  change aligns to it), so badge and skip-list agree there too; retiring
  that page is a separate decision.
- No change to how the executor passes params to `llama-bench` (a
  full-prefill run still rejects service-only flags like `--temp` at the
  bench itself — pre-existing, both pages share it, and removing
  non-bench flags from the prefill is a behaviour change beyond this fix).

Closes #218


teo-mateo:
READY_FOR_REVIEW 587fe5d43a42d2cb86e4699e87df3b636687aa31

Issue #218: align the apply skip-list with the bench-only flag badge.

- Branch: fix/bench-apply-skiplist (off main @ 6ed11ab)
- One canonical set (flag_metadata.LLAMACPP_BENCH_ONLY_FLAGS, the 11 the UI badges) now drives apply's skip-list and a new /api/benchmarks/bench-only-flags endpoint the frontend badges from; pinned to the recorded llama-bench surface and the bench metadata by the guard.
- Curated set, not the full bench surface: overlap flags (-ngl, -fa, ...) must stay applicable (existing test pins it), and a bench-minus-server set would still apply --list-devices. Rationale in the PR body.
- cancelRun 404-race catch (review NIT 2) included.
- Backend 1219 passed (1216 + 3), frontend 508 (507 + 1), lint --max-warnings 0 and build clean.
- New code driven from venv against the real DB/services.json/compose: endpoint 200 with 11 flags, apply on a real run skipped all 11 (gap flags included), sha256 of services.json + compose identical before/after (idempotent write), probe run deleted. llama-bench's rejection of service-only flags confirmed against the image (timeout-wrapped --rm).

Awaiting review.

teo-mateo:
VERDICT APPROVED

Reviewed the full diff (8 files, +139/−16, merge base `6ed11ab` →
`587fe5d`) and re-verified every factual claim in the body independently.
This closes #218 — the NIT I raised on #217 — and also takes NIT 2
(`cancelRun` 404 race).

## Re-verified

- **Single source of truth, pinned by identity.**
  `flag_metadata.LLAMACPP_BENCH_ONLY_FLAGS` (the 11 flags) is the
  canonical set; `benchmarking/validators.py` now binds
  `BENCHMARK_ONLY_FLAGS` to that *same object* (`=`, not a copy) and
  `test_skip_list_is_the_canonical_set` asserts `is` — so the badge, the
  apply skip-list and the "Skipped (benchmark-only)" banner cannot
  drift back apart silently. `GET /api/benchmarks/bench-only-flags`
  serves `sorted(set)` behind `@require_auth`; the frontend hook fetches
  it and badges from the fetched set, with a fetch failure degrading to
  "no badges" only (the server's skip-list remains the safety
  mechanism) — exactly the claim.
- **The curation argument is correct, not just asserted.** I checked the
  real binary: `llama-bench --help` in the image lists all 11 (`-pg`,
  `-d`, `-oe`, `--delay`, `--list-devices`, …). Overlap flags like
  `-ngl`/`-fa`/`-t` are valid `llama-server` flags that `apply` must
  still write back (the extended
  `test_denylist_flags_never_applied` asserts all 11 skipped **and**
  `-ngl` applied), and `-m`/`--list-devices` are valid server flags a
  service never carries as params — so "full surface" and "bench −
  server" are both wrong in the directions named. The guard test pins
  the set to the recorded `llamacpp_bench` surface (a typo or
  upstream-removed flag fails loudly) **and** to the bench metadata
  (every badged flag has a panel entry); I confirmed all 11 are in
  `flag_surfaces/llamacpp_bench.json`.
- **Suites, re-run from the worktree, not taken on faith:** backend
  `pytest tests/ -q` → **1219 passed, 1 skipped, exit 0** (body: 1216 +
  3, pre-existing empty-`STALE` skip — exact match); frontend `npm
  test` → **508 passed (38 files), exit 0** (507 + 1 — the new
  cancel-404 test), `lint --max-warnings 0` exit 0, build exit 0.
- **The venv-factory verification re-derived against the real
  main-checkout files** (the worktree has no compose file; the factory
  was pointed at `/github/teo-mateo/llm-dock/…`'s `benchmarks.db` +
  `services.json` + `docker-compose.yml`, service
  `llamacpp-glm-ocr-cpu`): endpoint → 200 with the 11 flags sorted ✓;
  apply on a real-DB completed run carrying all 11 + the service's own
  `-ngl 0`/`-fa 1` → 200, `skipped_flags` = all 11,
  `applied_params` = `{"-ngl": "0", "-fa": "1"}` ✓ (idempotent values,
  so the `sha256` of `services.json` and `docker-compose.yml` were
  **identical before and after** ✓); probe run deleted, run table back
  to 0 for the service ✓.
- **The motivation reproduces:** `llama-bench … --temp 0.1` in the
  image → `error: invalid parameter for argument: --temp` — the
  exact claimed output, no container left.
- **Out-of-scope claims hold:** the legacy `static/benchmark.html`
  keeps its own literal (line 210 — verified to be the identical 11
  flags, so badge and skip-list agree there too), the executor is
  untouched (not in the diff), and the running dashboard (which
  pre-dates the change, as honestly stated) answers 200 on
  `/api/health`.
- **NIT 2 is properly closed:** `cancelRun` now try/catch/finally —
  a 404 race drops the live card and refreshes history instead of
  leaving an unhandled rejection, pinned by the new test (card gone,
  start re-enabled, history refetch fired, rejection did not escape).

## NITs

None. The body is unusually thorough and every claim I probed held —
including the honest "running dashboard pre-dates this change, so the
new code was driven from the venv app factory" framing. One neutral
observation for the record, not a defect: `bench-only-flags` is a
static segment under the same prefix as `<run_id>`; Flask's static-over
converter precedence makes it safe (my probe hit the flags endpoint,
not a run lookup), so nothing to do.

## Conclusion

The fix is the right shape (one curated canonical set, identity-bound,
served to the UI, guard-pinned to the recorded surface), the
motivation is reproduced against the image, the suites match
claim-for-claim, and the apply path was re-derived end-to-end with zero
effective mutation. Approving.

PR #222, opened 2026-09-15
feat(ninfer): metrics endpoint and Live Metrics panel

### LINKED ISSUE #195: [PLAN] Add ninfer metrics

teo-mateo:
## Problem

`template_type: "ninfer"` services returned `{"metrics": {}}` from
`/api/services/<name>/metrics` and the Live Metrics panel was disabled for
them — even though the data exists in-process in NInfer's `RuntimeStats` /
`MemorySummary` snapshots (the engine's own 5 s reporter reads exactly
those). Per the spec at `docs/plans/ninfer-metrics.md` (R1–R6), this lands
all four steps in one PR.

## Changes

**Step 1 — engine patch** (`ninfer/ninfer-metrics.patch`, new; two Dockerfile lines):
- Authed `GET /metrics` next to `/health`. The pre-routing guard already
  protects every path but `/health` and OPTIONS, so **no auth-handler
  change** (R6); 503 mirrors `/health` until the service is attached.
- The handler renders the 12 unlabeled Prometheus families from
  `RuntimeStats` + `MemorySummary` via an anonymous-namespace formatter —
  same access pattern as the engine's own reporter. No CMake change; two
  files (`http_server.{h,cpp}`), disjoint from the port patch.
- Dockerfile applies it after the port patch; the check-upstream skill
  tracks its **separate retirement condition** (dies when upstream ships a
  stats endpoint of its own, unlike the port patch).

**Step 2 — dashboard** (`routes/metrics.py`, `tests/test_metrics.py`):
- `NINFER_CURATED_METRICS` (the 12 names); gate and curated-set selection
  become engine-keyed lookups; the vLLM-only `per_pos` flattening stays
  vLLM-only. Spec-acceptance and preemption families are **deliberately
  absent** — NInfer keeps no service-level aggregate for either (§1
  non-goal) — so those cells render "—" (R3). `/slots` unchanged: ninfer
  hits the existing 400 (R4).

**Step 3 — frontend** (`useServiceMetrics.js`, `ServiceDetailsPage.jsx`):
- `NINFER_TO_VLLM` rename table (7 panel-facing families; the raw extras
  pass through untouched); `normalizeMetrics` is now table-driven for
  both engines, so the hook's `vllm:*` reads, counter-delta rates, and the
  `/slots` skip are unchanged. The panel's `enabled` gate gains `ninfer`.
  Legacy/unknown engines pass through untouched.

**Step 4 — docs**: engine guide NInfer section (metrics out of "Not
supported", family set, authed-endpoint decision, the G2 caveat),
CHANGELOG entry.

## Tests

- `TestNinferMetrics` (8 new): curated happy path with the full 12-family
  sample + off-whitelist filter, connection error → `{}`, the service API
  key on the scrape, 401/503 upstream collapse to `{}`, dashboard 401,
  slots → 400, and the unsupported-engine gate still returns empty with
  the raw engine name.
- `useServiceMetrics.test.js` (4 new): rename (mapped names gone, raw
  extras kept); the derived datapoint — rates from counter deltas,
  `kvCache`, `prefixHitRatio`, `running`/`waiting`, `specAcceptRatio`
  **undefined** (no service-level aggregate → the donut renders "—", R3);
  **no `/slots` fetch for `engine: "ninfer"`**; unknown-engine passthrough.
- R4 held by the untouched vLLM/llama.cpp suites: `pytest tests/` →
  **1242 passed, 1 skipped** (the skip is the pre-existing empty-`STALE`
  parametrize), `pytest tests/test_metrics.py -q` → 27 passed. Frontend:
  **512 passed** (508 + 4), `npm run build` clean, lint
  `--max-warnings 0` clean.
- R5 apply check: both patches, **in sequence, on a fresh clone of the
  pin** `d4929686`: `git apply --check ninfer-cu129-port.patch` OK, then
  `git apply --check ninfer-metrics.patch` OK (the patch is a plain
  `git diff` of that tree, generated from it).

## Live gates — UNVERIFIED, pending operator rebuild

Per the PO directive, `./build-ninfer.sh` was **not** run, no container
was recreated or stopped (`ninfer-qwen38-27b-nvfp4` is untouched — Up 5h
through this work). Therefore:

- **G1 (build + probe): UNVERIFIED.** The unauth-401 / authed-200-with-12-
  families / 503-while-unavailable behavior is asserted by the patch's
  design (pre-routing guard + `/health` mirror) and unit-test-mocked on
  the dashboard side, but has not been observed on the rebuilt image.
- **G2 (KV gauge units): UNVERIFIED — and this is the one contract the
  plan does not prove from source.** `device_main_kv_occupied_pages` is a
  target-level physical page count while `kv_capacity_page_groups` is the
  capacity resolver's unit; the ratio has no in-tree precedent. The patch
  clamps to [0, 1] (0 on unknown capacity) and emits the **raw pair
  alongside** (`ninfer:kv_occupied_pages`, `ninfer:kv_capacity_page_groups`),
  both curated, so a wrong scale is a visible fix, not a silent one.
  **R2 is therefore not declared done on the Active KV donut until G2
  passes.**
- **G3 (running-request semantics): UNVERIFIED.** The patch maps
  `running_requests` as-is and also emits `ninfer:num_requests_prefilling`
  so the fallback (map `running + prefilling`) is a one-line patch edit
  if the panel reads low under 2 concurrent requests.

Nothing was smoothed over: with G1–G3 open, the engine-side contract is
authored and apply-checked but not observed live.

## Out of scope (plan §1 non-goals)

- Spec-acceptance metrics (no service-level aggregate exists in
  `RuntimeStats`; the gauge stays "—"; follow-up recorded in the plan).
- `/slots` parity, preemption counter, reasoning levels, Open WebUI,
  benchmarking — untouched.

Closes #195


teo-mateo:
READY_FOR_REVIEW b594def0290fcb96f889065709948a6545d99507

Issue #195: NInfer metrics endpoint + Live Metrics panel (steps 1-4 of docs/plans/ninfer-metrics.md).

- Branch: feat/ninfer-metrics (off main @ 8c39120)
- Engine: ninfer/ninfer-metrics.patch (authed GET /metrics, 12 Prometheus families from RuntimeStats + MemorySummary, no CMake change, no auth-handler change — the pre-routing guard already covers it). Both patches apply --check in sequence on a fresh clone of pin d4929686.
- Dashboard: NINFER_CURATED_METRICS + engine-keyed gate/whitelist; /slots stays llama.cpp-only; spec-acceptance/preemption deliberately absent (no service-level aggregate) so those cells render '—'.
- Frontend: NINFER_TO_VLLM rename table, table-driven normalizeMetrics, panel gate gains ninfer; no /slots fetch for ninfer.
- Suites: backend 1242 passed (1234 + 8), test_metrics 27, frontend 512 (508 + 4), build + lint --max-warnings 0 clean.
- Live gates G1/G2/G3 are UNVERIFIED pending operator rebuild, stated as such in the PR body — G2 (KV gauge units) named explicitly as the one contract the plan does not prove from source; R2 not declared done on the Active KV donut until G2 passes. No build run, no container touched (ninfer-qwen38-27b-nvfp4 Up 5h, untouched).

Awaiting review.

teo-mateo:
VERDICT APPROVED

Reviewed the full diff (10 files, +479/−13, merge base `8c39120` →
`b594def`) and re-verified every factual claim in the body independently.
Per the operator directive, the live gates G1/G2/G3 are treated as
UNVERIFIED-by-design (not a blocker), and no `./build-ninfer.sh` was run
nor any container recreated — `ninfer-qwen38-27b-nvfp4` is untouched
(Up 5h, as stated). Everything that needs no GPU was re-derived.

## Re-verified (all no-GPU)

- **R5 patch discipline, re-derived on a fresh clone of the pin
  `d4929686`:** `git apply --check ninfer-cu129-port.patch` OK, then
  `git apply --check ninfer-metrics.patch` OK *on the port-applied tree*
  (matching the Dockerfile's apply order). I also applied the port patch
  for real and confirmed the metrics patch lands clean on the actual
  result, not just `--check`.
- **The patch is honest against the pinned source.** Every field it reads
  exists in the pin's `include/ninfer/types.h`: `RuntimeStats`
  (`computed_prefill_tokens`, `committed_decode_tokens`, `decode_rounds`,
  `running/prefilling/waiting_requests`, `reused_prompt_tokens`,
  `device_main_kv_occupied_pages` `uint32`, `host_kv_occupied_bytes`
  `size_t`) and `MemorySummary.kv_capacity_page_groups` (`uint32`). Both
  KV operands are integers and the patch casts to `double` *before*
  dividing — no integer truncation. The `runtime_stats()` /
  `memory_summary()` accessors exist on `GenerationService` and return
  snapshots.
- **R6 auth, verified against the pin's `http_server.cpp`:** the
  pre-routing handler passes unauthenticated **only** for `/health` and
  OPTIONS, so `/metrics` is guarded and no auth-handler change is needed
  (as claimed). The dashboard scrape already sends the service key —
  `test_engine_api_key_header_passed` asserts `Bearer test-key`.
- **The 503 mirror is faithful:** same predicate as `/health`
  (`service_ != nullptr && is_available()`), same JSON body.
- **G2's "visible, not silent" design holds end-to-end** (the specific
  thing to confirm): the raw pair `ninfer:kv_occupied_pages` /
  `ninfer:kv_capacity_page_groups` is emitted by the patch, curated in
  `NINFER_CURATED_METRICS` (so it reaches the payload), and passes
  through the frontend rename untouched (the test asserts the raw extras
  survive while the mapped names are gone). A wrong scale is a visible
  fix, not a silent one.
- **G3 fallback present:** `ninfer:num_requests_prefilling` is emitted,
  making the `running + prefilling` remap a one-line patch edit.
- **The 12-family set matches the plan spec one-for-one** (names, source
  fields, types, the raw-pair design, and the G2-caveat wording in
  `docs/plans/ninfer-metrics.md`).
- **Suites, re-run from the worktree, not taken on faith:**
  `pytest tests/ -q` → **1242 passed, 1 skipped, exit 0** (body: 1242 +
  the pre-existing empty-`STALE` skip — exact match);
  `pytest tests/test_metrics.py -q` → **27 passed**; `npm test` →
  **512 passed (38 files), exit 0** (508 + 4 — exact match);
  `lint --max-warnings 0` → exit 0; `npm run build` → exit 0.
- **Docs step edited the right file:** `AGENTS.md` is a tracked symlink
  (mode 120000) to `CLAUDE.md`, so the `CLAUDE.md` edits surface in the
  live guide. `test_unsupported_engine_still_returns_empty` confirms the
  raw engine name still rides through for engines without a curated set.

## NIT (no action required before merge)

1. **G2's "not proven from source" is slightly overstated — in a good
   way.** Tracing the two operands in the pinned tree: the numerator
   `device_main_kv_occupied_pages` is a device main KV **page** count
   (`PhysicalUsageSnapshot.device_main_kv_pages`), and the denominator
   `kv_capacity_page_groups` is the capacity resolver's `main_page_groups`
   = `1 + (explicit_tokens − 1) / main_page_tokens` — also a main KV
   **page** count (the resolver's own error in `kv_capacity.cpp`
   `reservation_bytes` prints "Main KV **page count** is outside the
   target capacity curve" about it). So the ratio is dimensionally
   pages÷pages → an occupancy fraction in [0, 1]. What is *not* provable
   from source is the **empirical** occupancy under load (what G2's live
   probe measures), not the unit dimension — and the raw pair is emitted
   either way, so the conservative stance is safe. This sharpens what
   exactly remains open when the operator runs G2; nothing changes.

## Conclusion

The engine-side contract is authored, apply-checked on a fresh clone of
the pin, and faithful to the pinned source; the dashboard and frontend
wires are pinned by tests to the real engine family set; the suites match
claim-for-claim; and the one unproven contract (G2) is handled exactly
as the plan and the operator directive require — raw pair emitted so a
wrong scale is a visible fix. Approving.

PR #223, opened 2026-09-15
feat(android): select the thread's prompt by reference, not content

### LINKED ISSUE #221: android: switch prompt selection from content-matching to prompt_id
Follow-up from #123 (merged as PR #220, 8c39120) — the issue's own 'follow-up once this lands'.

The conversations API now returns `prompt_id` alongside `main_system_prompt` (additive; the client keeps working until it switches). Remove the content-matching in the Android client and drive prompt selection off the reference:

- ChatSettingsSheet: activePromptContent — match the conversation's prompt by id, not by string-comparing every prompt's content
- ThreadViewModel.selectPrompt — send {prompt_id} on selection (the server resolves the content); sending {prompt_id: null, main_system_prompt: ''} detaches and clears, matching the React picker (ChatArea.handlePromptSelect)
- PromptDto.content — drop the field if nothing else needs it

Behaviour note carried over from #220: a legacy conversation whose text only matches a prompt's content now shows None (the content-match guess was the bug); reselecting the prompt establishes the reference. The React picker already implements this contract — ChatArea.test.jsx is the reference for the expected wire shapes.

teo-mateo:
## Problem

The Android client still selected a thread's system prompt by
**content-matching** — string-comparing the conversation's stored text
against every managed prompt's content. That was the bug #123 removed
from the React picker: a pre-feature conversation whose text only
*matches* a prompt's content was treated as "on" that prompt. The
conversations API now returns `prompt_id` alongside
`main_system_prompt` (additive; merged in #220), so the phone can drive
the same reference-based contract.

## Changes

- **Reference parsed end to end** — `ConversationDetailDto` /
  `ConversationDetail` + the mapper gain `prompt_id`. The selection's
  identity is the stored reference; `main_system_prompt` stays the
  resolved copy the run path reads.
- **`ConversationsRepository.setMainSystemPrompt` (text PUT) →
  `setPrompt`** — selection PUTs `{"prompt_id": "<id>"}` alone (the
  server resolves the content and stores its copy); detach PUTs
  `{"prompt_id": null, "main_system_prompt": ""}` — the React picker's
  exact shape (`ChatArea.handlePromptSelect`). The body is hand-built
  for the same reason the reasoning-level clear is: with
  `explicitNulls = false` a null DTO field is *omitted*, and a detach
  of `{}` answers 200 while changing nothing.
- **`ThreadViewModel.selectPrompt(promptId: String?)`** — drives off
  the reference: same-id is a no-op, optimistic update, revert on
  failure. The old short-circuit (`storedText == prompt.content` →
  skip) is gone.
- **`ChatSettingsSheet`** — rows match by `id`, an explicit **None**
  row renders first (selected when the reference is null; tapping it
  detaches + clears to the default), and the "prompt that isn't in the
  list" footnote is replaced by that row. `PromptRow` generalises to
  `(id, name, …)`; test tags stay `prompt_option_<id>` with
  `prompt_option_none` for the new row.
- **`PromptDto.content` / `ManagedPrompt.content` dropped** — nothing
  else on the phone reads a prompt's text (selection sends the id, the
  sheet matches by reference).

## Behaviour note (carried from #220)

A legacy conversation whose text only matches a prompt's content now
shows **None** — the content-match guess was the bug. Reselecting the
prompt establishes the reference (the stored text is replaced by the
prompt's current content, since the server resolves it).

## Tests

- `ConversationsRepositoryTest` (2 new): select PUTs the id as the
  *whole* body (`main_system_prompt` absent); detach PUTs the exact
  `{"prompt_id":null,"main_system_prompt":""}` — the explicit `JsonNull`
  is the load-bearing assertion.
- `ThreadPromptSelectionTest` (5 new, new fixture
  `conversation_with_prompt.json`): optimistic select + id-only wire;
  the legacy conversation whose text equals a prompt's content is still
  unreferenced and the PUT still goes out; detach clears the reference
  and sends the explicit-null body; reselecting the current reference
  sends nothing (load's own GET drained first, then no request in
  200 ms); a 404 reverts the reference and surfaces the server's
  error.
- Full JVM suite: **549 passed** (542 + 7). No Android emulator
  involved; the sheet's Compose test (`ThreadToolsTest`) and the
  new-chat flow (already id-driven, untouched) are green in the same
  run.

## Out of scope

- Free-text prompt authoring on the phone (stays dashboard-only, F03
  "Out of scope" — picking None detaches to the default, it does not
  open an editor).
- The new-chat sheet — already id-driven (`prompt_option_default`);
  no change.

Closes #221


teo-mateo:
READY_FOR_REVIEW ffac64cc42d5a27661e59cdb759264ecf18e8a96

Issue #221: Android prompt selection by reference (follow-up to #123 / PR #220).

- Branch: feat/android-prompt-id (off main @ bc35e08)
- ConversationDetailDto/ConversationDetail + mapper gain prompt_id; selection's identity is the stored reference.
- ConversationsRepository.setMainSystemPrompt (text PUT) → setPrompt: select PUTs {prompt_id} alone; detach PUTs {prompt_id: null, main_system_prompt: ""} (React picker's exact shape, hand-built body — explicitNulls trap, same as the reasoning-level clear).
- ThreadViewModel.selectPrompt(promptId: String?) — no-op on same id, optimistic, revert on failure; the storedText == prompt.content short-circuit is gone.
- ChatSettingsSheet: id-matched rows, explicit None row (detach + clear to default), footnote replaced; PromptRow generalises to (id, name, …), test tags prompt_option_<id> / prompt_option_none.
- PromptDto.content + ManagedPrompt.content dropped (nothing else reads a prompt's text on the phone).
- Behaviour change stated in the body: a legacy conversation whose text only matches a prompt now shows None; reselecting establishes the reference.
- Tests: 2 wire-shape tests (id-only select, explicit-null detach) + 5 ThreadPromptSelectionTest cases (optimistic select, legacy content-match case, detach, reselect no-op, 404 revert). Full JVM suite 549 passed (542 + 7).

Awaiting review.

teo-mateo:
VERDICT APPROVED

Reviewed the full diff (14 Android files, +387/−58, merge base
`bc35e08` → `ffac64c`) and re-verified every factual claim in the body
independently. This closes #221 — the Android follow-up I filed after
#220, whose contract (wire shapes, `ChatArea.test.jsx` as the reference)
this PR implements exactly.

## Re-verified

- **The switch is complete and clean.** `prompt_id` is parsed end to
  end (`ConversationDetailDto` → `ConversationDetail` → mapper); the
  selection's identity is the stored reference. The old
  content-matching short-circuit is gone from `selectPrompt` (the
  `storedText == prompt.content` skip is replaced by
  `promptId == promptId` no-op), and **nothing else on the phone reads
  a prompt's text** — `PromptDto.content` and `ManagedPrompt.content`
  are dropped, `UpdateSystemPromptRequestDto` is deleted, and a repo-wide
  search finds zero dangling references to any of the three.
- **The wire contract matches the merged server exactly.** I checked
  both bodies against #220's `update_conversation` route: selection's
  `{"prompt_id": "<id>"}` alone → the server resolves and stores the
  content (reference kept); detach's `{"prompt_id": null,
  "main_system_prompt": ""}` → the resolve block is skipped (null),
  `prompt_id` stores NULL and the text clears — the React picker's
  `ChatArea.handlePromptSelect` shape byte for byte. The hand-built
  body is the right call: the `explicitNulls = false` trap is documented
  against the reasoning-level clear precedent, and the detach test pins
  the explicit `JsonNull` + empty string as the load-bearing assertion
  (a `{}` detach would 200 while changing nothing).
- **The sheet matches the body:** explicit **None** row rendered first
  (selected when the reference is null; tap = detach + clear to
  default), rows match by `id`, the "prompt that isn't in the list"
  footnote is replaced by that row, `PromptRow` generalises to
  `(id, name, …)` and the tag generalises to `prompt_option_none`.
- **Tests are on-point, one per claimed behaviour.** 2 repository tests
  pin the exact PUT bodies (id-only with `main_system_prompt` absent;
  the explicit-null detach body). 5 view-model tests cover optimistic
  select + id-only wire, **the legacy conversation whose text equals a
  prompt's content is still unreferenced and the PUT still goes out**
  (the behaviour change this switch exists for), detach with the
  explicit-null body, reselect-sends-nothing (load's GET drained first,
  then nothing in 200 ms), and 404 reverts the reference with the
  server's message. The new fixture's `id` matches the test constant.
- **JVM suite, re-run from the worktree, read from the result XMLs (not
  counted):** **549 tests, 0 failures, 0 errors, 0 skipped** — exact
  match for the body's "549 passed (542 + 7)". The XMLs are fresh and
  the 5 new `ThreadPromptSelectionTest` cases all ran and passed.
- **Out of scope holds:** the new-chat sheet is id-driven
  (`prompt_option_default` present in `NewChatScreen.kt`, not in this
  diff) and free-text authoring stays dashboard-only — no editor
  surface appears.

## NITs

None. The body is accurate and the one subtlety that usually bites this
contract (the kotlinx-serialization null-omission) is both handled and
pinned by the load-bearing test.

## Conclusion

The client now drives prompt selection off the reference exactly as
#221 prescribed, the wire shapes match the merged server routes and the
React picker, the removed field has zero dangling consumers, and the
suite matches claim-for-claim. Approving.

Issue #88, opened 2026-07-17
Chat/project file editor: browser Back/Forward bypasses the dirty-discard guard


teo-mateo:
Follow-up from PR #87 codex iteration 2, finding 2.2 (also applies to the pre-existing ProjectPage editor from PR #79).

The dirty-editor discard guard runs only in ChatPage's explicit navigation handlers plus `beforeunload`. A client-side POP navigation (browser Back/Forward) changes the route without either firing, so a dirty `ProjectFileEditor` (project page or the PR #87 chat-split overlay) unmounts silently.

A proper fix needs a route-level blocker (`useBlocker`), which requires migrating from `<BrowserRouter>` to `createBrowserRouter`/`RouterProvider` — a cross-cutting router change that should be its own PR.

teo-mateo:
Duplicate of #80 — same root cause (no route-level blocker), same required fix (migrate `<BrowserRouter>` → `createBrowserRouter`/`RouterProvider`, then `useBlocker`). Consolidating so one PR does the router migration once.

Folded into #80: this issue adds that the gap also covers the **chat-split `ProjectFileEditor` overlay** from PR #87, not just the ProjectPage editor from PR #79. Whoever takes #80 must cover both mount points.

Closing as duplicate during PO backlog triage; #80 is the live one.

PR #225, opened 2026-09-15
feat(frontend): data-router migration + dirty-editor Back/Forward guard (#80)

### LINKED ISSUE #80: Editor dirty-state guard: cover browser Back/Forward (needs data-router migration for useBlocker)
Follow-up from PR #79 codex iteration 7.

The project file editor guards unsaved changes on all explicit exits (sidebar navigation, new-conversation/new-file actions, editor close, browser unload via beforeunload). Same-document history transitions (browser Back/Forward) bypass these: React Router unmounts the editor without a prompt.

A clean fix needs `useBlocker`, which requires migrating the app from `<BrowserRouter>` + `<Routes>` to a data router (`createBrowserRouter`) — a structural change across all routes that was judged out of scope for PR #79. Alternative (popstate sentinel re-push) is fragile and can desync router state.

Scope when picking this up:
- Migrate to a data router
- Add a dirty-editor navigation blocker (`useBlocker`) fed by the existing onDirtyChange chain
- MemoryRouter regression: two history entries, dirty editor, navigate(-1), cancel preserves route + text

teo-mateo:
## Problem

Issue #80: the project file editor guards unsaved changes on every *explicit*
exit (sidebar navigation, new-conversation/new-file actions, editor close,
browser unload via `beforeunload`) — inherited from PR #79 and extended by
PR #87 to the chat-split overlay. Same-document history transitions
(browser **Back/Forward**) bypass all of them: the router unmounts the
editor without a prompt. The clean fix needs `useBlocker`, which only works
under a **data router** — hence the `BrowserRouter` + `<Routes>` →
`createBrowserRouter` migration. #88 (chat-split editor) was closed as a
duplicate into #80; both mount points are covered here.

## Changes

**Migration (mechanical — the only reason any route file is touched):**
- `App.jsx` — builds `createBrowserRouter(routes, { basename: "/v2" })` and
  renders `RouterProvider`. The `basename` is what the legacy
  `BrowserRouter` carried; nothing else about routing moves.
- `routes.jsx` (new) — the route table, verbatim paths, now as a `'/'`
  **layout route**: `AppShell` wraps an `Outlet`, so the shell (Sidebar,
  MobileNav) stays mounted across route changes — exactly the legacy shape
  where the shell wrapped `<Routes>` and only the matched route swapped.
  Child paths are relative to `'/'`, preserving the old matching and
  ranking (static `ghost` still beats the param route,
  `chat/project/:projectId` still beats `chat/:conversationId?`).
- `layout.jsx` (new) — `AppShell` and `DefaultLayout`, moved verbatim out
  of `App.jsx` (they live there only so `routes.jsx` can stay a pure
  route table; `react-refresh/only-export-components` forbids mixing a
  non-component export with component definitions in one file).
- `main.jsx` — drops the now-unused `BrowserRouter` wrapper (2 lines).

**The guard (`ChatPage.jsx`, ~35 lines):**
- `useBlocker` fed by the **existing** `editorDirtyRef` chain. Both mount
  points are covered by the same ref: the `ProjectPage` editor and the
  chat-split `ProjectFileEditor` overlay both report dirty state through
  `onEditorDirtyChange` → `editorDirtyRef` (PR #79 and #87 wired that up;
  this PR only adds the POP intercept on top).
- **Armed for `POP` only** (`historyAction === NavigationType.Pop`): every
  PUSH/REPLACE entry point already runs `confirmDiscardEdits()` before
  `navigate()`, so blocking all navigation types would prompt twice
  (handler + blocker). POP is the one class the explicit guards never see.
- The blocker renders a Stay/Leave modal. **Leave re-confirms through the
  same `window.confirm('Discard unsaved changes?')`** the explicit guards
  use (the blocker's `proceed()` has no prompt of its own); Stay calls
  `blocker.reset()`.

**No behaviour changes** beyond the new Back/Forward prompt (pinned, not
just claimed):
- Unknown paths still render the shell with an empty main — the layout
  route matches and the unmatched `Outlet` renders nothing, same as the
  old `<Routes>` no-match. No 404 boundary was added.
- The route table's matching is pinned by `routes.test.jsx` (every path →
  intended element, wildcard params, unknown → shell-only).
- `beforeunload` (browser close/reload) is untouched.

**Docs:** frontend `AGENTS.md` — routing entry, route table (including the
`/chat/ghost` row that was missing before this PR touched the table), the
`ChatPage` guard, and the data-router gotchas.

## Tests

- **`ChatPage.test.jsx` (18 existing → 18, mechanically converted)** — the
  page now renders under `createMemoryRouter` (a data router is mandatory:
  `useBlocker` throws under the legacy `MemoryRouter`). Same routes, same
  assertions. One non-mechanical detail, verified empirically: a data
  router does **not** propagate parent rerenders into the route tree (the
  legacy `MemoryRouter` did — a stable-router conversion made the URL-load
  harness vacuous: `probe 1 → 4` legacy vs `1 → 1` data; a same-location
  `router.navigate()` push is a no-op too). The harness therefore delivers
  its tick through a context consumed by the route element
  (`TickContext`/`ChatPageHost`), which re-renders `ChatPage` without
  remounting — restoring the test's original power.
- **New Back/Forward regression (the one the issue names)** —
  `createMemoryRouter` with **two history entries**, dirty editor,
  `navigate(-1)`: **cancel preserves the route AND the editor text**.
  Plus two siblings: *Leave* commits the Back navigation and unmounts the
  editor (the commit is async — `proceed()` resolves through a microtask,
  so the assertion flushes `act` first), and a **clean editor lets Back
  pass through without a prompt** (the blocker must arm only on dirty).
- **`routes.test.jsx` (7 new)** — pins the production route table's
  matching without rendering (initial matches are synchronous): `/` →
  dashboard inside `DefaultLayout`, `/chat/ghost` → `GhostChatPage`,
  `/chat` + `/chat/:id` + `/chat/project/:id` → `ChatPage`,
  `/tools`/`/settings` → `DefaultLayout`-wrapped pages,
  `/services/:name/*` wildcard params, unknown → shell-only.
- **`ChatPageDefaultModel.test.jsx` (4 → 4, converted)** — same
  mechanical data-router conversion; its forced rerender (default model
  changes) also moves to the tick-context mechanism, since the stable
  router won't propagate it either.

## Verification

- `npm test`: **535 passed** (525 pre-#80 baseline + 3 Back/Forward +
  7 route-table), 41 files, all green.
- `npm run lint -- --max-warnings 0`: clean (the flag was tightened in
  #209 and is unchanged).
- `npm run build`: clean.
- No backend or Android changes — this is a frontend-only PR.
- The running dashboard predates this PR and was not restarted; the UI is
  verified by the unit suites (the route table, the blocker wiring, and
  the migration's route shapes are all pinned there) rather than a live
  browser pass.

## Out of scope

- Data-router loader APIs / `useLoaderData` / scroll restoration — nothing
  in the app uses them; adopting them is a separate feature.
- A custom 404/error boundary — the app deliberately keeps the legacy
  "shell with empty main" for unknown paths.
- Any reformatting or restyling of the touched route files — the
  migration is line-for-line mechanical per the scope-of-changes rule.

Closes #80


teo-mateo:
READY_FOR_REVIEW 942fb563147e8c08b5b7400ca18f051771eec040

Issue #80: editor dirty-state guard for browser Back/Forward — data-router migration + useBlocker.

- Branch: feat/editor-dirty-guard (off main @ cb37e9c)
- Migration: App.jsx builds createBrowserRouter over a new route table (routes.jsx, verbatim paths as a '/' layout route so the shell stays mounted across route changes — the legacy shell-wrapped-<Routes> shape); AppShell/DefaultLayout move to layout.jsx; main.jsx drops the BrowserRouter wrapper. basename /v2 carried over.
- Guard: useBlocker in ChatPage fed by the existing editorDirtyRef chain — both mount points covered (ProjectPage editor + chat-split overlay funnel into the same ref via onEditorDirtyChange). Armed for POP only (PUSH/REPLACE already run confirmDiscardEdits before navigate; blocking them would double-prompt). Stay/Leave modal; Leave re-confirms with the same window.confirm wording.
- No behaviour changes beyond the new Back/Forward prompt: unknown paths still render shell + empty main (no 404 boundary added), route-table matching pinned by routes.test.jsx (7 tests), beforeunload untouched.
- Tests: ChatPage.test.jsx's 18 tests mechanically converted to createMemoryRouter (mandatory — useBlocker throws under legacy MemoryRouter). One verified non-mechanical detail: data routers do not propagate parent rerenders into the route tree (legacy did — probe 1→4 vs 1→1; same-location navigate is a no-op), so the URL-load harness and the default-model rerender test deliver their tick through a context consumed by the route element (TickContext), restoring the tests' original power. New Back/Forward regression (the issue's named test): two history entries, dirty editor, navigate(-1) — cancel preserves route AND editor text; plus Leave commits the Back nav, and a clean editor passes through without a prompt. ChatPageDefaultModel.test.jsx converted the same way.
- Suites: frontend 535 passed (525 + 3 + 7, 41 files), lint --max-warnings 0 clean, build clean. Frontend-only PR; no backend/Android changes.
- Docs: frontend AGENTS.md routing section updated (incl. the /chat/ghost row missing from the route table).

Awaiting review.

teo-mateo (commented):
VERDICT APPROVED

Reviewed #942fb56 against merge base cb37e9c, full diff, both mount points, and a
render-count probe of the router mechanics. Verification:

- **Both dirty-editor mount points are guarded.** `useBlocker` (armed for `POP` only)
  reads `editorDirtyRef`; the `ProjectPage` editor (#79) and the chat-split
  `ProjectFileEditor` overlay (#87) both feed that one ref via
  `onEditorDirtyChange` (ChatPage.jsx:400/407, traced through
  `ProjectPage.handleEditorDirtyChange` and `ProjectChatSplit.notifyDirty`).
  `ProjectFileEditor` reports clean on unmount (its empty-dep cleanup effect) and
  starts dirty for new files, so no stale `true` can keep the blocker armed.
  `beforeunload` is untouched (lives in `ProjectFileEditor`, out of this diff).
- **The regression test #80 names exists and asserts it**: `ChatPage.test.jsx` —
  "cancel preserves the route AND the editor text (issue #80 regression test)":
  `createMemoryRouter` with two history entries, dirty editor, `navigate(-1)`;
  cancel (Stay) keeps the project page's dirty editor mounted (the route cannot
  have committed) and the text intact. `MemoryRouter` → `createMemoryRouter` is
  the correct reading of the issue's "MemoryRouter": `useBlocker` throws under the
  legacy router, which is the point of the migration. Leave and clean-Back
  siblings pass as well.
- **No cosmetic churn**: `git diff -w` vs `git diff` differ by only 5 lines, all
  legitimate structural indentation (fragment wrap, tick-context `act` block,
  `BrowserRouter` unwrap). No re-quoting or restyling of unchanged lines.
- **Suite green at the claimed numbers**: `npm test` 535 passed / 41 files
  (= 525 baseline + 3 Back/Forward + 7 route-table; zero `it` blocks removed);
  `npm run lint` clean with `--max-warnings 0` still in package.json (file not in
  the diff); `npm run build` clean.
- **`routes.jsx` / `routes.test.jsx` / AGENTS.md match the code**: the table is a
  line-for-line match of the legacy one (paths, order, elements, `basename: "/v2"`,
  `AppShell`+`Outlet` = the legacy shell-wrapping-`<Routes>` shape); unknown paths
  → shell-only is pinned; the AGENTS.md route table matches `routes.jsx` including
  the added `/chat/ghost` row, which I confirmed was missing at the merge base.
- **The "no parent-rerender propagation" claim verified, not trusted**: react-router
  7.13.0 renders the route tree through `React.memo(DataRoutes)` with stable
  `routes`/`state` props, and a render-count probe (recreated router 1→4,
  stable router + `TickContext` 1→4, legacy `MemoryRouter` 1→4) confirms both
  converted harnesses still force `ChatPage` to rerender — the converted tests keep
  their original power.

SHOULD  dashboard/frontend/AGENTS.md:443 — "(see `ChatPage.test.jsx`'s `TickContext`)" points at the wrong file: `TickContext`/`ChatPageHost` are defined in `ChatPageDefaultModel.test.jsx`. The `ChatPage.test.jsx` URL-load harness does not use the context mechanism at all — it recreates `createMemoryRouter` on every parent rerender (fresh `routes` array identity breaks the `React.memo` over the route tree), which is what propagates the rerender there (probed: 1→4). Point the gotcha at the file that has it, and optionally name the second mechanism so future harnesses pick a working one.
NIT     dashboard/frontend/src/routes.test.jsx:48 — "serves the project page (beating the optional conversation param)" asserts `element.type === ChatPage`, which both candidate routes satisfy, so the ranking the name claims is not self-verifying. `/chat/project/p1` can only match the project route anyway (`:conversationId?` is one segment), so behaviour is covered — but `expect(r.state.matches[1].params.projectId).toBe('p1')` would make the claim checkable.
NIT     dashboard/frontend/src/components/chat/ChatPage.test.jsx + ChatPageDefaultModel.test.jsx — 14 double-quoted string literals in the new test code (`path: "/chat/…"`) where the surrounding file convention is single quotes (the adjacent `initialEntries: ['/chat/abc']` lines use them).
NIT     dashboard/frontend/src/components/chat/ChatPageDefaultModel.test.jsx:36 — double blank line before the TickContext comment block.
NIT     dashboard/frontend/src/components/chat/ChatPage.jsx:379 — the `<SidebarSplit>` block keeps its pre-fragment indentation (props aligned with the tag) under the new `<>` wrap. Leave it if you prefer the minimal diff — re-indenting ~70 lines is exactly the churn the scope-of-changes rule avoids; noted only for completeness.

PR #227, opened 2026-09-15
fix(ninfer): /metrics stalled behind execution_mutex_, and the scrape-gap math that turned it into spikes

teo-mateo:
## Symptom

On a live NInfer service the Live Metrics panel goes dark for the whole duration of a
generation, points arrive late, and the token lines jump by an order of magnitude the moment
data resumes. Reported for `ninfer-qwen38-27b-nvfp4`.

## Root cause

`GET /metrics` rendered every scrape from `runtime_stats()` **and** `memory_summary()`.

| read | lock | held by |
|---|---|---|
| `runtime_stats()` | `stats_mutex_` | briefly, at publish |
| `memory_summary()` | `execution_mutex_` | `worker_loop()`, across every execution unit (`engine_core.h:1962-2034`), released only for a 1 ms cv wait between units |

So a scrape issued during generation waits for the generation. The patch added no lock; it
inherited one from an existing API. Measured against the deployed image (pin `d4929686`):

| event | wall clock | detail |
|---|---|---|
| generation start | `1789501831.854` | 3000 tokens, non-streaming |
| scrape issued | `1789501833.882` | t+2.0 s into the generation |
| scrape returned | `1789501858.008` | **24.117 s — exactly at generation end** |

4/4 dashboard-shaped scrapes (`timeout=2`) timed out across a 1500-token generation; idle, the
same endpoint answers in 0.4–0.7 ms. The engine's counters were live throughout the stall — its
5 s reporter kept ticking (`decode 106.6 → 110.8 → 115.4 tok/s`) and `/v1/models` answered in
<1 ms while `/metrics` blocked 1.8 s. The pull was stuck, not the data.

The spikes are a second defect riding on top: the dashboard turns a timed-out scrape into
`{"metrics": {}}` with HTTP 200, the hook reset its counter baseline on those empty responses,
and `setInterval(fetchMetrics, 200)` had no in-flight guard — so a generation's accumulated
tokens were divided by one 200 ms window when scrapes resumed in a burst.

This is the contingency §6 of the plan already named ("if the live checks show contention, that
is a patch revision, not a dashboard workaround"). It also corrects the plan's §2 claim that the
scrape shared the reporter's access pattern — the reporter reads `runtime_stats()` only.

## Changes

**Engine patch** (`ninfer/ninfer-metrics.patch`) — `attach()` already read `MemorySummary` once,
before `listen()`, so the KV capacity (fixed at load) is captured there into
`HttpServer::kv_capacity_page_groups_`; `render_metrics_snapshot` takes that integer and the
handler reads `runtime_stats()` alone. `/metrics` no longer touches `execution_mutex_`.

**Dashboard** (`routes/metrics.py`) — a failed scrape returns `scrape_error` instead of collapsing
to `{}`, so the panel can distinguish an unreachable engine from an idle one. First failure of a
streak logs at warning (it was silent at `debug` behind `LOG_LEVEL=INFO`), the rest at debug.

**Frontend** (`useServiceMetrics.js`) — one scrape in flight; a failed scrape appends no point and
leaves the baseline intact, so the next success is divided by the interval it actually spans.

**Docs** — plan §8 records the measurement, the cause and the revision; §2/§6/§3.1 corrections;
`CLAUDE.md` + `check-upstream` skill updated (the skill now states that a regenerated patch must
keep `memory_summary()` out of the handler).

## Verified

- both patches apply to a fresh clone at the pin, port patch then metrics patch (Dockerfile order)
- image compiles with the revised patch (`./build-ninfer.sh`)
- backend 1259 passed; frontend 540 passed, eslint clean — new tests pin `scrape_error`, the log
  streak, single-flight, gap-spanning rates, and no-point-on-failure

## Not verified

**G1–G3 and the under-load probe still need the service running.** The container is stopped and
starting it competes with the running vLLM service for GPU memory, so the latency-under-load gate
is *listed, not claimed* — same gap that let #222 merge with idle-only gates. The expected
result is sub-ms `/metrics` during a generation; if it is not, the contention is deeper than this
handler and §8 is wrong.

## Prior art, not cherry-picked

`sergiuszm/ninfer-5090` independently added `/metrics` (`fc582982`) and later moved its counters
off completion-only totals (`b38ae92`) with a commit message describing this same symptom. Its
trees have diverged, so it stands as precedent that the scrape path is the thing to fix. Upstream
`Neroued/ninfer` still ships no `/metrics` (master `6cc95cc5`, 25 commits ahead of the pin, none
touching serve observability), so the patch's retirement condition has not triggered.


teo-mateo:
VERDICT APPROVED

Verified against the pinned engine source (fresh clone at d4929686):

- Root cause holds: memory_summary() takes execution_mutex_ (engine_core.h:235-236); the worker loop holds the same mutex across the whole execution unit (engine_core.h:1961, released only for the 1 ms cv wait at :2038-2041). runtime_stats() takes stats_mutex_ alone (:252-253), so the revised handler reads no hot lock.
- Startup order holds: model load (GenerationService ctor) → attach() → listen() (apps/serve/main.cpp:63-90), so capturing kv_capacity_page_groups_ at attach is a load-time-fixed value. Type matches the struct field (both std::uint32_t, include/ninfer/types.h:831).
- Both patches apply cleanly to the pin in Dockerfile order (port patch, then metrics patch) — checked with git apply --check on a fresh clone.
- The patched handler reads service_->runtime_stats() + the cached integer only; the 5 s reporter reads runtime_stats() alone (http_server.cpp:367, 380), which makes the §2 correction accurate.
- Backend suite at the PR head: 1 failed, 1258 passed. The one failure (test_flag_surface_guard.py::test_unpinned_llamacpp_image_id_still_current) reproduces identically at the merge base in a scratch copy (1 failed, 1253 passed) — the local llm-dock-llamacpp image was rebuilt after the flag-surface snapshot; environmental, not the diff. The PR adds 5 green backend tests.
- Frontend at the PR head: 538 passed (merge base: 535) — the delta is exactly the 3 new single-flight/gap tests, all green. eslint clean at --max-warnings 0, production build clean.
- Hook changes reviewed: single-flight guard resets in finally (including the unmount/abort early-returns); the no-point-on-failure path leaves prevRef untouched so the next success divides by the real gap; gauges keep their last value during the dark window, which the new test pins.

NIT pr body — "frontend 540 passed": actual head is 538 (535 at base + 3 added); backend "1259" is right when the flag-surface test's image matches. Numbers drifted, substance verified.

PR #229, opened 2026-09-16
Show which image a service runs with, with a provenance popover

teo-mateo:
Two related additions to the v2 service-details page, both answering "what is this service actually running on?":

1. **Image in the metadata row** — the service payload now carries `image`: the running container's creation-time image (`Config.Image`), or the resolved default for not-created services, honouring the `services.json` `image` override for llamacpp/ik_llamacpp/vllm. `flag_metadata.ENGINE_IMAGES` is the single owner of per-engine image defaults (mirrors `ENGINE_INTERNAL_PORTS`), `compose_manager._render_service` reads it, and `tests/test_service_image.py` pins the table against `templates/*.j2`.
2. **Provenance popover** — the Image cell is clickable and opens a popover fetched on demand from `GET /api/services/<name>/image` (deliberately not in the SSE payload): built-locally vs pulled badge, created date + relative age, build date/commit from the `org.llm-dock.build.*` labels (a `-` when the image was built outside the repo's build scripts), registry, upstream source and size. `get_image_build_metadata` is generalised into `get_image_info` in `docker_utils`; source/registry classification is pure and covered by `tests/test_image_info.py` alongside the endpoint.

`docs/image-info-mockups.html` is the design mockup page (four options; the popover — option A — is what shipped).

Verification (real runs, 2026-09-16):
- backend: `venv/bin/python -m pytest tests/ -q` → **1280 passed, 2 skipped**
- frontend: `npm test` → **535 passed**; `npm run build` clean; `ruff check` clean
- live: popover verified in the browser against a running service (`llm-dock-vllm:cu129-580`, built, no build labels) and a pulled image (`ghcr.io/open-webui/open-webui:main`, registry + upstream populated)


teo-mateo:
READY_FOR_REVIEW 807af53cc3033010389541e5f095c38d0adeafaa

No linked issue (direct request). Branch feature/service-image-display, two commits: the payload/image cell, then the provenance popover.

teo-mateo:
VERDICT CHANGES_REQUESTED

BLOCKER CHANGELOG.md — commit 807af53 rewrites pre-existing entries (the 2026-09-15 "Ghost chats" and "NInfer metrics" lines are shortened, 2026-09-13 lines shortened, and older lines deleted/merged) — checked by diffing each commit against the merge base: commit 0ca6a1b only adds the new section, commit 807af53 carries the rewrites. That is history-rewriting of a shared file, which AGENTS.md's scope rule puts in its own commit — and it is exactly what makes the PR CONFLICTING against main (main's entries were amended since the branch point, and #227 just appended to the same section). Fix: keep only the new `## 2026-09-16` section, drop every change to pre-existing lines, rebase onto main, and post a fresh READY_FOR_REVIEW <sha>.

Everything else verified at head 807af53:

- `flag_metadata.ENGINE_IMAGES` + `default_engine_image()` mirror the `ENGINE_INTERNAL_PORTS` pattern; `compose_manager._render_service` reads it for both override-capable branches (ik_llamacpp/llamacpp and vllm); ds4/tabbyapi/ninfer keep their hardcoded template images, and the new `tests/test_service_image.py::test_table_matches_the_templates` pins the table against all six `templates/*.j2` (each has exactly one image line: `{{ image }}` for the three override engines, the literal for the other three — matches what the templates render).
- `get_docker_services()` payload: the created branch reports `container.attrs["Config"]["Image"]` (creation-time reality, test pins that it wins over a later services.json override), the not-created branch the resolved default honouring the `image` override only for the three engines that render `{{ image }}`; unknown engine → `""`. Map population order is safe (`template_type_map` set in the same `if config:` block before use).
- `get_image_build_metadata` → `get_image_info`: no remaining callers of the old name (grep over dashboard/); the legacy `dashboard/static/js/utils.js` consumer of `/api/images/metadata` reads only `exists`/`build_date`/`build_commit`, all of which survive in the new shape.
- New `GET /api/services/<name>/image`: authed, exact-name lookup against the service list, image name goes to `client.images.get()` (SDK exact reference, no CLI/shell interpolation) — no new injection surface.
- `_image_source` / `_image_registry` are pure and covered by `tests/test_image_info.py` (build label wins over name, `llm-dock-` prefix fallback, ghcr/localhost:port/docker.io/bare-name cases, built-with-labels, pulled-with-OCI-source, missing image, endpoint inspects the payload's image, 404 unknown service).
- Frontend: `ImagePopover` lazy-fetches on first open, caches in state, closes on outside click/Escape, stops propagation on the upstream link (`rel="noopener noreferrer"`), failure closes the popover and reports through the page's existing `onError`; `useServiceDetails.actions.fetchImageInfo` wired through `ServiceDetailsPage` (actions/onError already flow to `ServiceDetailsHeader` for the toolbar row).
- The two one-line test diffs add `Config: {Image}` to the fake containers' attrs — required because `get_docker_services` now reads that key; in scope.
- Backend suite at head: 1280 passed (exact match with the PR body), 1 failed, 1 skipped. The one failure is `test_flag_surface_guard.py::test_unpinned_llamacpp_image_id_still_current` — the local `llm-dock-llamacpp:latest` (58dd84544286) no longer matches the recorded flag-surface snapshot (4e9495854d69); the same test failed identically at this PR's merge base, so it is environmental (image rebuilt since the snapshot), not the diff. Frontend: 535 passed (exact match), eslint clean at --max-warnings 0, production build clean.

SHOULD docker_utils._image_registry — a bare single-component name that is a Docker Hub image (`nginx:latest`) classifies as "pulled" with registry `None`, so the popover shows no registry row for it; harmless here (local services use `llm-dock-*` or registry-qualified names) but a `docker.io` fallback for the no-slash pulled case would make the rule total.

PR #231, opened 2026-09-18
feat: smooth token throughput sparkline with animated runner

teo-mateo:
## Summary

Adds a smoothing layer to the token throughput sparkline and an animated pixel-art runner overlay.

## Changes

- tokenSeries.js — extracted series preparation logic: moving-average smoothing over raw scrapes, axis scale computation, and the animation clock (frame stepping, pose cycling).
- runnerSprite.js — pixel-art runner sprite frames, pose animation, and canvas drawing helpers.
- TokenSparkline.jsx — draws smoothed lines over the raw data (raw kept visible at low alpha), manages the canvas backing store only when the box changes, and advances the animation frame.
- useServiceMetrics.js — feeds the new series/scale/frame shape to the sparkline.
- utils.js — small helper additions.
- Tests — tokenSeries.test.js and runnerSprite.test.js for the extracted logic; updated TokenSparkline.test.jsx and useServiceMetrics.test.js.


teo-mateo:
VERDICT APPROVED

Reviewed the whole diff against the merge base, and ran everything in the exact state that would ship (this PR's head auto-merges cleanly onto main post-#227; that merged tree is what the runs below used):

- Frontend suite at the merged state: 584 passed (538 on current main + exactly the 46 tests this PR adds: 5 new TokenSparkline, 17 runnerSprite, 23 tokenSeries, 1 new useServiceMetrics). eslint clean at --max-warnings 0, production build clean. No Python touched.
- Semantic interaction with #227 verified: #227's failed-scrape path appends no points, so `history` (and thus the memoised `series`) is stable during a scrape gap — the runner keeps its last pose cadence and the badges keep their last values while the hook's `error` surfaces the outage; no double-append or baseline interaction with the single-flight guard.
- tokenSeries.js math checked by hand: smoothRate is a dt-weighted trailing-window mean (the weight argument is load-bearing — a bunched poll covers a short window; the 91.818 case in the test is the discriminator), falls back to the plain mean only when no point carries a dt; niceCeiling(216) = 240 and every gridline is an integer (ceiling is step·magnitude·4 with a floor of 4); advanceScale snaps up immediately, holds 10 s, and an unchanged candidate does not re-arm the hold — all pinned by tests.
- prepareSeries scales the axis off the smoothed peak (one 400-t/s spike in a steady 80 series yields peak 144, ceiling 240, not 400/600) and the badges report the unpadded peak, so the "max" numbers and the line cannot disagree; the raw underlay at 0.22 alpha keeps the burstiness visible.
- TokenSparkline.jsx: the canvas backing store is only reallocated when the box actually changed (immediate-mode canvas, so the old per-draw assignment paid for it every stride under the animation loop); setTransform instead of scale because a skipped assignment keeps the previous transform; the rAF loop repaints only on a pose change and is not scheduled at all when genRate ≤ IDLE_GEN_RATE, so an idle service costs zero frames; rAF is cancelled on cleanup.
- The `POLL_INTERVAL` move to utils.js is justified in-place: tokenSeries.SMOOTH_LABEL derives the window label from the same constant, so the chart cannot claim a window it does not average.
- useServiceMetrics.js: one data-point field added (`dt`, the interval each rate covers) plus the constant move; the two one-line fake-container diffs are not in this PR (that was #229) — this PR's hook test file change is one new test only.

SHOULD pr body — no verification numbers ("tests pass" territory): the merged-state run is 584 passed / lint clean / build clean, recorded here instead.
SHOULD CHANGELOG.md — no entry; both sibling PRs (#227, #229) carry one, and the page change is user-visible. A one-liner under a new date section belongs in the follow-up, not a gate.

PR #232, opened 2026-09-18
fix(v2): service creation — the dead + New Service button (#230)

### LINKED ISSUE #230: When I click on the 'New Service' button in v2 I get a page saying the service 'new' does not exist
Image

teo-mateo:
Fixes #230.

## What was broken

The **\+ New Service** button in the v2 dashboard navigated to `/services/new`. The data router's only match there is `services/:serviceName/*`, so React rendered the details page with `serviceName = "new"` — which fetches `/api/services/new`, 404s, and shows the error from the issue screenshot.

Deeper than a route: **v2 had no service-creation flow at all.** `ServiceConfigPanel` is a PUT-only editor, and the only create path was the legacy vanilla-JS modal (`static/js/service-modal/create-service.js`) — and that one only offers three of the six engines.

## What this PR does

Ports creation to the v2 SPA as a `CreateServiceModal` on the Services page (the button now opens it instead of navigating):

- **All six engines** — llama.cpp, ik llama.cpp, vLLM, ds4, TabbyAPI, NInfer — with per-engine model field: a discovered-file picker (over `GET /system/info`, host→container path conversion mirroring `_CONTAINER_PATH_MAP`) plus a manual container-path input, or a HuggingFace org/name for vLLM with a cached-model picker (containers run `HF_HUB_OFFLINE=1`, so the hint says so)
- **Alias with live service-name preview** — mirrors the server's `generate_service_name` + `sanitize_service_name` in `utils/serviceNaming.js` (preview only; the server owns the real name)
- **Next-free port in 3301–3399** computed from the live service list
- **Optional API key** (auto-generated server-side when empty), **reasoning levels**, and **flag+value params** with the existing `ParameterReference` docked in a right column (sticky while the form scrolls)
- **Collapsible command preview** — the details page's `renderCommandPreview` moved to `components/commandPreview.js` so both views share it (kept out of the component file for the react-refresh lint rule)
- **Success view** — shows the created name/port, the generated API key with copy button, any unknown-flag warnings from the #198 guard, and an **Open service** button that navigates to the new service's details page

Backend change is one line of plumbing: `fetchAPI` now attaches the structured `details[]` array from a 400 body as `err.details` (same pattern as `err.code`), and the modal renders server validation errors verbatim.

## Verification

- 553/553 frontend tests pass (9 new modal tests + 7 naming-mirror tests), lint clean, build clean
- Live browser pass against the running dashboard: button opens the modal (no more 'service new does not exist'), file picker populated from real discovery (13 GGUF entries), picking a file fills the container path + suggests an alias, name preview updates per engine prefix (llamacpp-/exl-…), next port computed from in-use ports, vLLM swap swaps the model field and picker, reference click adds a param row with its default, reference panel stays pinned while the form scrolls


teo-mateo:
## Verification (2 rounds, live browser + static)

**Round 1** (against `1303ca1`): PASS-WITH-NITS — full functional matrix green, two MAJOR findings, five nits. All fixed in `0c01bcf`.

**Round 2** (against `0c01bcf`): **PASS** — all six fixes confirmed live, full matrix re-run fresh, no regressions.

Fixes from round 1, confirmed in round 2:
- **ik_llamacpp flag-metadata 400** (route allow-list gap; reference column vanished for 1 of 6 engines) → fixed + `test_flag_metadata_route.py` pins "every service `template_type` serves its metadata"
- **390px params-row overflow** (clipped value field + remove button) → `min-w-0`/`w-32 sm:w-44`; measured `scrollWidth == clientWidth` at exactly 390×844, both themes
- **GGUF picker footguns** → mmproj excluded, sharded files collapsed to one entry (`— N shards`, fills shard-1 path)
- **a11y** → Escape closes (form + success views), `role="dialog"`/`aria-modal`, labeled pickers and param inputs
- **clipboard** → awaited + guarded (secure-origins-only API; LAN-IP deployments would have thrown)
- **test gap** → adapter-level `err.details` transport tests in `api.test.js`

Round-2 checks: 557/557 frontend tests, lint + build clean, 38 backend tests; live create→verify→delete for **llama.cpp and ik_llamacpp** (config byte-matched against `GET /api/services/<name>`, no leftover services/containers); command preview correct per engine; zero console errors; visual geometry clean at 1440/1024/390 × dark/light.

Pre-existing observations (outside this diff, not fixed here): vLLM picker can list GGUF-only repos whose active snapshot is empty (discovery attaches no files, so the filter can't see them); case-duplicated orgs + junk cache entries in the picker; `ik_llama` truncation in the table's ENGINE cell.

Issue #238, opened 2026-09-18
inspector: read/delete HTTP API


teo-mateo:
Part of #234. **Depends on: #235** (#237 not required — the endpoints read the DB, and
`proxy_supervisor.status()` can be stubbed until #237 lands; if #237 is already merged,
wire the real one).

## Scope

The HTTP surface the viewer reads: list captures newest-first, fetch one in full, delete
one, delete many. No search — ordering by time and an optional service filter is the whole
query surface, by design.

## Files

- `dashboard/routes/inspector.py` — `inspector_bp`
- `dashboard/routes/__init__.py` — export it
- `dashboard/app.py` — register it after `metrics_bp`
- `dashboard/tests/test_inspector_routes.py`

All endpoints carry `@require_auth`. Follow the existing blueprints for error shape:
`jsonify({"error": "..."}), <code>`.

### `GET /api/inspector/captures`

Query params: `service` (optional), `limit` (default 50, max 200), `offset` (default 0).

```json
{
  "captures": [
    {"id": "…", "service_name": "llamacpp-qwen3", "model": "qwen3.8-27b",
     "template_type": "llamacpp", "method": "POST", "path": "/v1/chat/completions",
     "status_code": 200, "stream": true, "prompt_tokens": 1843, "completion_tokens": 402,
     "total_tokens": 2245, "ttfb_ms": 412, "duration_ms": 8391, "error": null,
     "request_truncated": false, "response_truncated": false,
     "created_at": "2026-09-18T20:14:03.412Z"}
  ],
  "total": 137,
  "limit": 50,
  "offset": 0
}
```

Ordered newest first. **No body fields** on a list row (#235's summary read). `stream`,
`request_truncated` and `response_truncated` are JSON booleans, not the stored 0/1.

### `GET /api/inspector/captures/<capture_id>`

The full row, adding `request_headers` (parsed object, already redacted at capture time),
`request_body` (string, verbatim as sent), `response_body` (string), `response_text`,
`reasoning_text`, `tool_calls` (parsed array or `[]`), `finish_reason`. 404 on an unknown
id. `request_body` is returned as a **string**, not re-parsed — a malformed or truncated
body must still be fetchable and displayable.

### `DELETE /api/inspector/captures/<capture_id>`

`{"deleted": 1}` on success, 404 if it does not exist.

### `DELETE /api/inspector/captures`

Optional `?service=<name>`. Deletes every matching capture; `{"deleted": <n>}`.

[...]

Issue #237, opened 2026-09-18
inspector: enable/disable lifecycle — services.json field, compose port relocation, proxy supervisor


teo-mateo:
Part of #234. **Depends on: #236.** This is the only sub-issue that touches compose
rendering and container recreation.

## Scope

Make the proxy a per-service toggle: the `inspect` field in `services.json`, the compose
port relocation, upstream-port allocation, the enable/disable endpoint, and a supervisor
that keeps the running proxies in step with the DB across boot, toggle, rename, port
change and delete.

## Files

- `dashboard/inspector/ports.py` — `allocate_upstream_port` (pure)
- `dashboard/inspector/supervisor.py` — `ProxySupervisor` + the `proxy_supervisor` singleton
- `dashboard/compose_manager.py` — port relocation in `_render_service`
- `dashboard/docker_utils.py` — host-port authority + payload fields
- `dashboard/routes/services.py` — `POST /api/services/<name>/inspect`, sync on the other write paths
- `dashboard/app.py` — boot-time init + `atexit` shutdown
- `dashboard/tests/test_inspector_ports.py`, `dashboard/tests/test_inspector_supervisor.py`,
  `dashboard/tests/test_inspector_compose.py`, `dashboard/tests/test_inspector_toggle_route.py`

## Schema addition (`services.json`)

| Field | Type | Notes |
|---|---|---|
| `inspect` | bool | Optional, default false. Chat/UI-facing like `favorite`; never a container flag. |
| `inspect_upstream_port` | int | Optional. Allocated at first enable, **retained** after a disable so re-enabling doesn't churn the container. |

Add both rows to the `services.json` schema table in `CLAUDE.md` (#241 owns the wider doc
pass, but this table entry belongs with the code that reads it).

`validate_service_config()` accepts `inspect` as a bool and rejects a non-bool, the same
way `favorite` is handled. `inspect_upstream_port` is server-owned: a `POST`/`PUT
/api/services` body **must not** be able to set it — strip it from client input.

## Port allocation (`ports.py`, pure)

```python
UPSTREAM_PORT_BAND = range(34000, 35000)

def allocate_upstream_port(services: dict, service_name: str) -> int: ...
```

Returns the existing `inspect_upstream_port` if the service already has one, else the
lowest free port in the band, where "free" means not equal to any other service's `port`
or `inspect_upstream_port`. Raises `ValueError` when the band is exhausted.

## Compose relocation (`compose_manager._render_service`)

[...]

Issue #236, opened 2026-09-18
inspector: intercepting HTTP proxy core — forwarding, streaming, capture


teo-mateo:
Part of #234. **Depends on: #235.** Pure backend, no Docker, no GPU — tests run against
a fake upstream HTTP server started in-process.

## Scope

The intercepting proxy itself: bind a port, forward every request to an upstream
`127.0.0.1:<port>` verbatim, relay the response as it arrives, and record the
completion-style calls into `InspectorDB`. This issue does **not** wire the proxy to
services.json or compose — #237 does. Deliverable is a class that can be started against
any (host, port) pair.

## Files

- `dashboard/inspector/sse_assembly.py` — pure response assembly
- `dashboard/inspector/proxy.py` — `ServiceProxy`
- `dashboard/tests/test_inspector_sse_assembly.py`
- `dashboard/tests/test_inspector_proxy.py`

## `sse_assembly.py` (pure — write and test this first)

```python
def assemble_stream(raw: str) -> dict: ...
def assemble_json(raw: str) -> dict: ...
```

Both return the same shape, with `None`/empty for anything absent:

```python
{"text": str, "reasoning": str, "tool_calls": list, "finish_reason": str | None,
 "usage": {"prompt_tokens": int, "completion_tokens": int, "total_tokens": int} | None}
```

`assemble_stream` parses an OpenAI-style SSE body: lines starting with `data: `, the
terminal `data: [DONE]`, one JSON object per data line.

- `choices[0].delta.content` concatenates into `text`.
- `choices[0].delta.reasoning_content` **or** `.reasoning` concatenates into `reasoning`
  (llama.cpp and vLLM disagree on the name; both appear on this wire).
- `choices[0].delta.tool_calls` accumulates **by `index`**: `id` and `function.name` are
  set once, `function.arguments` concatenates. The result is a list ordered by index.
- `usage` is taken from whichever chunk carries a non-null `usage`.
- `finish_reason` is the last non-null `choices[0].finish_reason`.
- A malformed data line is skipped, not raised on. A body that parses to nothing returns
  the empty shape rather than raising — a truncated capture must still store.

`assemble_json` reads the non-streaming shape: `choices[0].message.{content,
reasoning_content|reasoning, tool_calls}`, `choices[0].finish_reason`, `usage`. An
embeddings response (no `choices`) returns the empty shape plus its `usage`.

## `ServiceProxy`

```python
class ServiceProxy:
[...]

Issue #234, opened 2026-09-18
[PLAN] Request inspector: a toggleable proxy in front of each model + payload viewer


teo-mateo:
## What

A per-service **request inspector**: a toggleable HTTP proxy that sits in front of a model
service on **the service's own host port**, forwards every request to the container
untouched, and records the full OpenAI-compatible payload — system prompt, all messages,
tool definitions, tool calls, sampling fields, reasoning fields — plus the response. A new
**Inspector** page in the v2 frontend renders the captures, newest first, and can delete
them.

Motivation: when a model behaves oddly in chat, there is today no way to see what was
actually sent. `chat/llm_proxy.py` builds a payload from conversation history, system
prompt, MCP tool definitions, reasoning level and sampling params; the only evidence of
what came out is the model's answer. This makes the wire payload directly inspectable,
for **every** host-side client (llm-dock chat, ghost chat, the Android client, `curl`,
benchmark scripts), not just chat.

## Architecture (decided — implement as written)

**The proxy runs inside the dashboard process and binds the service's public host port;
the container is relocated to a loopback-only upstream port.**

```
before:  client ──► 0.0.0.0:3317 ──► container :8080
after:   client ──► 0.0.0.0:3317 (dashboard proxy thread) ──► 127.0.0.1:34001 ──► container :8080
                          │
                          └──► inspector.db (captures)
```

**Inspection is per service, on whatever port that service holds, and any number of
services can be inspected at once** — one listener and one upstream port each. There is
nothing special about the 3301 public slot here; 3317 above is just an example port. A
service that is moved onto or off the public slot by `set-public-port` keeps its
inspection state and its proxy simply rebinds to the new port.

- `services.json` gains two fields: `inspect` (bool) and `inspect_upstream_port` (int,
  allocated from the 34000–34999 band at first enable and kept thereafter).
- When `inspect` is true, `_render_service()` publishes the container as
  `"127.0.0.1:<upstream>:<internal>"` instead of `"<port>:<internal>"`. All six templates
  already render `{{ port }}:<internal>`, so this is one change point in
  `compose_manager.py`. Binding the upstream to loopback means the proxy is not
  bypassable from the LAN.
- The dashboard runs one `ThreadingHTTPServer` per inspected service on `0.0.0.0:<port>`.
  Clients keep the same URL and the same API key — nothing outside the dashboard is
  reconfigured.
- Capture is path-scoped: only POSTs to the completion-style paths are recorded, so
[...]

PR #243, opened 2026-09-18
feat(inspector): storage layer for captured payloads (#235)

### LINKED ISSUE #235: inspector: storage layer — inspector.db schema, CRUD, retention, redaction
Part of #234. **Depends on: nothing.** Pure backend, no Docker, no GPU.

## Scope

The storage layer for captured payloads: a new SQLite database, its CRUD surface,
retention trimming, header redaction and body truncation. Nothing in this issue proxies
anything or is reachable over HTTP — #236 and #238 are the callers.

## Files

- `dashboard/inspector/__init__.py` — empty (the package's wiring lands in #237)
- `dashboard/inspector/models.py` — `Capture` dataclass
- `dashboard/inspector/db.py` — `InspectorDB`
- `dashboard/inspector/redact.py` — `redact_headers`, `truncate`
- `dashboard/tests/test_inspector_db.py`

Follow `dashboard/benchmarking/db.py` for the class shape: module-level `SCHEMA_SQL`,
`__init__(self, db_path)`, a persistent connection when `db_path == ":memory:"`, and
`_get_conn` / `_close_conn` helpers.

## Schema

```sql
CREATE TABLE IF NOT EXISTS captures (
    id                   TEXT PRIMARY KEY,
    service_name         TEXT NOT NULL,
    model                TEXT,
    template_type        TEXT,
    method               TEXT NOT NULL,
    path                 TEXT NOT NULL,
    status_code          INTEGER,
    stream               INTEGER NOT NULL DEFAULT 0,
    request_headers_json TEXT NOT NULL DEFAULT '{}',
    request_body         TEXT NOT NULL DEFAULT '',
    response_body        TEXT,
    response_text        TEXT,
    reasoning_text       TEXT,
    tool_calls_json      TEXT,
    finish_reason        TEXT,
    prompt_tokens        INTEGER,
    completion_tokens    INTEGER,
    total_tokens         INTEGER,
    ttfb_ms              INTEGER,
    duration_ms          INTEGER,
    error                TEXT,
    request_truncated    INTEGER NOT NULL DEFAULT 0,
    response_truncated   INTEGER NOT NULL DEFAULT 0,
    created_at           TEXT NOT NULL DEFAULT (strftime('%Y-%m-%dT%H:%M:%fZ','now'))
);

CREATE INDEX IF NOT EXISTS idx_captures_created ON captures(created_at DESC);
CREATE INDEX IF NOT EXISTS idx_captures_service ON captures(service_name, created_at DESC);
```

`created_at` uses `%f` (millisecond) precision, **not** the `%S` the chat and benchmark
schemas use: several captures per second is the normal case here and second-resolution
timestamps would make "order by time" unstable. Every ordered query still adds
`rowid DESC` as the tiebreak.

`request_body` is stored as the **raw request text** exactly as received (after
truncation) — not re-serialized. The point of the feature is to see what went on the wire.

## API

```python
class InspectorDB:
    def __init__(self, db_path: str = "inspector.db") -> None: ...
    def insert_capture(self, capture: Capture) -> Capture: ...
    def get_capture(self, capture_id: str) -> Optional[Capture]: ...
    def list_captures(self, service: Optional[str] = None, limit: int = 50,
                      offset: int = 0) -> Tuple[List[Capture], int]: ...
    def delete_capture(self, capture_id: str) -> bool: ...
    def delete_captures(self, service: Optional[str] = None) -> int: ...
    def rename_service(self, old_name: str, new_name: str) -> int: ...
    def trim(self, max_rows: int = MAX_CAPTURES) -> int: ...
    def services_with_captures(self) -> List[str]: ...
```

- `list_captures` returns `(rows, total_matching)` and its rows carry **no body fields** —
  `request_body`, `response_body`, `response_text`, `reasoning_text`,
  `request_headers_json` and `tool_calls_json` are `None` on a list row. The list endpoint
  must not pay for megabytes it won't render. Add a `summary: bool` parameter or a
  separate column list; either is fine, but the SELECT must not read the body columns.
- `trim` deletes the oldest rows beyond `max_rows` and returns how many it deleted.
  `MAX_CAPTURES = 2000`, a module constant.
- `rename_service` exists for #237's service rename path.
- `delete_captures(service=None)` deletes everything when `service` is None.

## Redaction and truncation

`dashboard/inspector/redact.py`, both pure:

```python
REDACTED = "***redacted***"
SENSITIVE_HEADERS = {"authorization", "x-api-key", "api-key", "x-totp-code", "cookie"}
MAX_BODY_CHARS = 1_000_000

def redact_headers(headers: dict) -> dict: ...
def truncate(text: str, cap: int = MAX_BODY_CHARS) -> tuple[str, bool]: ...
```

- `redact_headers` compares header names case-insensitively, replaces the **whole** value
  with `REDACTED` (never a prefix of the real key), and preserves the original header
  casing in the returned dict.
- `truncate` returns `(text, False)` when under the cap, and
  `(text[:cap] + "\n…[truncated by llm-dock inspector]", True)` when over it.

## Acceptance criteria

- [ ] `InspectorDB(":memory:")` works without touching disk; `InspectorDB(path)` creates
      the file and schema on first use and is idempotent on a second construction.
- [ ] Insert → `get_capture` round-trips every field, including `None`s.
- [ ] `list_captures()` returns newest first. A test that inserts 5 rows inside the same
      second still gets a deterministic, insertion-reverse order.
- [ ] `list_captures(service="x")` filters, and its `total` is the count of matching rows,
      not of returned rows; `limit`/`offset` paginate.
- [ ] List rows carry no body fields (asserted explicitly).
- [ ] `delete_capture` returns True once and False on a second call for the same id.
- [ ] `delete_captures()` empties the table and returns the count; `delete_captures("x")`
      leaves other services' rows alone.
- [ ] `trim(3)` on 10 rows leaves the 3 newest and returns 7.
- [ ] `rename_service("a", "b")` moves exactly the matching rows.
- [ ] `services_with_captures()` returns distinct service names.
- [ ] `redact_headers({"Authorization": "Bearer key-abc"})` yields the literal
      `***redacted***` — a test asserts `"key-abc" not in str(result)`.
- [ ] `truncate` flags and marks an oversized body; a body at exactly the cap is not
      flagged.
- [ ] `cd dashboard && pytest tests/test_inspector_db.py` passes; `pytest tests/` shows no
      new failures.

teo-mateo:
Part of #234 — the first sub-issue, no dependencies. Nothing here proxies anything or is reachable over HTTP; #236 and #238 are the callers. `dashboard/inspector/__init__.py` is empty; package wiring lands in #237.

## What

The storage layer for captured payloads:

- **`dashboard/inspector/models.py`** — `Capture` dataclass mirroring the 24 `captures` columns (plus `to_dict`).
- **`dashboard/inspector/db.py`** — `InspectorDB` following the `benchmarking.db` shape: module-level `SCHEMA_SQL`, persistent connection for `:memory:`, per-call connections with `PRAGMA journal_mode=WAL` for file paths, `_get_conn`/`_close_conn`. Schema exactly as specified in #235: millisecond `created_at` (`strftime('%Y-%m-%dT%H:%M:%fZ','now')`), both indexes, and `rowid DESC` as the tiebreak on every ordered query. `MAX_CAPTURES = 2000`.
- **`dashboard/inspector/redact.py`** — pure `redact_headers` (case-insensitive match, whole value replaced with `***redacted***`, original casing preserved) and `truncate` (1 MB cap, exact marker, a body at exactly the cap is not flagged).

## Design notes

- **List rows never read body columns.** `list_captures` selects an explicit 18-column summary list; the six body fields (`request_body`, `response_body`, `response_text`, `reasoning_text`, `request_headers_json`, `tool_calls_json`) are `None` on a list row — asserted explicitly in the tests.
- **`insert_capture` honours an explicit `created_at` when set**, falling back to the DB `strftime` default otherwise. The #235 same-second tie test needs a controllable stamp (the real clock cannot be forced into a tie), and #236 will want the request-start time, not the post-response insert time. `created_at = None` keeps the schema default.
- **`trim` keeps the newest N by `(created_at DESC, rowid DESC)`** and returns how many it deleted; `delete_captures(None)` empties the table; `rename_service` mirrors the benchmark DB for #237's service-rename path.
- `request_headers_json` is a `Dict[str, str]` on the dataclass (redacted before storage, per constraint 1 of #234); `tool_calls_json` is raw JSON text, stored verbatim — the storage layer does not interpret it.

## Verification

- `cd dashboard && venv/bin/python -m pytest tests/test_inspector_db.py -q` → **32 passed** in 0.2s (construction, full/None round-trips, same-second ordering, filter/pagination, no-body list rows, delete, trim, rename, services, file persistence, redaction, truncation)
- `venv/bin/python -m pytest tests/ -q` (unpipeled) → **1319 passed, 1 failed, 1 skipped** in 84.5s
  - the single failure is `tests/test_flag_surface_guard.py::test_unpinned_llamacpp_image_id_still_current`: the local `llm-dock-llamacpp:latest` image is `58dd84544286` while the recorded flag-surface snapshot names `4e9495854d69` — the image was rebuilt on this host after the snapshot was recorded. Verified environmental per protocol: the **full suite at the merge base** (`origin/main` `c9af463`) in a scratch worktree gives **1 failed, 1287 passed, 1 skipped** — same failure id, same count. This diff adds files only (5 new files, 747 insertions, 0 modified), so no existing test's input changed.
- `ruff check dashboard/inspector/ dashboard/tests/test_inspector_db.py` → clean

Closes #235


teo-mateo:
READY_FOR_REVIEW bf31209e298b6cbb12dbb89f712e7cabe71ffdc8

teo-mateo:
VERDICT APPROVED

Verified against the #235 spec and the merge base (`c9af463`), not just the PR body:

- Schema matches #235 column-for-column (24 columns, both indexes, `%f` millisecond `created_at` default); all 9 API methods carry the spec'd signatures.
- All 13 acceptance criteria are pinned by tests: same-second tie ordering via `rowid DESC`, filter/pagination with `total` = matching count, no-body list rows asserted explicitly, `delete_capture` True-then-False, trim/rename/delete-all, redaction literal + `"key-abc" not in str(result)`, exactly-at-cap not flagged.
- Shape mirrors `benchmarking/db.py` (module `SCHEMA_SQL`, persistent `:memory:` conn, per-call WAL conn, `_get_conn`/`_close_conn`); `rename_service` is identical to the benchmark one for #237.
- Ran `pytest tests/test_inspector_db.py -q` → 32 passed; full suite unpipeled → 1319 passed, 1 failed, 1 skipped. The single failure (`test_unpinned_llamacpp_image_id_still_current`) reproduces identically at the merge base in a scratch worktree — the local `llm-dock-llamacpp:latest` was rebuilt after the flag-surface snapshot; environmental, not this diff. `ruff check` on the new files is clean.
- 5 new files, 747 insertions, 0 modified — no unrelated restyling. `*.db` is gitignored, same exposure as `chat.db`/`benchmarks.db`.

SHOULD dashboard/inspector/models.py:16 — `request_headers_json: Dict[str, str]` and `request_body: str` are `None` on list rows (spec-mandated, test-pinned), so both annotations should be `Optional[...]` — the dataclass is what #236/#238 callers will read, and the type is where this shape belongs. Non-blocking: runtime behaviour is exactly the spec'd one.
NIT dashboard/inspector/db.py:8 — `logger` is never used (the benchmarking db's logger earns its keep in `recover_stale_runs`; this one logs nothing).

PR #244, opened 2026-09-18
feat(inspector): intercepting proxy core with streaming relay and capture (#236)

### LINKED ISSUE #236: inspector: intercepting HTTP proxy core — forwarding, streaming, capture

teo-mateo:
Part of #234 — sub-issue #236 (proxy core). Depends on #235 (merged in #243). Nothing here wires the proxy into `services.json` or compose — #237 does. Deliverable is a class that can be started against any `(host, port)` pair.

## What

**`dashboard/inspector/sse_assembly.py`** (pure, tested first per the issue):
- `assemble_stream` / `assemble_json` return the same 5-key shape; `None`/empty for anything absent.
- `assemble_stream`: `data:` lines, terminal `[DONE]`, `choices[0].delta.content` → `text`; `reasoning_content` **or** `reasoning` → `reasoning` (llama.cpp and vLLM disagree on the name — both accepted); `tool_calls` accumulated **by `index`** with `id`/`function.name` set once and `function.arguments` concatenated, result ordered by index; `usage` from the chunk carrying a non-null one; `finish_reason` the last non-null. A malformed data line is skipped, not raised; a body that parses to nothing returns the empty shape.
- `assemble_json`: `choices[0].message.{content, reasoning_content|reasoning, tool_calls}`, `finish_reason`, `usage`; an embeddings response (no `choices`) returns the empty shape plus its `usage`.

**`dashboard/inspector/proxy.py`** — `ServiceProxy`:
- `ThreadingHTTPServer` subclass with `daemon_threads = True` and `allow_reuse_address = True`; one `BaseHTTPRequestHandler` with `protocol_version = "HTTP/1.1"` routing `GET/POST/PUT/PATCH/DELETE/HEAD/OPTIONS` to a single `_forward()`.
- **Forwarding**: upstream call is `requests.request(method, url, headers=..., data=body, stream=True, timeout=(10, 1800), allow_redirects=False)` on the full original path including query. The read timeout is large because a long generation legitimately takes minutes. Hop-by-hop headers (`connection, keep-alive, proxy-authenticate, proxy-authorization, te, trailers, transfer-encoding, upgrade`) are dropped **both directions**; `Host` is rewritten to the upstream; **`Authorization` is forwarded verbatim** (engine-side key enforcement stays intact).
- **Response**: status + headers copied through minus hop-by-hop. If upstream sent `Content-Length` it is forwarded and the body written; if it did not (the streaming case) the proxy omits `Content-Length`, sends `Connection: close`, sets `self.close_connection = True`, and relays raw bytes as they arrive. Relay granularity is `iter_content(chunk_size=None)` with a `wfile.flush()` after every chunk.
- **Failures**: upstream unreachable → `502` `{"error": {"message": "llm-dock inspector: upstream for <service> is not reachable on 127.0.0.1:<port>", "type": "upstream_unavailable"}}`; read timeout → `504`, same shape. A `BrokenPipeError`/`ConnectionResetError` while writing to the client stops the relay, closes the upstream, and still persists the row with `error = "client disconnected"`.
- `start()` never raises on a bind failure — it records the message in `bind_error`, logs a warning, and leaves `running` False. `stop()` is idempotent and releases the port.

**Capture**: a row is written iff the method is POST, the query-stripped path ends with one of `CAPTURED_PATHS`, and the request `Content-Type` starts with `application/json` — so `/metrics`, `/health`, `/props`, `/slots` and `/v1/models` write no row. `model` falls back to the `default_model` constructor arg when the body has no `model`. `request_body` is the raw request text after `truncate`; `response_body` is the relayed bytes decoded `utf-8`/`errors="replace"` after `truncate`, accumulated into a bounded buffer that **stops appending once the cap is reached** while relaying continues. Headers are `redact_headers`-ed before storage. `ttfb_ms`/`duration_ms` are first/last byte times. After a successful insert, `db.trim(MAX_CAPTURES_ROWS)`. Every capture step is wrapped in `try/except Exception` logged at WARNING — a capture failure never changes the relayed bytes and never raises out of the handler. `log_message` is silenced to `logger.debug`.

**`db.py` (from #235) — one touch.** The proxy is the first multi-threaded caller: handler threads share the instance, and the persistent `:memory:` connection refused cross-thread use (`sqlite3.ProgrammingError`). The fix is an `RLock` serialising every public method plus `check_same_thread=False` on the in-memory connection — the sqlite3 docs make that serialisation the caller's job. File mode already opens a connection per call, so it's a no-op there. `MAX_CAPTURES_ROWS` in `proxy.py` is `MAX_CAPTURES` re-exported, not a second literal.

## Test semantics worth knowing

- The fake upstream speaks **HTTP/1.1 + `Transfer-Encoding: chunked`** for streaming, like the real engines. For a close-delimited HTTP/1.0 body, `iter_content(None)` reads to EOF, so chunked is the faithful streaming case.
- The first-byte-before-last-send assertion reads with **`iter_lines(chunk_size=1)`**. The proxy answers the client *without* `Content-Length` (it dropped the upstream's chunking), and `http.client._safe_read(n)` blocks until `n` bytes accumulate *or* the connection closes — so the default 512-chunk `iter_lines` cannot observe the first byte before close. At `chunk_size=1` the first byte is observed as it arrives. A raw-socket client against the proxy confirms each chunk reaches the wire at upstream send time (0.00/0.50/1.00/1.50 s against 0.00/0.50/1.00/1.50 s sends).
- Row assertions wait for the row with a deadline helper (`_wait_for_rows`) because the capture persists **after** the last byte is relayed — the response is never blocked on the DB.
- The client-disconnect test reads the first line then closes; the next relayed write raises `EPIPE`/`ECONNRESET`, and the row lands with `error = "client disconnected"`.

## Verification

- `venv/bin/python -m pytest tests/test_inspector_sse_assembly.py tests/test_inspector_proxy.py -q` → **23 passed** (proxy module stable across 3 consecutive runs; streaming test uses a real timestamped `iter_lines` read, not sleep-and-hope)
- `venv/bin/python -m pytest tests/ -q` (unpipeled) → **1342 passed, 1 failed, 1 skipped** in 88.9 s
  - The single failure is `tests/test_flag_surface_guard.py::test_unpinned_llamacpp_image_id_still_current`: the local `llm-dock-llamacpp:latest` image is `58dd84544286` while the recorded flag-surface snapshot names `4e9495854d69` — the image was rebuilt on this host after the snapshot. Verified environmental per protocol: the **full suite at the merge base** (`421c992`) in a scratch worktree gives **1 failed, 1319 passed, 1 skipped** — same failure id, same count. This diff adds inspector files only (+23 tests).
- `ruff check` over the new/changed modules → clean.

Closes #236


teo-mateo:
READY_FOR_REVIEW 75f21dc4a26a9ec9e2b3a44eeb8b204a17f50a54

teo-mateo:
VERDICT APPROVED

Verified against the #236 spec and the merge base (`421c992`), not just the PR body:

- Spec cross-check, module by module: constructor signature (including the `default_model` the spec mandates, appended so positional calls keep working); the 8-name hop-by-hop set dropped in both directions with `Host` rewritten and `Authorization` passing through; the exact 502 message shape; `timeout=(10, 1800)`; the bounded buffer that stops appending at the cap while relaying continues (`response_truncated` OR'd with `seen_bytes > cap`, so a multi-byte body at the byte cap is still flagged); `db.trim(MAX_CAPTURES_ROWS)` after every insert; the whole capture path wrapped in `try/except Exception` logged at WARNING; `log_message` → `logger.debug`; all seven methods routed to one `_forward()`.
- All 13 acceptance criteria are pinned by tests, including the ones that matter: upstream receives the full 2 MB body while the row stores it truncated; `Authorization` verbatim upstream vs `***redacted***` stored; first streamed chunk observed before upstream finishes via a timestamped `iter_lines(chunk_size=1)` read; client disconnect persists `error = "client disconnected"`; a raising `insert_capture` still yields the correct proxied response; bind failure → `bind_error` without raising, and `stop()` releases the port for a second proxy.
- The one `db.py` touch (RLock around every public method + `check_same_thread=False` on the in-memory connection) is justified: the proxy's handler threads share the `:memory:` instance the tests use, and the re-entrant lock makes `insert_capture`'s internal `get_capture` safe. File mode is unaffected (per-call connections).
- Ran the evidence myself: the two new test files → 23 passed, proxy module stable across 3 consecutive runs; full suite unpipeled at head → 1 failed, 1342 passed, 1 skipped (matches the body). The single failure is the known image-id guard. At the merge base the same failure reproduces, plus one extra — `test_docker_events.py::test_event_manager_unregister_callback` — which passes 3/3 in isolation at the baseline: a timing flake (it spins a real container and races the event stream under full-suite load), pre-existing and unrelated to this diff. `ruff check` over the changed files: clean.
- 5 files, all inspector-scoped; the `db.py` hunks are a mechanical lock wrap with no behaviour change beyond the threading fix. No unrelated restyling.

NIT dashboard/inspector/db.py:8 — `logger` is still never used (carried over from the #243 nit; the file is touched here, so it earns a repeat mention).

Issue #241, opened 2026-09-18
inspector: docs, CHANGELOG and end-to-end verification


teo-mateo:
Part of #234. **Depends on: #235–#240.** Closes the feature out.

## Scope

Documentation, changelog, and one end-to-end verification on a live service.

## Documentation

**`CLAUDE.md`:**

- add `inspect` and `inspect_upstream_port` to the *services.json Schema* table, both
  marked as never being container flags (the `favorite` / `reasoning_levels` wording);
- a new top-level section, **Request inspector**, after *Per-conversation sampling
  parameters*, covering: what the proxy is and where it sits; the port relocation
  (`"127.0.0.1:<upstream>:<internal>"`) and the 34000–34999 band; that
  `get_docker_services()` is the host-port authority once inspection is on, and that
  `get_compose_service_ports()` must parse a three-part mapping; the capture path
  allowlist and why `/metrics`, `/health`, `/props`, `/slots` are excluded; the 1 MB body
  cap and the 2000-row trim; that `Authorization` is forwarded verbatim and redacted only
  in storage; and the endpoints;
- under *Gotchas*, two entries:
  **Only host-side traffic is captured.** Anything dialling a container directly over
    `llm-network` never reaches a host port and is not recorded. **Open WebUI** is the
    exception that matters (it dials `http://<service>:<internal>/v1` over the shared
    network, keeps working, and produces no captures), and `scripts/bench/vllm/bench.sh`
    runs `vllm bench serve` inside the container — likewise invisible. Chat, ghost chat,
    Android, `curl` against the published port, and `scripts/bench/llamacpp/bench.py`
    are host-side and captured.
  - **The proxy lives in the dashboard process.** Restarting the dashboard drops every
    inspected service's listener until `create_app()` re-syncs, and a request in that
    window fails to connect rather than falling through to the container — the container
    is on loopback under a different port by then.
- add `dashboard/inspector.db` to the machine-local storage paths paragraph in
  *Configuration*, with its `LLM_DOCK_INSPECTOR_DB` override;
- add `dashboard/inspector/` to the *Project Structure* tree, and `inspector_bp` to the
  blueprint registration list and the API routes table.

**`dashboard/frontend/AGENTS.md`:** the Inspector page's code map (components, hook,
service client) in the same shape as the existing feature maps.

**`CHANGELOG.md`:** one entry for the feature, matching the file's existing style.

## Verification (operator-run; ask before touching a running service)

On one service, with the operator's go-ahead:

[...]

PR #245, opened 2026-09-18
feat(inspector): enable/disable lifecycle — services.json field, compose relocation, supervisor (#237)

### LINKED ISSUE #237: inspector: enable/disable lifecycle — services.json field, compose port relocation, proxy supervisor

teo-mateo:
Part of #234 — sub-issue #237 (lifecycle). Depends on #236 (merged in #244). The proxy becomes a per-service toggle: the `inspect` field, the compose port relocation, upstream-port allocation, the enable/disable endpoint, and a supervisor that keeps the running proxies in step with `services.json` across boot, toggle, rename, port change and delete.

## What

**New files**

- **`dashboard/inspector/ports.py`** — `allocate_upstream_port`, pure. Band 34000–34999, lowest free where "free" means taken by no service's `port` **or** `inspect_upstream_port` (the service's own public port counts too — a public port inside the band would collide with the proxy's own `0.0.0.0` bind). Stable for a service that already has one; `ValueError` when exhausted.
- **`dashboard/inspector/supervisor.py`** — `ProxySupervisor` + the `proxy_supervisor` singleton.
  - `sync()` reads `services.json` through `ComposeManager` and reconciles: start where `inspect` is true (`default_model` = the alias, `template_type` passed through), stop where `inspect` was lost or the service deleted, rebind where `port`/`inspect_upstream_port` moved.
  - **Bindings are released before new ones are claimed**: on a port swap (`set-public-port`) one service's still-held port is handed to another, and a bind-before-release fails with EADDRINUSE.
  - Idempotent (a running proxy with unchanged ports is never touched), retries a previously failed bind on the next call, never raises from a partially configured service (inspects without an upstream port log a warning and skip).
  - Owns the single `InspectorDB` every proxy writes into — path `inspector.db`, override `LLM_DOCK_INSPECTOR_DB` or the app's `INSPECTOR_DB_PATH` config. A new db path invalidates the open connection via `configure()`. `rename_service` moves the capture rows only when a db exists, so an unrelated rename never creates the file. `stop_all()` releases every port.
  - `status()` → `[{service, listen_port, upstream_port, running, error}]`.

**Touched files**

- **`compose_manager._render_service`** — with `inspect` on, `context["port"]` becomes `"127.0.0.1:<upstream>"` so the mapping renders `"127.0.0.1:<upstream>:<internal>"`: one change point, all six templates, no template edits. Loopback binding is deliberate (the container must not be reachable around the proxy from the LAN). Absent/false (or `inspect` without an upstream port) renders byte-identical to before.
- **`docker_utils`** — `get_compose_service_ports()` parses the host port as the **second-to-last** colon component: the inspected mapping is `"127.0.0.1:34001:8080"` and `int("127.0.0.1")` takes out the whole port map for every service. `get_docker_services()` gains `inspect_map`/`upstream_port_map`, and `host_port` prefers the `services.json` port when `inspect` is on — the compose file then names the loopback upstream, not the public port clients dial; `services.json` is the authority. Both payload branches (container-exists and not-created) carry `"inspect"` and `"upstream_port"`, the way `reasoning_levels` does, so the SSE snapshot reads the same builder and `chat/llm_proxy.py` stays pointed at the proxy.
- **`flag_metadata.validate_service_config`** — `inspect` must be a boolean. `inspect_upstream_port` is **server-owned**: stripped from `POST`/`PUT /api/services` bodies by both write routes (a client-supplied value could pick a colliding or out-of-band port).
- **`routes/services.py`** — `POST /api/services/<name>/inspect` (`@require_auth`, `@_serialize_db`), body `{"enabled": bool}`:
  - **Enable order is the contract**: allocate the upstream port → write both fields → `rebuild_compose_file()` → `_recreate_if_running()` (the container releases the public port) → `proxy_supervisor.sync()` (the proxy claims it). Disable runs it backwards and **keeps** `inspect_upstream_port` so a re-enable doesn't churn the container.
  - Response `{"service", "enabled", "port", "upstream_port", "proxy_running", "restarted", "error"}` — 200 with `proxy_running: false` + populated `error` when the proxy could not bind (the state change did happen; the UI must say what went wrong).
  - Emits the `metadata-changed` SSE event exactly as `favorite` does, with `"metadata": {"inspect": enabled}`.
  - The other write paths all call `sync()`: **rename** (plus the capture-row rename), **set-public-port** (both the moved and the displaced proxy rebind), **delete** (captures are kept — deleting them is a deliberate action in the Inspector UI), **PUT update**, and `_rebuild_and_restart` for the paths that funnel through it.
- **`app.py`** — `create_app()` configures the singleton from app config and calls `sync()` once; `atexit.register(proxy_supervisor.stop_all)` next to `event_manager`'s.
- **`CLAUDE.md`** — the two new `services.json` schema rows (this belongs with the code that reads the fields; #241 owns the wider doc pass).

## Test semantics worth knowing

- 47 new tests, all backend, no Docker, no GPU: container start/stop/recreate is mocked in every test; no test runs `docker compose`.
- Supervisor tests run **real** `ServiceProxy` instances on ephemeral ports (real sockets, real bind/rebind), including three services inspected at once on arbitrary ports, a forced EADDRINUSE retry, and port-free assertions via raw-socket binds.
- The set-public-port test skips itself when 3301 is in use on the host (it is the public slot), and pins the displaced service's replacement port to a known-free ephemeral — the picker scans 3300–3399 against the compose file only, which knows nothing about host listeners (Open WebUI sits on 3300 on this machine; the picker picking it is exactly the class of collision this test exists to catch without colliding).
- The toggle tests drive a real `create_app` against a temp `services.json`; enable asserts the proxy answers **502 JSON** (upstream down) rather than refusing the connection.

## Verification

- `venv/bin/python -m pytest tests/test_inspector_ports.py tests/test_inspector_supervisor.py tests/test_inspector_compose.py tests/test_inspector_toggle_route.py -q` → **47 passed**, stable across 3 consecutive runs (~12.5 s each)
- `venv/bin/python -m pytest tests/ -q` (unpipeled) → **1389 passed, 1 failed, 1 skipped** in 105.5 s
  - The single failure is `tests/test_flag_surface_guard.py::test_unpinned_llamacpp_image_id_still_current`: the local `llm-dock-llamacpp:latest` image is `58dd84544286` while the recorded flag-surface snapshot names `4e9495854d69`. Verified environmental per protocol: the **full suite at the merge base** (`e54123e`) in a scratch worktree gives **1 failed, 1342 passed, 1 skipped** — same failure id, same count, and the +47 delta is exactly this PR's tests.
- `ruff check` over every touched file → clean.
- **Operator verification (not done here, per the issue's ground rules):** enabling on a live service recreates its container — the issue's live steps (`docker compose config` shows the loopback mapping, `curl` the public port for `/health` + one short completion, confirm a row exists, disable and confirm the mapping is back) are the operator's to run on a service that isn't serving anything.

Closes #237


teo-mateo:
READY_FOR_REVIEW 05597a96b0ff860ecd37ddb91fc3bdb517bca1ca

teo-mateo:
VERDICT APPROVED

Verified against the #237 spec and the merge base (`e54123e`), not just the PR body:

- **Spec cross-check, module by module.** `ports.py` matches the spec'd band/semantics (lowest free where "free" excludes every service's `port` *and* `inspect_upstream_port`, stable when one exists, `ValueError` on exhaustion). `supervisor.py`'s `sync()` reconciles start/stop/rebind and, crucially, **releases all bindings before claiming new ones** — I traced the `set-public-port` case: the displaced proxy (holding the target port) is stopped in the release loop *before* the start loop runs, so the moved proxy can claim it regardless of dict order; that's why the rebind test asserts both proxies running. The single-`InspectorDB` ownership, `LLM_DOCK_INSPECTOR_DB` override, `configure()` invalidation, `rename_service` no-op when no db exists, and `stop_all` all match. `compose_manager` is the one change point (`context["port"] = "127.0.0.1:<upstream>"`), absent/false/no-upstream-port renders the plain mapping. `docker_utils` parses the host port as `parts[-2]` (fixing the `int("127.0.0.1")` whole-map failure) and reports `host_port` from `services.json` when inspect is on. `flag_metadata` rejects a non-bool `inspect`; both write routes strip the server-owned `inspect_upstream_port`. The toggle route's enable order (write → rebuild → recreate → sync) and disable order (write → sync → rebuild → recreate, keeping the upstream port) match the contract; the SSE emit mirrors `set_favorite` exactly.
- **Ran the evidence myself.** All 47 new tests pass (token set; the 13 toggle-route tests import `app`, whose module-level `create_app()` calls `init_config()`, which needs `DASHBOARD_TOKEN` — a fresh worktree has no machine-local `.env`, so they error at setup without it; they pass 13/13 with the token, confirming that's environmental, not the diff). Full suite unpipeled at head → **1389 passed, 1 failed, 1 skipped** (matches the body). The single failure is the known image-id guard, which reproduces at the merge base; my baseline run also hit the pre-existing `test_docker_events` flake (passes 5/5 in isolation at baseline — a container-racing timing flake, unrelated to this diff). `ruff` clean over every touched file.
- **I verified the one behaviour I couldn't see a direct test for** by driving the real `get_docker_services()` with a fake Docker client: an inspected service reports `host_port` = the public port (not the loopback upstream) and carries `inspect`/`upstream_port`; a plain service reports `inspect: false`, `upstream_port: null` — in both the container-exists and not-created branches. So acceptance criterion #4's behaviour holds; see the SHOULD below for why it should still be pinned.

SHOULD dashboard/docker_utils.py:242 — `get_docker_services()` for an inspected service (host_port authority + `inspect`/`upstream_port` in both payload branches) is acceptance criterion #4 with no committed test. I verified it is correct by driving the real builder, but it is the one criterion pinned only by `get_compose_service_ports` (the helper) plus my manual check. This is exactly the class of bug the repo has been burned by before — the reasoning-levels gotcha in AGENTS.md is a hand-written payload that hid a dropped `template_type` and silently sent nothing. Add a test that builds the payload through the real `get_docker_services` for one inspected service (the `_payload_for` helper in `test_service_reasoning_levels.py` is the template). Non-blocking: the behaviour is present and verified.
NIT dashboard/tests/test_inspector_toggle_route.py — the disable→re-enable cycle (reusing the kept `inspect_upstream_port`) is not exercised end-to-end; `allocate_upstream_port`'s stability is tested in isolation but the full toggle round-trip is not.
NIT dashboard/inspector/supervisor.py:96 — a running proxy whose `default_model`/alias changed (via a non-rename `PUT alias=…`) is not restarted, because `stale` keys only on name/ports. Narrow: `default_model` is a body-has-no-`model` fallback, and the rename path (the common alias change) does restart it.
NIT dashboard/compose_manager.py:296 — the loopback-binding note is a real security invariant a future reader could "fix", but it sits in the function body; the repo convention places such comments at a declaration/interface boundary.

PR #248, opened 2026-09-18
feat(v2): request inspector viewer page (#240)

### LINKED ISSUE #240: inspector v2 UI: the capture viewer page
Part of #234. **Depends on: #238.** React v2 only. Read `dashboard/frontend/AGENTS.md`
before starting.

## Scope

The viewer: a new `/inspector` page listing captures newest-first and rendering the
selected one in full — the conversation as sent, the tool definitions, the raw request,
the response, and the metadata. Plus deletion.

## Files

- `dashboard/frontend/src/services/inspector.js` — API client
- `dashboard/frontend/src/hooks/useInspectorCaptures.js` (+ `.test.js`)
- `dashboard/frontend/src/components/inspector/InspectorPage.jsx` (+ `.test.jsx`)
- `dashboard/frontend/src/components/inspector/CaptureList.jsx` (+ `.test.jsx`)
- `dashboard/frontend/src/components/inspector/CaptureDetail.jsx` (+ `.test.jsx`)
- `dashboard/frontend/src/components/inspector/ConversationView.jsx` (+ `.test.jsx`)
- `dashboard/frontend/src/routes.jsx` — `{ path: 'inspector', element: <DefaultLayout><InspectorPage /></DefaultLayout> }`
- `dashboard/frontend/src/components/navItems.js` — `{ to: '/inspector', end: false, icon: 'fa-magnifying-glass', label: 'Inspector' }`, placed after Tools

Reuse, do not reimplement: `CopyablePre` for code/JSON blocks, the chat's markdown
rendering approach (`react-markdown` + `remark-gfm`) for rendered message bodies, and
`useProseClass` for prose styling. Theme tokens only — no new colors.

## Page layout

Header bar: `Inspector` title; then, right-aligned —

- a **service filter** `<select>`: `All services (137)` plus one option per entry from
  `GET /api/inspector/services`, each `<name> (<count>)`. Selecting one sets `?service=`
  in the URL (`useSearchParams`) so the view is linkable and survives a reload; #239's
  card links straight into it.
- a **Live** toggle (default on). When on, the list refetches every 5 s. When a capture is
  selected, refetching must not steal the selection or scroll position.
- a **Refresh** button (always enabled, spins while loading).
- a **Delete all** button, danger-styled. Its label follows the filter: `Delete all` or
  `Delete all for <service>`. Opens a confirm modal reading `Delete <n> captures? This
  cannot be undone.` with `Cancel` / `Delete`.

Body: a two-pane grid, `xl:grid-cols-[400px_1fr] gap-6`. Below `xl`, the panes stack with
the list first, capped at `max-h-80` and scrollable.

### Left pane — capture list

Newest first. Each row, in two lines:

- line 1: a status pill (`200` green, `4xx`/`5xx` red, `—` muted when the row has an
  `error` and no status), the **model** in monospace (truncated with a `title`), and a
  right-aligned relative time (`just now`, `4m ago`, `2h ago`, then `Sep 18 20:14`), with
  the absolute ISO timestamp as its `title`;
- line 2, muted and smaller: the service name, a stream icon (`fa-bolt`) when
  `stream === true`, `↑<prompt_tokens> ↓<completion_tokens>` when present, and the
  duration (`8.4s`, or `412ms` under a second).

A row that is truncated (`request_truncated || response_truncated`) shows a `truncated`
chip. A row with a non-null `error` shows a red `fa-triangle-exclamation` and its message
as the `title`.

Interactions:

- Click selects; the selected row gets an accent left border and a raised background.
  Selection is held in `?capture=<id>` so a reload keeps it.
- Hover reveals a trash icon on the right of line 1. Clicking it deletes **that** row
  after an inline confirm (the icon turns into `Delete?` / `Cancel`) — no modal. Deleting
  the selected row clears the detail pane and selects the next-newest remaining row.
- `↑`/`↓` move the selection when the list has focus.
- Footer: `Showing <n> of <total>` and a `Load more` button (50 per page) while
  `n < total`. Appended pages never reorder what is already shown.
- Empty state, no captures at all: a magnifier glyph and `No captures yet.` plus
  `Turn on Request inspection on a service to start recording.` with a link to `/`.
- Empty state under a filter: `No captures for <service>.`

### Right pane — capture detail

Header: the model in monospace as the title; under it, muted, `<service> · <method> <path>
· <status> · <duration>` and the absolute timestamp. Right-aligned: a `Delete` button
(inline confirm, same pattern) and a `Copy request JSON` button.

Below it, a sub-tab bar styled like `ServiceDetailsPage`'s `TabBar`, tabs in this order.
Tab state lives in component state (not the URL) and resets to **Conversation** on
selecting a different capture.

**1. Conversation** (default)

Renders `request_body.messages` in order, one card each:

- a role chip: `system` (violet), `user` (blue), `assistant` (green), `tool` (amber),
  anything else (muted), with the 1-based index and, for `tool`, the `tool_call_id` in
  monospace;
- string content: preserved whitespace, monospace, wrapped — with a per-card
  `Rendered / Raw` switch defaulting to **Raw**, since seeing the literal prompt is the
  point. `Rendered` runs it through the chat's markdown renderer;
- array content (multimodal): text parts rendered as above, `image_url` parts rendered as
  a thumbnail (a `data:` URL renders inline; anything else renders as a link), each
  labelled with its approximate size;
- an assistant message carrying `tool_calls`: each call as a collapsible block — the
  function name in monospace, and pretty-printed `arguments` (parsed when it is a JSON
  string, otherwise shown verbatim with a `not valid JSON` chip);
- any card longer than 40 lines collapses to 40 with a `Show all (<n> lines)` control;
- each card has a `Copy` button for its own content.

Above the cards, a one-line summary: `<n> messages · <k> tools · <approx chars> chars`.

The **system prompt is never collapsed by default** — it is the payload people come here
to read.

If `request_body` does not parse as JSON, or has no `messages`, the tab shows
`This request has no message array.` and a `View raw request` link that switches to the
Request tab. The pane must never blank on malformed input.

**2. Tools** — badge with the count. One collapsible block per entry in
`request_body.tools`: the function name (monospace) and its `description` when open, plus
the pretty-printed `parameters` schema. A `Expand all` / `Collapse all` control at the
top. Empty state: `No tools were sent with this request.`

**3. Request** — first a small table of every top-level request field **except**
`messages` and `tools` (`model`, `stream`, `temperature`, `max_tokens`, `reasoning_effort`,
`chat_template_kwargs`, …), object values pretty-printed; then the complete raw request
body in a `CopyablePre`. When `request_truncated`, a banner above it: `Request body was
truncated at 1 MB.`

**4. Response** — the assembled `response_text` (with the same `Rendered / Raw` switch,
defaulting to **Rendered** here); `reasoning_text` above it in a collapsed
`Thinking` disclosure when present; response `tool_calls` rendered like the assistant
ones; then a collapsed `Raw response` disclosure holding `response_body` verbatim (the SSE
stream for a streamed call). `finish_reason` as a chip. When `response_truncated`, the
same style of banner.

**5. Meta** — a definition list: service, model, template type, method, path, status,
stream, TTFB, duration, prompt/completion/total tokens, created at (absolute, with the
local timezone), error, both truncation flags, capture id; then the redacted request
headers as a table. A test asserts `***redacted***` renders and no bearer value ever does.

Empty right pane when nothing is selected: `Select a capture to inspect it.`

## Behaviour details

- Fetch the detail only when a capture is selected, and cache the last-fetched detail so
  re-selecting the same row does not refetch.
- A detail fetch that 404s (the row was deleted in another tab) clears the selection and
  shows a muted `This capture no longer exists.`
- Live refetch is paused while a delete confirm or the delete-all modal is open.
- All timestamps display in the browser's local timezone; the stored value is UTC.

## Acceptance criteria

- [ ] `/inspector` renders from the nav item and from a direct URL; `?service=` and
      `?capture=` are honoured on load and kept in sync with the UI.
- [ ] List renders newest-first with status, model, service, stream icon, tokens, duration
      and relative time; the truncated chip and the error icon appear when the flags say so.
- [ ] Selecting a row loads the detail; selecting another resets the sub-tab to
      Conversation.
- [ ] Conversation renders system / user / assistant / tool messages with distinct role
      chips, in order, with the system prompt expanded.
- [ ] An assistant message with `tool_calls` renders the function name and pretty-printed
      arguments; arguments that are not valid JSON still render, flagged.
- [ ] A multimodal message with a `data:` image renders a thumbnail and does not crash on
      a non-data URL.
- [ ] A capture whose `request_body` is `"{ not json"` renders the fallback copy and the
      Request tab still shows the raw text — **no blank pane, no thrown error**.
- [ ] Tools tab lists definitions with a count badge, and shows the empty state when
      `tools` is absent.
- [ ] Request tab lists non-message fields and the raw body; the truncation banner appears
      when flagged.
- [ ] Response tab renders assembled text, a collapsed Thinking block when
      `reasoning_text` is present, and the raw SSE under a disclosure.
- [ ] Meta shows `***redacted***` for `Authorization`.
- [ ] Deleting a row from the list removes it without a full page refetch and moves the
      selection to the next row.
- [ ] `Delete all` is scoped by the active filter and asks for confirmation naming the
      count.
- [ ] `Load more` appends without reordering; `Showing n of total` is accurate.
- [ ] Live refetch does not change the selection or reset the sub-tab.
- [ ] Layout works at phone width (panes stacked, no horizontal page scroll) and in both
      themes.
- [ ] `cd dashboard/frontend && npm test` passes and `npm run lint` is clean.

teo-mateo:
feat(v2): request inspector viewer page

Part of #234 (sub-issue #240, depends on #238). React v2 only. The viewer: a new
/inspector page listing captures newest-first and rendering the selected one in
full. Backend untouched (the #238 API is the whole surface).

New files:
- services/inspector.js: listCaptures/getCapture/deleteCapture/deleteCaptures/
  inspectorServices over fetchAPI.
- hooks/useInspectorCaptures.js: paging (50/page), live 5 s refetch with a pause
  flag, merge-based refresh (a Load-more'd tail survives a refresh), append-only
  deduping loadMore, reload for the Delete-all reset (merge-based refresh cannot
  empty the list), local removeRow. Pure mergeReload/appendPage exported.
- components/inspector/parse.js: shared pure helpers (relativeTime,
  formatDuration, formatAbsolute, parseRequestBody, prettyArguments,
  messageContentChars, approxImageSize, copyText) — a component file exporting a
  non-component breaks Fast refresh, so they live in a plain module.
- components/inspector/CaptureList.jsx: two-line rows (status pill, model,
  relative time; service, stream icon, token pair, duration, truncated chip),
  hover trash with inline Delete?/Cancel (confirm state held per-row by the
  list, reported upward so the page can pause live refetch), ArrowUp/ArrowDown
  selection, footer with Showing n of total + Load more, both empty states.
- components/inspector/ConversationView.jsx: message cards with role chips
  (system violet, user blue, assistant green, tool amber via badge tokens),
  1-based index and tool_call_id, per-card Rendered/Raw defaulting to Raw,
  >40-line collapse that never collapses the system prompt, multimodal parts
  (data: URLs inline with approx size, others as links), assistant tool_calls
  as collapsible pretty-arguments blocks with a not-valid-JSON flag, and the
  no-message-array fallback with a View raw request callback.
- components/inspector/CaptureDetail.jsx: header (model, meta line, inline
  confirm Delete, Copy request JSON) + ServiceDetailsPage-styled sub-tab bar —
  Conversation / Tools (count badge, expand/collapse all) / Request (non-
  message field table + raw CopyablePre + 1 MB truncation banner) / Response
  (Rendered-by-default text, collapsed Thinking, response tool_calls, finish
  chip, collapsed raw SSE disclosure) / Meta (definition list + redacted header
  table).
- components/inspector/InspectorPage.jsx: header bar (title, service filter
  select with per-service counts, Live switch default on, always-enabled Refresh
  that spins, danger Delete all whose label follows the filter and confirms with
  the count), xl:grid-cols-[400px_1fr] body with stacked max-h-80 list below xl,
  ?service=/?capture= URL state (linkable, reload-stable), detail-on-demand with
  an in-memory cache (capture content is immutable), 404 → selection cleared +
  "This capture no longer exists.", deleting the selected row selects the
  next-newest remaining row, live refetch paused while any confirm is open.
- Five test files: 63 cases over the acceptance list.

Touched files:
- routes.jsx: { path: 'inspector', element: DefaultLayout-wrapped InspectorPage }.
- navItems.js: Inspector after Tools (Sidebar and MobileNav both consume it).

Verified:
- npm test -- src/hooks/useInspectorCaptures.test.js src/components/inspector/
  -> 63 passed.
- npm test -> 688 passed, 53 files.
- npm run lint (--max-warnings 0) -> clean.
- npm run build -> clean (pre-existing >500kB chunk warning only).
- Backend untouched, so no pytest run is implicated.

Closes #240

## Reconstruction note

Work on this branch was interrupted mid-edit (the undo the operator mentioned);
the tree was reconciled before committing. What that found and fixed:
`CaptureRow` was mid-refactor — destructuring a `confirming` prop while also
declaring a same-name local `useState` (a hard SyntaxError) and calling an
undeclared `onAskDelete`. The consistent design is the one the parent already
wired (per-row confirm state held by the list, reported upward via
`onConfirmStateChange`), so the remnant was removed; the parent-managed props
were kept verbatim.

## Notes

- Pure helpers live in `components/inspector/parse.js`, not in the component
  files — `react-refresh/only-export-components` is a lint **error** under
  `--max-warnings 0`, and this repo lints at that level.
- The `Delete all` confirm uses the hook's `reload()` rather than `refresh()`:
  refresh merges (so a Load-more'd tail survives — deliberate), and a merge
  cannot empty a list.
- The live 5 s refetch pauses while any delete confirm (row, detail, modal) is
  open, so a capture cannot vanish under the user mid-confirm.
- Detail responses are cached per id in memory: a stored capture is immutable,
  so re-selecting a row is free; a delete evicts the id.

Live-browser verification on real captures is the operator's (nothing is being
inspected until a service toggles it on); the UI contract is covered by the 63
vitest cases against mocked API responses.


teo-mateo:
READY_FOR_REVIEW 2c21857233af6447124459a4ab6617f86108eaa5

teo-mateo:
VERDICT APPROVED

Verified against the #240 spec and the merge base (`bc714be`), not just the PR body. All 17 acceptance criteria hold.

**Spec conformance verified in code**
- Route and nav are verbatim per spec: `{ path: 'inspector', element: <DefaultLayout><InspectorPage /></DefaultLayout> }` and the nav entry after Tools with `fa-magnifying-glass`/`Inspector`; `Sidebar` and `MobileNav` both consume `NAV_ITEMS`, so one line covers both. URL state is real `useSearchParams` — `?service=` drives the fetch (`listCaptures({service})`), the `<select>` value, the `Delete all for <service>` label and the modal count; `?capture=` drives selection and survives reload.
- API client matches the #238 contract field-for-field (`captures/total/limit/offset`, summary fields from `_summary_row`, detail adds `request_headers/request_body/response_body/response_text/reasoning_text/tool_calls`); newest-first is server-guaranteed (`ORDER BY created_at DESC, rowid DESC`), and `mergeReload` keeps fresh rows on top, so the ordering invariant holds through live ticks.
- **Reuse, not reimplement**: `CopyablePre` (`{children, ...rest}`, so the `className` uses are valid) and `useProseClass(...extra)` are called against their real signatures; markdown goes through `react-markdown` + `remark-gfm` exactly as the chat does.
- **Theme tokens only**: every class resolves to an existing CSS variable in *both* themes (`success-subtle`, `warning-subtle/-fg`, `border-subtle`, `surface-muted`, `surface-strong`, `accent`, the `badge-*` family), and the role chips land on the spec's hues precisely — vllm violet → system, llamacpp blue → user, webui green → assistant, ds4 amber → tool. Modal overlay `bg-black/60` and `hover:bg-danger` match existing precedent (`CreateServiceModal`, `DefaultPromptEditor`), not new colours. The TabBar class strings are identical to `ServiceDetailsPage`'s.
- Reconstruction note checks out: `CaptureRow` now destructures `confirming` as a prop with no same-name local `useState`, `onAskDelete` is declared in the signature and passed by the parent, and the list owns the only confirm state (`CaptureList.jsx:123`) — the SyntaxError remnant is genuinely gone, and the parent-managed props are the design the list already implemented.
- Comments all carry constraints rather than narration (parse.js's Fast-refresh reason, the merge/dedup semantics, the live-pause reason, the immutable-capture cache reason, the insecure-origin clipboard fallback) — the allowed category.

**Evidence I ran myself**: the 5 new files → **63 passed**; full `npm test` → **688 passed / 53 files**; `npm run lint` (`--max-warnings 0`) → clean; `npm run build` → clean, pre-existing >500 kB warning only. All four match the body. No backend files and no dependency changes, so no pytest implication — the diff is 14 files, 2173 insertions, all under `dashboard/frontend/src/`.

**Two criteria the committed tests do not pin, which I verified myself** (in a throwaway worktree, nothing written to this tree):
- *Live refetch does not steal the selection or the sub-tab* — I drove the real page with fake timers at `?capture=c1`, opened the Request tab and advanced a full 5 s tick: the list refetched, the URL kept `capture=c1`, the Request tab stayed. Worth committing that case (SHOULD below).
- *Phone width / both themes* — `grid-cols-1 xl:grid-cols-[400px_1fr]` with the list first and `max-h-80`, `min-w-0` on the detail pane plus `break-words`/`break-all` on every long-run field, and tokens only, so there is nothing to overflow the viewport or go unthemed. Live-browser confirmation stays with the operator as the PR states.

SHOULD dashboard/frontend/src/components/inspector/ConversationView.jsx:150 — a **multimodal** card over 40 lines collapses but renders no expand control, so its text past line 40 is unreachable. The spec says "*any* card longer than 40 lines collapses to 40 **with a `Show all (<n> lines)` control**", and the control sits inside the `typeof content === 'string'` branch only. I confirmed it with a scratch test: `collapsed tail present: true`, `Show all control present: false`. Two-line fix — move the `{collapsible && <button …>}` block out of the string-only branch so both content shapes get it, and pin it with the multimodal case.
SHOULD dashboard/frontend/src/components/inspector/InspectorPage.jsx:22 — the hook returns `refreshServices` but the page never calls it, so the `<select>` counts come from the mount fetch only: after any delete they over-report, and a service that turns inspection on mid-session never appears while the tab stays open, because the live tick refetches captures only. Call it from the live tick (and after `handleRemove`/`confirmDeleteAll`) so the header stays honest in the one mode the page defaults to.
SHOULD dashboard/frontend/src/components/inspector/InspectorPage.test.jsx — the live-tick case described above is the missing piece of an otherwise complete pass over the acceptance list; it needs no new scaffolding beyond `vi.useFakeTimers()` and the `ParamsProbe` already in the file.
NIT dashboard/frontend/src/components/inspector/InspectorPage.jsx:70 — the detail `catch` treats *every* failure as a 404: it clears `?capture=` and reports "This capture no longer exists." A transient blip then tells the user their capture was deleted and throws away the selection. The repo convention is to branch on the status (`useServiceDetails` does `err.message.includes('404')`) — keep the message and the URL clear for 404 only.
NIT dashboard/frontend/src/components/inspector/parse.js:26 — `formatAbsolute` is a bare `toLocaleString()`, so Meta shows local time but never names the zone; the spec asks for "created at (absolute, with the local timezone)". `toLocaleString(undefined, { timeZoneName: 'short' })` closes it.
NIT dashboard/frontend/src/components/inspector/CaptureDetail.test.jsx:186 — the Meta case asserts `***redacted***` renders but not the second half of the criterion ("no bearer value ever does"): the fixture carries only the marker, so nothing would fail if a future edit rendered an unredacted field. Add a token-shaped decoy to the fixture and assert no `Bearer`/token substring appears in the rendered tab.
NIT dashboard/frontend/src/components/inspector/CaptureList.test.jsx:78 — `expect(screen.getByText('8.4s')).toBeInTheDocument()` appears twice; the second reads like it was meant for the relative time. The `it(...)` title also promises "an empty detail prompt state", which `CaptureList` does not own.
NIT dashboard/frontend/src/hooks/useInspectorCaptures.js:80 — `loadMore` derives its offset from `rows.length`, which assumes every shown row still exists server-side; a row deleted in another tab is kept in the tail by `mergeReload`, shifting the offset so `Load more` can skip one row. And the call is uncaught (`onLoadMore={() => loadMore()}`), so a 400/500 becomes an unhandled rejection with a button that silently does nothing. Deriving the offset from a fetched-row counter and catching the rejection fixes both.

